PR #2, opened 2025-04-07
Extracted DF report types

silaslenihan:


capabilities/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/smartcontractkit/chainlink-evm/capabilities
+
+go 1.24.1
+
+require (
+	github.com/ethereum/go-ethereum v1.15.7
```
jmank88: Does it need to be this new? We are struggling to upgrade core already, so you'll be blocked if you bump this far.

silaslenihan: downgraded to match core

capabilities/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/smartcontractkit/chainlink-evm/capabilities
```
pavel-raykov: do we need a separate go module for this ?

krebernisak: No, I don't think we do. At least not with the initial move.

@pavel-raykov and I discussed how https://github.com/smartcontractkit/chainlink-integrations is moving here to root Go module, with `<source>/evm` moving to `<dest>/pkg` as an evm set of packages.

I suggest this PR follows the same model - add to root Go module, under `pkg/report/<type>/..`

capabilities/writetarget/report/mercury/v3/types.go
```diff
@@ -0,0 +1,57 @@
+// TODO: duplicate of https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/mercury/v3/types/types.go
```
krebernisak: Notice that these types are duplicated from core repository to avoid importing core. @silaslenihan could you please scope out a ticket to dedupe this in a follow up. Thx!

cc @jmank88

capabilities/go.mod
silaslenihan: done

capabilities/writetarget/report/mercury/v3/types.go
silaslenihan: https://smartcontract-it.atlassian.net/browse/BCFR-1226

dhaidashenko (approved):

PR #21, opened 2025-04-09
Support for Tron

krebernisak:
This PR adds support for Tron though a couple small changes:

- Adds a new `chainType` of `tron`
- Adds `ToBackwardCompatibleCallArgWithChainTypeSupport` which allows custom call arg behaviour for a given `chainType`
- New config option of `ChainSpecificURL` which can be added to a node  
  - Used for chains that require an additional URL in conjunction to the standard JSON-RPC URL
- A new method on `MessageSigner` which signs raw bytes without `keccak256` hashing the message
- More flexible hash unmarshalling for certain hashes within the head

----
Moving code from: https://github.com/smartcontractkit/chainlink-integrations/pull/25


pkg/types/models.go
```diff
@@ -254,17 +255,45 @@ func (h *Head) AsSlice(k int) (heads []*Head) {
...
 }
 
+// Hash represents the 32 byte Keccak256 hash of arbitrary data.
+// Mainly to add compatibility with Tron as some hashes are returned as 0x00 or 0x
+type LessStrictHash [32]byte
+
+// UnmarshalJSON parses a hash in hex syntax.
+func (h *LessStrictHash) UnmarshalJSON(input []byte) error {
+	// "0x00" or "0x" or []byte{}
+	if len(input) == 6 || len(input) == 4 || len(input) == 0 {
```
krebernisak: We don't check these bytes but just check the length.

Shouldn't we unmarshal as before and then pad left with 0 until [32]byte as a result?

pkg/client/chain_client.go
```diff
@@ -334,6 +336,15 @@ func (c *chainClient) IsL2() bool {
 	return c.chainType.IsL2()
 }
 
+// GetExternallyUsedChainSpecificURL returns the additional URL that a chain may require to be used in conjunction with the standard JSON-RPC URL
+func (c *chainClient) GetExternallyUsedChainSpecificURL(ctx context.Context) (*commonconfig.URL, error) {
```
KodeyThomas: couldn't find an escape hatch into the RPC client, if anyone knows of a better way to handle this i'm open for suggestions

krebernisak: Yeah, I was looking, didn't find a good solution yet...

krebernisak: I think it's going to be cleaner to pass it in through a Chain object (EVMLegacy supporting TRON), vs. using an EVM client here to pass this info. I'll give it a go.

pkg/client/rpc_client.go
```diff
@@ -775,6 +805,13 @@ func (r *RPCClient) NonceAt(ctx context.Context, account common.Address, blockNu
 
 	lggr.Debug("RPC call: evmclient.Client#NonceAt")
 	start := time.Now()
+
+	if r.isChainType(chaintype.ChainTron) {
```
dhaidashenko, line 799: If we are not using EVM TXM for Tron, does it make sense to disable it and return an error here instead of 0?

pkg/client/rpc_client.go
```diff
@@ -709,6 +726,12 @@ func (r *RPCClient) SendTransaction(ctx context.Context, tx *types.Transaction)
 	lggr.Debug("RPC call: evmclient.Client#SendTransaction")
 	start := time.Now()
 	var err error
+
+	if r.isChainType(chaintype.ChainTron) {
+		err = errors.New("SendTransaction not implemented for Tron, this should never be called")
+		return struct{}{}, ClassifySendError(err, r.clientErrors, logger.Sugared(logger.Nop()), tx, common.Address{}, r.chainType.IsL2()), err
```
dhaidashenko:
```suggestion
		return struct{}{}, multinode.Fatal, err
```

pkg/client/rpc_client.go
```diff
@@ -845,6 +882,12 @@ func (r *RPCClient) EstimateGas(ctx context.Context, c interface{}) (gas uint64,
 
 	lggr.Debug("RPC call: evmclient.Client#EstimateGas")
 	start := time.Now()
+
+	if r.isChainType(chaintype.ChainTron) {
+		err = r.wrapHTTP(http.rpc.CallContext(ctx, &gas, "eth_estimateGas", ToBackwardCompatibleCallArgWithChainTypeSupport(call, r.chainType)))
```
dhaidashenko: Do we need `ToBackwardCompatibleCallArgWithChainTypeSupport` to be public?
It seems that we need to modify call arguments only for CallContext. Does it make sense to create a single wrapper method for it?

pkg/client/rpc_client_test.go
```diff
@@ -712,6 +712,117 @@ func TestRpcClientLargePayloadTimeout(t *testing.T) {
 	}
 }
 
+func TestRPCClient_SendTransaction_Tron(t *testing.T) {
```
dhaidashenko: The configuration between Tron-specific tests takes 90% of the code. Should we just have a single test that covers all Tron specific calls?

pkg/types/models.go
```diff
@@ -254,17 +255,45 @@ func (h *Head) AsSlice(k int) (heads []*Head) {
 	return
 }
 
+// Hash represents the 32 byte Keccak256 hash of arbitrary data.
+// Mainly to add compatibility with Tron as some hashes are returned as 0x00 or 0x
+type LessStrictHash [32]byte
```
dhaidashenko: Nit: should we just call it Hash as there is no StrictHash or Hash?

pkg/types/models.go
```diff
@@ -280,17 +309,17 @@ func (h *Head) UnmarshalJSON(bs []byte) error {
 		return nil
 	}
 
-	h.Hash = jsonHead.Hash
+	h.Hash = common.BytesToHash(jsonHead.Hash.Bytes())
```
dhaidashenko: We can just cast it to common.Hash. This might help us to avoid allocation during conversion of [32]bytes to []bytes

pkg/config/toml/config.go
```diff
@@ -1145,11 +1145,12 @@ func (o *OCR) setFrom(f *OCR) {
...
 type Node struct {
-	Name     *string
-	WSURL    *commonconfig.URL
-	HTTPURL  *commonconfig.URL
-	SendOnly *bool
-	Order    *int32
+	Name             *string
+	WSURL            *commonconfig.URL
+	HTTPURL          *commonconfig.URL
+	ChainSpecificURL *commonconfig.URL
```
dhaidashenko: Is it a URL that points to the same RPC instance but just a different set of API?

KodeyThomas: yeah we're using the standard JSON-RPC API for reading for Tron just need the full node API to send transactions

pkg/client/chain_client.go
krebernisak: WIP here:

- https://github.com/smartcontractkit/chainlink-evm/pull/21/commits/7955219e0ca9e01a2d749305b35f3dea463c7c03
- https://github.com/smartcontractkit/chainlink/pull/17096/commits/12259aef7fb9d205756573206856c956197ae34d

pkg/client/rpc_client.go
KodeyThomas: makes sense have updated

pkg/client/rpc_client.go
KodeyThomas: Originally made it public for easier testing and as we already made the original method public. have added a wrapper method

pkg/types/models.go
KodeyThomas: Think common.Hash is a type so we can't easily cast to it, my go experience isn't the best but think the only way to do that and avoid allocation would be something like this which i'd rather avoid. 

If theres a better way happy to implement it

```
func (h Hash) ToCommonHash() common.Hash {
	return *(*common.Hash)(unsafe.Pointer(&h))
}
```

dhaidashenko: common.Hash is a type alias to the `[HashLength]byte`, so casting should work. Just tried it locally - works for me
```suggestion
	h.Hash = common.Hash(jsonHead.Hash)
```

dhaidashenko (approved):

DylanTinianov (approved):

PR #22, opened 2025-04-09
Logpoller Latency Monitor

DylanTinianov:
Create logpoller LatencyMonitor to log warnings if client methods latency is too high

Ticket: https://smartcontract-it.atlassian.net/browse/BCFR-1199

### DIFF
--- a/pkg/logpoller/latency_monitor.go
+++ b/pkg/logpoller/latency_monitor.go
@@ -1,0 +1,97 @@
+package logpoller
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


reductionista (approved):

reductionista:
As others have suggested, would be nice to have the threshold configurable. But approving it since I don't see that as a strict necessity. I'm fine with us adding that later if it becomes more clear that it's needed.

huangzhen1997 (approved):
lgtm

PR #24, opened 2025-04-09
Moved Generalized WT out, created chain-specific injected processors

silaslenihan:
The generalized WT was previously in this repo as it required EVM-specific code. In this latest version, the generalized WT has been moved to `chainlink-framework`, with chain-specific functionality extracted in favor of being injected. This PR houses the chain-specific code for EVM. 

# Requires
- https://github.com/smartcontractkit/chainlink-common/pull/1160
- https://github.com/smartcontractkit/chainlink-framework/pull/43

# Supports
- https://github.com/smartcontractkit/chainlink/pull/17216


pkg/writetarget/write_target_monitor.go
```diff
@@ -30,7 +30,7 @@ const (
 // includes decoding messages as specific types and deriving metrics based on the decoded messages.
 // TODO: Report decoding uses the same ABI for EVM and Aptos, however, future chains may need a different
 // decoding scheme. Generalize this in the future to support different chains and decoding schemes.
-func NewMonitor(ctx context.Context, lggr logger.Logger) (*monitor.BeholderClient, error) {
+func NewMonitor(ctx context.Context, lggr logger.Logger, decodeFn func(m *wt.WriteConfirmed) ([]*registry.FeedUpdated, error)) (*monitor.BeholderClient, error) {
```
jmank88: Pre-existing, but why does this take a context?

jmank88 (dismissed):

pkg/report/platform/types.go
```diff
@@ -3,16 +3,15 @@ package platform
 import (
 	"bytes"
 	"encoding/binary"
-	"encoding/hex"
 	"fmt"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
+	"github.com/smartcontractkit/chainlink-evm/pkg/report/datafeeds"
```
krebernisak: Why did you change from a common to a specific `chainlink-evm/pkg/report/datafeeds` type which is not even a datafeeds type?

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -128,7 +71,7 @@ func newFeedUpdated(
 // for most use-cases. For big numbers, benchmark bytes should be used instead.
 //
 // Returns `math.NaN()` if report data type not a number, or `+/-Inf` if number doesn't fit in double.
-func toBenchmarkVal(feedID datafeeds.FeedID, val *big.Int) float64 {
+func ToBenchmarkVal(feedID datafeeds.FeedID, val *big.Int) float64 {
```
krebernisak: Why is this now public?

pkg/report/datafeeds/evm/types.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/smartcontractkit/chainlink-evm/pkg/report/datafeeds"
+)
+
+type Report struct {
```
krebernisak: Why exactly are you duplicating this type here?

pkg/writetarget/write_target_monitor.go
krebernisak: 1. `registry.FeedUpdated` is product specific, we don't want this type here
2. Typing it like `decodeFn func(m *wt.WriteConfirmed) ([]*registry.FeedUpdated, error)` is super lazy - if you add a type like this it needs to be extracted and documented.
3. `monitor.ProtoProcessor` general type to process messages already exists - so reuse that one and inject a specific processor for specific chan

krebernisak (changes requested):

pkg/report/datafeeds/evm/decode.go
```diff
@@ -0,0 +1,38 @@
+package evm
```
krebernisak: Why is this in an evm package? This is a chainlink-evm repo

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -1,74 +1,17 @@
 package registry
```
krebernisak: This file was specifically put on a path `pkg/monitoring/pb/data-feeds/on-chain/registry` as this is the package to specifically deal with "monitoring/pb" (monitoring protobufs decoding), specifically the "beholder_domain" attr as "data-feeds/on-chain/registry" - does this makes sense? Please move it back and structure all Beholder monitoring packages for EVM, Aptos and every other chain like that, until a better proposal is documented and aligned.

krebernisak (changes requested):

pkg/report/datafeeds/evm/decode.go
silaslenihan: this was a mistake leftover from my early prototyping

pkg/report/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: another mistake from early prototyping

pkg/writetarget/write_target_monitor.go
silaslenihan: fixed

pkg/report/platform/types.go
silaslenihan: We were also duplicating this type before, I just moved it from core to chainlink-evm. I tried to generalize the `common` type, but the differences in types `[]byte`s vs `string`s was causing issues in the ocr tests, so I elected to leave this for a future improvement when I have more time to work out the kinks

pkg/writetarget/write_target_monitor.go
silaslenihan: Updating the constructor to accept a list of `monitor.ProtoProcessor` as chain-specific processors, which will be used to inject chain-specific df processors for EVM and Aptos.

pkg/report/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: Moved this file back, will mimic the structure in the aptos PR as well

pkg/report/datafeeds/evm/types.go
silaslenihan: deduplicated

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -33,42 +29,18 @@ func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*FeedUpdated, error) {
...
+		msgs = append(msgs, NewFeedUpdated(m, feedID, rf.Timestamp, rf.Price, r.Data, false))
 	}
 
 	return msgs, nil
 }
 
 // newFeedUpdated creates a FeedUpdated from the given common parameters.
 // If includeTxInfo is true, TxSender and TxReceiver are set.
-func newFeedUpdated(
+func NewFeedUpdated(
```
krebernisak: No need for this to be public, right?

pkg/report/pb/data-feeds/on-chain/registry/decode.go
krebernisak: Thank you!

pkg/report/platform/types.go
krebernisak: Makes sense! Please scope out a ticket and lets clean it up in a follow up.

pkg/report/datafeeds/id.go
```diff
@@ -90,3 +92,33 @@ func GetDecimals(dataType uint8) (uint8, bool) {
...
+func (rm Metadata) Length() int {
+	bytes, err := rm.Encode()
+	if err != nil {
+		return 0
+	}
+	return len(bytes)
+}
+
+// Note: This should be a shared type that the OCR3 package validates as well
+type Metadata struct {
```
krebernisak: I'd prefer if this goes to a specific `metadata.go` file with a comment and a TODO to clean up and align with the common type.

pkg/report/datafeeds/types.go
```diff
@@ -2,43 +2,46 @@ package datafeeds
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 )
 
-// This is ABI encoding - abi: "(bytes32 FeedID, bytes RawReport)[] Reports" (set in workflow)
-// Encoded with: https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/cap_encoder.go
```
krebernisak, line 10: I'd leave these comments and scope out follow-up tickets. There is no point we have three (3) different encoders/decoders for the same thing.

The idea is that the same encoder/decoder that's used on the workflow side (doing the encodign) should be used on the WT side (DF processor doing the decoding).

pkg/report/pb/data-feeds/on-chain/registry/decode_test.go
```diff
@@ -1,240 +1,91 @@
...
-					BlockHeight:    "17",
-					BlockTimestamp: 0x66f5bf69,
-
-					TxSender:   "example-transmitter",
-					TxReceiver: "example-forwarder",
-				},
-			},
-			wantErr: false,
-		},
+func TestDecodeAsFeedUpdated(t *testing.T) {
```
krebernisak: Please compare left and right sides.
I personally don't understand how right is better than left.

pkg/report/platform/types.go
```diff
@@ -3,16 +3,15 @@ package platform
...
 )
 
 // Report represents an OCR3 report with metadata and data
 // version | workflow_execution_id | timestamp | don_id | config_version | ... | data
 type Report struct {
```
krebernisak: You changed this general (common) type to make it fit EVM, but common and Aptos used this type.
So are we now we're going to refactor Aptos, to use this new type introduced by EVM work?

pkg/writetarget/write_target_monitor.go
```diff
@@ -23,22 +22,21 @@ const (
 	schemaBasePath     = repoCLLCommon + "/" + versionRefsDevelop + "/pkg/capabilities/writetarget/pb"
 )
 
+type emitterAwareProcessor interface {
```
krebernisak: This type should be public

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -33,42 +29,18 @@ func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*FeedUpdated, error) {
...
-			if err != nil {
-				return nil, fmt.Errorf("failed to decode Mercury v%d report: %w", t, err)
-			}
-			// For Mercury v4, skip TxSender and TxReceiver (if not applicable)
-			msgs = append(msgs, newFeedUpdated(m, feedID, rm.ObservationsTimestamp, rm.BenchmarkPrice, rf.Data, false))
-		default:
-			return nil, fmt.Errorf("unsupported Mercury report type: %d", t)
-		}
+		// TODO: unsure if r.Data is correct for Report
+		msgs = append(msgs, NewFeedUpdated(m, feedID, rf.Timestamp, rf.Price, r.Data, false))
```
krebernisak: Nope. It should be `rf.Data` (report feed data).

`r.Data` is platform report data -> data which contains an encoded list of feed updates data

pkg/report/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: Aptos uses it as well https://github.com/smartcontractkit/chainlink-aptos/blob/hotfix/2.21.1-internal/relayer/monitoring/pb/data-feeds/on-chain/registry/decode.go#L52

pkg/report/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: But there is no rf.Data for the EVM type, it follows this schema https://github.com/smartcontractkit/chainlink-evm/blob/b18ae9cad0b175959fe5fdc2f6c849ff22e789a6/pkg/report/datafeeds/types.go#L36-L38

krebernisak: Ok, makes sense.

So for EVM we get a decoded report (part of it) and on Aptos we get a raw report that we decode ourselves (a full Mercury report - v3 or v4).

I'd say the correct way to express this is that we just didn't get a DF underlying raw report, so let's just attach an empty byte array there.

pkg/report/datafeeds/types.go
silaslenihan: https://smartcontract-it.atlassian.net/browse/BCFR-1234

pkg/report/pb/data-feeds/on-chain/registry/decode_test.go
silaslenihan: Brought back the old test and made it compatible with EVM

pkg/report/platform/types.go
silaslenihan: Yes Aptos will have to import from here until we generalize the type

krebernisak: I would have preferred the other way around as a more cleaner approach, but ok now.

Either way we'll need to follow up here.

pkg/report/pb/data-feeds/on-chain/registry/metrics.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // ns returns a namespaced metric name
 func ns(name string) string {
-	return fmt.Sprintf("datafeeds_on_chain_registry_%s", name)
+	return fmt.Sprintf("data_feeds_on_chain_registry_%s", name)
```
krebernisak: Not sure how this ended up being `datafeeds` in this PR, it was always `data_feeds`

pkg/report/datafeeds/types.go
```diff
@@ -19,26 +21,28 @@ type Reports = []FeedReport
...
 		}
 		return result
 	}
 
 	return abi.Arguments([]abi.Argument{
-		// TODO: why is the workflow encoder_config "(bytes32 FeedID, bytes RawReport)[] Reports"?
 		{
-			Type: mustNewType("tuple(bytes32, bytes)[]", "", []abi.ArgumentMarshaling{
+			// This defines the array of tuple records.
+			Type: mustNewType("tuple(bytes32,uint224,uint32)[]", "", []abi.ArgumentMarshaling{
```
krebernisak: This schema doesn't fit the above `FeedReport` type.

But also, how do you plan to support current two types of EVM schema's used (one for Celo, and another for all other EVM workflows): https://github.com/search?q=repo%3Asmartcontractkit%2Fchainlink-deployments%20encoder_config&type=code

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -33,42 +29,18 @@ func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*FeedUpdated, error) {
...
-			if err != nil {
-				return nil, fmt.Errorf("failed to decode Mercury v%d report: %w", t, err)
-			}
-			// For Mercury v4, skip TxSender and TxReceiver (if not applicable)
-			msgs = append(msgs, newFeedUpdated(m, feedID, rm.ObservationsTimestamp, rm.BenchmarkPrice, rf.Data, false))
-		default:
-			return nil, fmt.Errorf("unsupported Mercury report type: %d", t)
-		}
+		// TODO: unsure if r.Data is correct for Report
+		msgs = append(msgs, NewFeedUpdated(m, feedID, rf.Timestamp, rf.Price, []byte{}, false))
```
krebernisak:
```suggestion
		// Notice: this encoding of a DF report doesn't provide a raw underlying report
		rawReport := []byte{}
		msgs = append(msgs, NewFeedUpdated(m, feedID, rf.Timestamp, rf.Price, rawReport, false))
```

pkg/report/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -6,13 +6,9 @@ import (
 	"math/big"
```
krebernisak: I've said this multiple times during reviews, you're not structuring the code properly. But you're also not using existing hints/structure while moving code - is there a pattern I'm missing?

1. `report/pb` and `report/monitor` have nothing to do with reports so why are they in a `/report`
2. Also, why did components like `pkg/report/pb/platform/write_accepted.proto` lost the `write_target` subsection? It should be put in a specific folder `pkg/monitoring/pb/platform/write_target/write_accepted.proto` as this is it's specific domain.

silaslenihan: Sorry, I'm not familiar with the reasoning for the existing structure so the pattern is not obvious to me.

> have nothing to do with reports

I wouldn't say that, `report/pb` is for monitoring reports on-chian. But I understand your point, I will move them to the `pkg/monitoring` directory. 

There is already an existing `pkg/monitor` package which has some conflicting definitions relating to balance monitoring so I decided to move `report/monitor` to `pkg/monitor/beholder`, is that acceptable?

krebernisak: I'm sharing context and feedback.

Considering your deadlines, I'd suggest doing what makes sense to you to get this working, and then follow up and clean up the situation.

.github/workflows/solidity.yml
```diff
@@ -253,16 +253,16 @@ jobs:
...
           else
             echo "::error::No release version found."
             exit 1
           fi
-          PACKAGE_JSON_VERSION="$(jq -r '.version' package.json)"
-          if [[ "${PACKAGE_JSON_VERSION}" != "${VERSION}" ]]; then
-            echo "::error::Version mismatch: package.json version (${PACKAGE_JSON_VERSION}) does not match version computed from tag ${VERSION}"
+          package_json_version="$(jq -r '.version' package.json)"
+          if [[ "$PACKAGE_JSON_VERSION" != "${version}" ]]; then
+            echo "::error version mismatch: package.json version ($package_json_version) does not match version computed from tag ${version}"
```
erikburt: This seems like merge issues, I'd keep the version from default branch

https://github.com/smartcontractkit/chainlink-evm/blob/44ec1c3b5fc9ca269fc3620ecf429d4da42dafc9/.github/workflows/solidity.yml#L256-L265

pkg/report/por/processor/processor.go
```diff
@@ -0,0 +1,59 @@
...
+				return fmt.Errorf("failed to publish feed updated metrics: %w", err)
+			}
+		}
+		return nil
+	default:
+		return nil // fallthrough
+	}
+}
+
+func (p *porFeedsProcessor) SetEmitter(e monitor.ProtoEmitter) {
```
dhaidashenko: Why do we need this method?

pkg/config/toml/defaults/fallback.toml
```diff
@@ -98,3 +98,6 @@ GasLimit = 5400000
 
 [Workflow]
 GasLimitDefault = 400_000
+TxAcceptanceState = 3
```
dhaidashenko: It should be set to 2 (Unconfirmed, which means in case of EVM TXM that transaction was included into a block, but can be reorged). If on fast chains, it proves to be insufficient, we can override it with 3 (finalized).

Otherwise, we need to go through all chains where finality is slower than inter-don request timeout and set  TxAcceptanceState to 2. Second approach seems to be much more effort.

pkg/monitoring/pb/data-feeds/on-chain/registry/feed_updated.proto
```diff
@@ -10,19 +10,20 @@ message FeedUpdated {
   string feed_id = 1; // bytes as hex string for readability
   uint32 observations_timestamp = 2;
   bytes benchmark = 3;
-  bytes report = 4;
+  bytes bundle = 4;
```
dhaidashenko: Why do we need to introduce this field? Is not it a breaking change for the beholder?

silaslenihan: I introduced this to add support for PoR feedUpdated messages, since those include a bundle rather than a price. To avoid breaking changes, I shifted this to a previously unused index, 9

dhaidashenko (dismissed):

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -5,25 +5,22 @@ import (
...
-	"github.com/smartcontractkit/chainlink-evm/pkg/report/platform"
-
-	mercury_vX "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/common"
-	mercury_v3 "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/v3"
-	mercury_v4 "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/v4"
+	"github.com/smartcontractkit/chainlink-evm/pkg/report/por"
 )
 
-func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*FeedUpdated, error) {
+func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed, ccip bool) ([]*FeedUpdated, error) {
```
krebernisak: I'm don't particularly like this ccip bool flag on the input, and that these two fn are completely the same (`DecodeAsFeedUpdated` & `DecodePORAsFeedUpdated`), except the underlying report decoding step.

It would be best to consider extracting a higher-level fn out, and injecting just the underlying report decoding step.

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -89,28 +91,28 @@ func newFeedUpdated(
 		BlockTimestamp: m.BlockTimestamp,
 
 		// Execution Context - Source
-		MetaSourceId: m.MetaSourceId,
+		MetaSourceId: m.ExecutionContext.MetaSourceId,
```
krebernisak: Wait, how come we didn't extract the `ExecutionContext` in the FeedUpdated proto, like we did for WriteConfirmed. Would that be a follow up?

pkg/report/datafeeds/types.go
```diff
@@ -2,53 +2,95 @@ package datafeeds
...
+	Timestamp uint32
+	Price     *big.Int // *big.Int is used because go-ethereum converts large uints to *big.Int.
 }
 
-type Reports = []FeedReport
+type CCIPFeedReport struct {
+	FeedID    [32]byte
+	Price     *big.Int
+	Timestamp uint32
+}
```
krebernisak, line 23: It's not clear how this changed so drastically. Why did it change so drastically?

Before, we had a Mercury report (underlying) wrapped in a FeedReport (root), now there's no underlying report anymore?

pkg/monitoring/pb/data-feeds/on-chain/registry/feed_updated.proto
```diff
@@ -23,6 +23,8 @@ message FeedUpdated {
   string block_hash = 6;
```
krebernisak: Where do we plan to eventually move the `data-feeds`, specifically these `data-feeds.on-chain` protos? (these are common across chains)

pkg/report/datafeeds/processor/processor.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
+	"github.com/smartcontractkit/chainlink-evm/pkg/monitoring/pb/data-feeds/on-chain/registry"
+	wt "github.com/smartcontractkit/chainlink-framework/capabilities/writetarget/monitoring/pb/platform"
+)
+
+// EVM Data-Feeds specific processor decodes writes as 'data-feeds.registry.FeedUpdated' messages + metrics
+type dataFeedsProcessor struct {
+	emitter beholder.ProtoEmitter
+	metrics *registry.Metrics
+	ccip    bool
```
krebernisak: imo this just leaks CCIP details here in this type, and makes it handle a special case

vs keeping the type general, and injecting your special case logic with a general type (e.g., inject a specific report decoder for CCIP, vs another one for X)

krebernisak (dismissed):

pkg/config/config.go
```diff
@@ -188,6 +190,9 @@ type Workflow interface {
 	FromAddress() *types.EIP55Address
 	ForwarderAddress() *types.EIP55Address
 	GasLimitDefault() *uint64
+	TxAcceptanceState() *commontypes.TransactionStatus
+	PollPeriod() *commonconfig.Duration
+	AcceptanceTimeout() *commonconfig.Duration
```
jmank88, line 195: Life is much

jmank88: Life is much easier with std durations:
```suggestion
	PollPeriod() time.Duration
	AcceptanceTimeout() time.Duration
```
The config types are specialized for parsing, but don't need to be exposed through the config interfaces (though pointer is OK case by case).

pkg/config/toml/config.go
```diff
@@ -628,9 +628,12 @@ func (a *Automation) setFrom(f *Automation) {
...
 type Workflow struct {
-	FromAddress      *types.EIP55Address `toml:",omitempty"`
-	ForwarderAddress *types.EIP55Address `toml:",omitempty"`
-	GasLimitDefault  *uint64
+	FromAddress       *types.EIP55Address `toml:",omitempty"`
+	ForwarderAddress  *types.EIP55Address `toml:",omitempty"`
+	GasLimitDefault   *uint64
+	TxAcceptanceState *commontypes.TransactionStatus
+	PollPeriod        *commonconfig.Duration
+	AcceptanceTimeout *commonconfig.Duration
```
jmank88, line 637: Let's sort these alphabetically so they are easier to look up in the generated docs

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: I didn't think it was possible to import a proto from a separate repo with our repo structure, so I left it out in `FeedUpdated`. Or are you saying I should re-define ExecutionContext?

pkg/report/datafeeds/types.go
silaslenihan: That's correct, there is no underlying report for these types. Keep in mind these or for EVM. On Aptos there would a procesor that uses the structure you are describing. I left that out of scope for the current set of PRs though since we would not be using that any time soon on Aptos.

pkg/report/datafeeds/processor/processor.go
silaslenihan: Refactored the Processor logic quite a bit in [this commit](https://github.com/smartcontractkit/chainlink-evm/pull/24/commits/ba6e1bf5dc3be833a611507c5e1157b6ef466cb9) to be injectible and cleaner as you said

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
krebernisak: Please scope out a follow up for this - we should share the metadata proto definition across these events

pkg/report/datafeeds/processor/processor.go
krebernisak: That commit you mention is not available

silaslenihan: here you go, hash changed bc of amend sorry about that http://github.com/smartcontractkit/chainlink-evm/pull/24/commits/744075cf446f9482dd6218e418d0b865284c1a02

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: Followup ticket: https://smartcontract-it.atlassian.net/browse/PLEX-1492

pkg/monitoring/pb/data-feeds/on-chain/registry/feed_updated.proto
silaslenihan: Sorry, missed this before. I have just moved all the data-feeds protos to `chainlink-framework`

pkg/monitoring/pb/data-feeds/on-chain/registry/decode.go
silaslenihan: Closing this out since `FeedUpdated` now uses `ExecutionContext` in [chainlink-framework](https://github.com/smartcontractkit/chainlink-framework/blob/de26feab4ee7af1e098e030d518a0a6b99009457/capabilities/writetarget/monitoring/pb/data-feeds/on-chain/registry/feed_updated.proto#L39)

pkg/report/datafeeds/processor/processor.go
```diff
@@ -0,0 +1,94 @@
...
+func GetCCIPDataFeedsSchema() abi.Arguments {
+	return datafeeds.GetSchema("tuple(bytes32,uint224,uint32)[]", "", []abi.ArgumentMarshaling{
+		{Name: "feedID", Type: "bytes32"},
+		{Name: "price", Type: "uint224"},
+		{Name: "timestamp", Type: "uint32"},
+	})
+}
+
+// Decode is made available to external users (i.e. mercury server)
+func Decode(m *wt.WriteConfirmed, data []byte, schema abi.Arguments) ([]*df.FeedUpdated, error) {
```
krebernisak, line 46: Sorry, something is off here... This is underlying report decoding - it has nothing to do with our WT, with our messages, monitoring, anything on the higher levels. This is just low-level binding to a specific type and decoding in/out that type.

Please decouple this implementation from using outer types (e.g., `wt.WriteConfirmed`) and have the outer layer (WT, WT monitor) use the decoder.

pkg/config/toml/config.go
krebernisak, line 637: What are these configs? scoped to a workflow?

They don't seem to be used in this PR.

krebernisak: Aha, I see, they are used in the core PR: https://github.com/smartcontractkit/chainlink/pull/17216/files#diff-9b55aa19c69723ca59deb033266f910479defbc34e400b37e1c42a8897c3a373R127

Well, 2-3 things here:

- Configs are inconsistent (e.g., AcceptanceTimeout vs ConfirmerTimeout)
- Configs should be namespaced and scoped - this is not a Workflow config but a WT capability config (should be expressed and nested that way in toml)
- Configs should be composable (just import framework defined configs and use those, don't redefine and rename)

pkg/report/datafeeds/decode.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
+
+	"github.com/smartcontractkit/chainlink-framework/capabilities/writetarget/monitoring/pb/common"
+	df "github.com/smartcontractkit/chainlink-framework/capabilities/writetarget/monitoring/pb/data-feeds/on-chain/registry"
+	wt "github.com/smartcontractkit/chainlink-framework/capabilities/writetarget/monitoring/pb/platform"
+	"github.com/smartcontractkit/chainlink-framework/capabilities/writetarget/report/platform"
+)
+
+// EVM POR specific processor decodes writes as 'data-feeds.registry.FeedUpdated' messages + metrics
+type Processor struct {
```
krebernisak, line 21: This should also be extracted as part of the framework, nothing specific for EVM here - underlying payload report decoding can be injected (currently is)

pkg/report/datafeeds/processor/processor.go
silaslenihan: followup ticket: https://smartcontract-it.atlassian.net/browse/PLEX-1535

pkg/config/toml/config.go
silaslenihan: Fixed first bullet point, scoped out ticket for following two: https://smartcontract-it.atlassian.net/browse/PLEX-1536

pkg/report/datafeeds/decode.go
silaslenihan: followup ticket: https://smartcontract-it.atlassian.net/browse/PLEX-1537

dhaidashenko (dismissed):

krebernisak (dismissed):

krebernisak (approved):

dhaidashenko (approved):

jmank88 (approved):

PR #27, opened 2025-04-14
Use framework metrics

DylanTinianov:
### Description
Align the following metrics across chain families:
- Logpoller Metrics
- Node Balance Metrics
- RPC Latency Metrics

Epic: https://smartcontract-it.atlassian.net/browse/BCFR-1205

### Requires Dependencies
https://github.com/smartcontractkit/chainlink-framework/pull/38

### DIFF
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -25,6 +25,7 @@ import (
 
 	commonassets "github.com/smartcontractkit/chainlink-common/pkg/assets"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-framework/metrics"
 	"github.com/smartcontractkit/chainlink-framework/multinode"
 
 	"github.com/smartcontractkit/chainlink-evm/pkg/assets"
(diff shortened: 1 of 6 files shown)


pkg/client/rpc_client.go
```diff
@@ -263,6 +265,19 @@ func (r *RPCClient) logResult(
 			append(results, "err", err)...,
 		)
 	}
+
+	metrics.RPCCallLatency.
+		WithLabelValues(
+			"evm",                          // chain family
```
mateusz-sekara: Shouldn't we use some sort of enums/consts for chain families (ideally a single source of truth) across the entire codebase?

pkg/client/rpc_client.go
```diff
@@ -263,6 +265,19 @@ func (r *RPCClient) logResult(
...
+			metrics.EVM,                    // chain family
+			r.chainID.String(),             // chain id
+			rpcDomain,                      // rpc url
+			"false",                        // is send only
+			strconv.FormatBool(err == nil), // is successful
+			callName,                       // rpc call name
+		).
+		Observe(float64(callDuration))
+
+	// TODO: Remove deprecated metric
```
Farber98, line 280: just curious, why marking as deprecated and leaving a TODO? some compatibility issue?

pkg/monitor/balance.go
```diff
@@ -133,6 +135,7 @@ func (bm *balanceMonitor) promUpdateEthBalance(balance *assets.Eth, from common.
 		return
 	}
 
+	metrics.NodeBalance.WithLabelValues(from.Hex(), bm.chainIDStr, metrics.EVM).Set(balanceFloat)
```
Farber98, line 138: should we leave a TODO to remove this deprecated metric as well?

pkg/client/rpc_client.go
DylanTinianov: We need to keep the old metrics to not break any dashboards people may have for backwards compatibility, but eventually everyone should move to the chain-agnostic ones.

Farber98 (approved):

reductionista (approved):

PR #38, opened 2025-04-24
Fix/log index rsk hedera

simsonraj:
Added Hedera & Rootstock to the list of Log Index fix where these chains, do not have unique logIndex in a block


dhaidashenko (commented):
The only difference between existing tests and newly added is the chain type that we use. It should be possible to iterate through chain types that require special handling and run tests for each of them, instead of duplicating code

jmank88 (dismissed):

dhaidashenko (approved):

jmank88 (approved):

PR #40, opened 2025-04-24
Initial work for workflow registry v2: limits

eutopian:


contracts/gas-snapshots/workflow.gas-snapshot
```diff
@@ -56,6 +56,33 @@ WorkflowRegistry_registerWorkflow:test_WhenTheWorkflowInputsAreAllValid() (gas:
 WorkflowRegistry_requestForceUpdateSecrets:test_WhenTheCallerIsAnAuthorizedAddress_AndTheWorkflowIsInAnAllowedDON() (gas: 936092)
 WorkflowRegistry_requestForceUpdateSecrets:test_WhenTheCallerIsAnAuthorizedAddress_AndTheWorkflowIsNotInAnAllowedDON() (gas: 510822)
 WorkflowRegistry_requestForceUpdateSecrets:test_WhenTheCallerIsNotAnAuthorizedAddress() (gas: 509176)
+WorkflowRegistry_setDONOverride:test_WhenCallerIsNOTTheOwner() (gas: 120)
```
eutopian: not sure if we should set up a new profile for this and possibly also use pragma 0.8.26 if we do, especially as there will be duplicate functions when it comes to things like register

ibrajer: For VRF, we set up a different profile for each version to benefit from having a newer Solidity version, so it's certainly possible. In this case, the Foundry profile could be called "workflow_v2".

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
```diff
@@ -0,0 +1,172 @@
...
+  // |                        Limits Config                         |
+  // ================================================================
+
+  /// @dev Instead of a big struct with mappings, we store
+  ///      defaults in a single 32-byte slot, and use nested mappings
+  ///      for overrides.
+  struct Config {
+    // Pack three uint32 defaults into a single 96-bit field:
+    // Layout in bits: [maxUser(0..31) | maxDON(32..63) | maxUserDON(64..95)]
+    uint96 defaultsPacked;
```
ibrajer: Why pack it like this instead of specifying three separate uint32 fields that will also fit into a single storage slot (uint32 = 4 bytes, storage slot size uint256 = 32 bytes, total of 12 bytes in the slot, so you still have some space left)?

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
```diff
@@ -0,0 +1,172 @@
...
+  function setDefaults(uint32 maxPerUser, uint32 maxPerDON, uint32 maxPerUserDON) external onlyOwner {
+    s_cfg.defaultsPacked = _packDefaults(maxPerUser, maxPerDON, maxPerUserDON);
+  }
+
+  /// @notice Override the maximum # of workflows a specific user can register.
+  function setUserOverride(address user, uint32 limit, bool isSet) external onlyOwner {
+    if (isSet) {
+      s_cfg.userOverride[user] = Config.Value(limit, true);
+    } else {
+      delete s_cfg.userOverride[user];
```
ibrajer: Not 100% sure on this, but generally, deleting elements in mappings/arrays is avoided due to gas costs, so resetting values is more desirable. Since you're already using this `isSet` field value, maybe you can just set it `false`? In that case, this code is as simple as:
```
s_cfg.userOverride[user] = Config.Value(limit, isSet);
```

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
```diff
@@ -0,0 +1,172 @@
...
+  ///      defaults in a single 32-byte slot, and use nested mappings
+  ///      for overrides.
+  struct Config {
+    // Pack three uint32 defaults into a single 96-bit field:
+    // Layout in bits: [maxUser(0..31) | maxDON(32..63) | maxUserDON(64..95)]
+    uint96 defaultsPacked;
+    // Struct to distinguish between unset and explicitly set zero values
+    struct Value {
+      uint32 value;
+      bool isSet;
```
ibrajer: You could also use the `uint32` max value or any other arbitrarily large value to have a special meaning, but this is certainly easier to understand!

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
```diff
@@ -0,0 +1,172 @@
...
+  function setDONOverride(bytes32 donLabel, uint32 limit, bool isSet) external onlyOwner {
+    if (isSet) {
+      s_cfg.donOverride[donLabel] = Config.Value(limit, true);
+    } else {
+      delete s_cfg.donOverride[donLabel];
+    }
+  }
+
+  /// @notice Override the max # of workflows for a specific (user, DON) pair.
+  function setUserDONOverride(address user, bytes32 donLabel, uint32 limit, bool isSet) external onlyOwner {
```
ibrajer: Just thinking out loud... The parameter `isSet` can be misleading because what does it mean? Set the user override? You are already calling this function, so of course, you are setting the user override. Maybe we can use `isEnabled` instead? Or something else that would indicate "I revoke permissions for this user on this DON to register any workflows". Wdyt?

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
eutopian: its one less SSTORE when you write. we need to check it often so it saves a little. Even though storage slot is packed together accessing them as separate fields is still separate...

maybe it's not worth it :D

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
eutopian: delete is only costly for large structures, but this is one slot size, so delete versus setting this to false cost the same storage write but has the slight benefit of the cleared slot refund. the only benefit potentially in allowing it is record I think.

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
eutopian: This looks like it's more complicated but actually not much more as both are small and fits into one storage slot, but much more explicit and readable than using an arbitrary value to get around the falsey zeros

contracts/src/v0.8/workflow/dev/WorkflowRegistry.sol
eutopian: although... if we have them as separate, we might need separate setters too, otherwise, there's not much gained, and I don't know if this is worth having different setters. otherwise, it's also not worth doing 
config.user = a
config.don = b
config.userdon = c each time as that's 3 writes

ibrajer (approved):

shileiwill (approved):

PR #41, opened 2025-04-25
Add contracts for ZK Sync

ilanolkies:
For testing:
- Weth9
- Mocks

For production:
- Capabilities registry 1.1.0, copied from Capabilities registry not indexed


contracts/scripts/zksync_compile_all
```diff
@@ -24,3 +24,7 @@ compileContract shared/token/ERC677/LinkToken shared "--use 0.8.19"
 compileContract shared/token/ERC677/BurnMintERC677 shared "--use 0.8.19"
 compileContract vendor/multicall/ebd8b64/src/Multicall3 shared "--use 0.8.19"
 compileContract keystone/CapabilitiesRegistry keystone
+
+compileContract vendor/canonical-weth-zk/WETH9 shared "--use 0.8.19"
+compileContract shared/mocks/MockV3Aggregator shared "--use 0.8.19"
+compileContract automation/testhelpers/MockETHUSDAggregator automation
```
ilanolkies:
```suggestion
compileContract automation/testhelpers/MockETHUSDAggregator automation

```

contracts/scripts/zksync_compile_all
```diff
@@ -24,3 +24,7 @@ compileContract shared/token/ERC677/LinkToken shared "--use 0.8.19"
 compileContract shared/token/ERC677/BurnMintERC677 shared "--use 0.8.19"
 compileContract vendor/multicall/ebd8b64/src/Multicall3 shared "--use 0.8.19"
 compileContract keystone/CapabilitiesRegistry keystone
+
+compileContract vendor/canonical-weth-zk/WETH9 shared "--use 0.8.19"
```
RensR: canonical-weth-zk should be the repo name for anything in vendor. This looks like we made it up?

contracts/src/v0.8/vendor/canonical-weth-zk/WETH9.sol
```diff
@@ -0,0 +1,86 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+pragma solidity ^0.8.0;
+
+// solhint-disable
+contract WETH9 {
```
RensR: Can we rename this to include ZKSync to avoid people using the wrong one?

contracts/src/v0.8/shared/canonical-weth-zk/WETH9ZKSync.sol
```diff
@@ -0,0 +1,87 @@
+// Submitted for verification at Etherscan.io on 2017-12-12
```
RensR: Please put it in an appropriate folder. We have `token/erc20` which should be used, you can create a folder `zksync` inside with this contract.

contracts/src/v0.8/shared/canonical-weth-zk/WETH9ZKSync.sol
```diff
@@ -0,0 +1,87 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+pragma solidity ^0.8.0;
+
+// solhint-disable
+contract WETH9ZKSync {
```
RensR: Let's add a comment on what was changed from the original version. It is clearly based on the Dapphub contract as is visible in the license above.

gethwrappers/generation/zksync/wrap.go
```diff
@@ -14,6 +14,11 @@ func main() {
 	contractName := os.Args[2]
 	packageName := os.Args[3]
 
+	className := contractName
+	if len(os.Args) > 4 {
```
RensR: Why do we need this?

ilanolkies: We are using the contract name as the class name, similar to the EVM generation. In this case we want to put this code inside the `WETH9` class but the contract was renamed

RensR: Why do you want to put the code inside the WETH9 class? My comment about renaming was also meant to ensure the offchain code is not confusing by having the same names

ilanolkies: For simplicity. The contract does virtually the same and it is only used for testing. The wrapper for ZK adds the deploy function to existent classes.

I can craft a new wrapper for this specific case: new package, new class name, but using the ABI of the existent `WETH9` class. That would be simple to consume from `chainlink/deployment` too. What do you think?

RensR: If there's a single byte that's different, it's a different contract. Renaming files like this makes it harder to find the source code, as that is now names differently. I think it's better for consistency to not rename anything

RensR (dismissed):

gethwrappers/keystone/generated/capabilities_registry_1_1_0/capabilities_registry_zksync.go
```diff
@@ -0,0 +1,61 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+package capabilities_registry
```
krehermann: help me understand what is happening here please.

this contract is a singleton; across all chains there is only one instance and it will never be on zksync

from the point of view of the CRE platform, the need is that keystone forwarder contract can be deployed to zksync. it is the forwarder contract that exists on all chains in the CRE ecosystem, not this capability registry contract

ilanolkies: After discussing we cleared out the reason why the `CapabilitiesRegistry` was created: testing purposes. 

As it shouldn't be necessary to deploy this in production, I'm going to review the tests and improve them so that they simulate what a production deployment would be like. After reviewing that I am going to add the necessary bindings, such as `KeystoneForwarder`

For now, we are removing this contract from this PR and let the others in.

krehermann (dismissed):

RensR (approved):

simsonraj (approved):

krehermann (approved):

bolekk (approved):

PR #51, opened 2025-05-02
Return Transaction Fee with Capability Response

EasterTheBunny:
This commit links the transaction fee data return by an EVMService to a `CapabilityResponse` within the write target. The intention is to surface transaction costs for billing purposes.

Additionally, this commit adds tool setup scripts streamlined with `asdf`.

### DIFF
--- a/pkg/writetarget/write_target.go
+++ b/pkg/writetarget/write_target.go
@@ -358,6 +358,7 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 
 	fee, err := c.evm.GetTransactionFee(ctx, txID)
 	if err != nil {
+		fee = &commontypes.TransactionFee{TransactionFee: new(big.Int)}
 		c.lggr.Errorw("failed to get transaction fee for transaction id", "err", err, "txid", txID)
 	}
 
--- a/pkg/writetarget/write_target.go
+++ b/pkg/writetarget/write_target.go
@@ -7,11 +7,13 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math/big"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/assets"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
@@ -87,6 +89,7 @@ type writeTarget struct {
 
 	nodeAddress      string
 	forwarderAddress string
+	p2pID            string
 
 	targetStrategy TargetStrategy
 }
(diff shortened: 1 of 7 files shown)


pkg/writetarget/write_target.go
```diff
@@ -204,7 +228,7 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 	receiver, err := c.configValidateFn(request)
 	if err != nil {
 		msg := builder.buildWriteError(info, 0, "failed to validate config", err.Error())
-		return capabilities.CapabilityResponse{}, c.asEmittedError(ctx, msg)
+		return failure(), c.asEmittedError(ctx, msg)
```
patrickhuie19: /nit - but while failure isn't doing anything interesting, it'll save the reader some trouble by keeping it to an empty CapabilityResponse struct

pkg/writetarget/write_target.go
```diff
@@ -241,7 +265,7 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 		if err != nil {
 			c.lggr.Errorw("failed to emit write skipped", "err", err)
 		}
-		return success(), nil
+		return success("", nil), nil
```
patrickhuie19: /nit - but i think the same as my previous comment applies here. I think the reader will benefit to just see that an empty capability response is returned

pkg/writetarget/write_target.go
```diff
@@ -336,7 +360,13 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 	// relevant to this forwarder), and emit write-tx-accepted/confirmed events.
 
 	go c.acceptAndConfirmWrite(ctx, *info, txID)
-	return success(), nil
+
+	fee, err := c.evm.GetTransactionFee(ctx, txID)
```
patrickhuie19, line 359: Above this line, a go routine is kicked off that looks to confirm the write. `GetTransactionFee` will return an error unless the transaction status is finalized, so I think we'll fail to get the transaction fee here for the node that successfully writes on chain. 

The approach in this PR will work only if we sync wait to confirm the write - cc @Unheilbar

pkg/writetarget/write_target.go
```diff
@@ -160,14 +163,35 @@ func NewWriteTarget(opts WriteTargetOpts) capabilities.TargetCapability {
...
 	return capabilities.CapabilityResponse{}
 }
 
+func success(p2pID string, transactionFee *big.Int) capabilities.CapabilityResponse {
+	var detail []capabilities.MeteringNodeDetail
+
+	if transactionFee != nil {
+		detail = []capabilities.MeteringNodeDetail{
+			{
+				Peer2PeerID: p2pID,
```
patrickhuie19: `opts.ID` isn't the p2pID. We also don't need to set this value here, we set it in the Don2Don layer on the engine

pkg/writetarget/write_target.go
Unheilbar: Confirming that GetTransactionFee in this case will return error, unless we wait for Tx being finalized. For now as far as I know that changes introduced[here](https://github.com/smartcontractkit/chainlink-aptos/pull/139/files#diff-0481a986484667a3a9788d2e150972d5980ffe9d27df30c5c3e5acbcd0c9fe95) are planned on migrating to EVM as well

EasterTheBunny: For now, in the case of an error, we could set the fee to zero until these new changes get wired up.

patrickhuie19: We may need a new API - GetTransactionFee was designed around finality. The latency of that isn't acceptable for us here, so we either need to accept the fee on inclusion, or use an estimation.

PR #63, opened 2025-05-14
[CR-403] Move txmgr into chainlink-evm

pavel-raykov:
Supports https://github.com/smartcontractkit/chainlink/pull/17736


mchain0 (approved):

mchain0:
@pavel-raykov most of this code seems a "move" but committed as new code, I recommend for consideration moving the code with the history preserved, like e.g.: https://medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b That's not always easy (or even possible), but worth trying due to a number of significant benefits.

pavel-raykov:
> @pavel-raykov most of this code seems a "move" but committed as new code, I recommend for consideration moving the code with the history preserved, like e.g.: [medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b](https://medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b) That's not always easy (or even possible), but worth trying due to a number of significant benefits.

Yes, ideally, the files would be moved like that. Unfortunately, there are two impediments for doing that: 
1. One needs to manually rebase PRs touching non-moved files
2. One needs to resign every PR by manually pressing the key (think about doing this > 1000 times =( )

So, we just move the code as it is currently.

jmank88 (approved):

PR #67, opened 2025-05-20
DEVSVCS-1806: Add new foundry profile for workflow registry v2

eutopian:


gethwrappers/generation/wrap.go
```diff
@@ -21,7 +22,7 @@ func main() {
 	}
 
 	// Once vrf is moved to its own subfolder we can delete this rootDir override.
-	if project == "vrf" || project == "automation" {
+	if project == "vrf" || project == "automation" || strings.Contains(project, "workflow") {
```
pavel-raykov: please keep using "workflow" as the project. You can use ugly syntax like https://github.com/smartcontractkit/chainlink-evm/blob/63559812f68891f67e42d633c7c15be1b65c3b82/gethwrappers/go_generate_automation.go#L3 in go_genereate directive to compile it.

gethwrappers/go_generate.go
```diff
@@ -15,4 +15,5 @@ package gethwrappers
 //go:generate go generate ./llo-feeds
 //go:generate go generate ./operatorforwarder
 //go:generate go generate ./shared
-//go:generate go generate ./workflow
+//go:generate go run ./generation/wrap.go workflow/v1 WorkflowRegistry workflow_registry_wrapper_v1
```
pavel-raykov: sorry, why did you decide to start using go run directive, while we use go generate everywhere else?

eutopian: it's used here too https://github.com/smartcontractkit/chainlink-evm/pull/67/files/b0fa992bc3e8e2929d7f90eb21bfc2a7ec505feb#diff-645df94cf159337a986efc1fbb98a250e83dbbddecff87e581203b96343a0fbbR8 

right above these lines in the same file

gethwrappers/generation/wrap.go
eutopian: yeah... we'd have to name the contract ugly too :-/

gethwrappers/go_generate.go
pavel-raykov: sorry, you are right - I am just proposing to be consistent

OffchainAggregator and wrap_link are really the special cases.

Can you switch to go generate here?

gethwrappers/generation/wrap.go
pavel-raykov: I know, but right now there are two ways to generate the contract and this PR would basically introduced a 3-rd special way where the project string only !contains! the project, this is yet another confusing step. Would it be possible to use the current ugly way I referred to?

eutopian: hmm... I'm not sure if renaming the existing v1 contract code is a good idea since it requires changing the class name of a production contract. Let me play around with the dir maybe we don't even need this

gethwrappers/go_generate.go
eutopian: I think that actually fixed our problem above as well :D

gethwrappers/workflow/go_generate.go
```diff
@@ -4,4 +4,5 @@ package gethwrappers
 
 // Workflow
 
-//go:generate go run ../generation/wrap.go workflow WorkflowRegistry workflow_registry_wrapper
+//go:generate go run ../generation/wrap.go workflow/v1 WorkflowRegistry workflow_registry_wrapper_v1
+//go:generate go run ../generation/wrap.go workflow/v2/dev WorkflowRegistry workflow_registry_wrapper_v2
```
pavel-raykov: sorry, missed that - is there any reason you chose workflow/${version}/dev structure here?

Right now we already have automation (https://github.com/smartcontractkit/chainlink-evm/tree/develop/contracts/src/v0.8/automation) with: automation/${version} and automation/dev

and functions (https://github.com/smartcontractkit/chainlink-evm/tree/develop/contracts/src/v0.8/functions) with functions/${version} and functions/dev/${version},

so it seems that this PR introduces a 3rd way for versioning here.

eutopian: I guess there was no reason other than, the original was just workflow/dev pre audit, and was moved to workflow post audit, and that everything pre audit should be in dev... maybe due to the profile paths.

it was just assumed we'd moved /dev out into /v2 post audit similar to v1

i'm fine with /dev/v2

pavel-raykov: Good, let's use /dev/v2 to avoid another versioning pattern.

eutopian: done. maybe the question is why dev still exist in automation or functions (especially if no one is actually developing it)... because post audit once something is done we move it out so it shouldn't be a lingering folder :D

pavel-raykov (dismissed):

contracts/foundry.toml
```diff
@@ -92,11 +92,23 @@ src = 'src/v0.8/operatorforwarder'
 test = 'src/v0.8/operatorforwarder/test'
 
 [profile.workflow]
```
RensR, line 94: You could consider a separate compile profile that is used for wrapper generation. That one then uses via-ir while the main one devs use during development uses fast compilation. It can also have low optimizer runs to speed it up further. This is what we do with CCIP as compiling with via-ir takes ages. 

What you currently have is totally fine as well!

.github/workflows/solidity-foundry.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             { "name": "operatorforwarder", "setup": { "run-coverage": true, "min-coverage": 55.7, "run-gas-snapshot": true, "run-forge-fmt": false }},
             { "name": "shared", "setup": { "run-coverage": true, "extra-coverage-params": "--no-match-path='*CallWithExactGas*' --ir-minimum", "min-coverage": 32.6, "run-gas-snapshot": true, "run-forge-fmt": false }},
             { "name": "vrf", "setup": { "run-coverage": false, "min-coverage": 98.5, "run-gas-snapshot": false, "run-forge-fmt": false }},
-            { "name": "workflow", "setup": { "run-coverage": true, "extra-coverage-params": "--ir-minimum", "min-coverage": 96.0, "run-gas-snapshot": true, "run-forge-fmt": true }},
+            { "name": "workflow", "setup": { "run-coverage": true, "extra-coverage-params": "--ir-minimum", "min-coverage": 96.0, "run-gas-snapshot": true, "run-forge-fmt": true, "via-ir": true }},
```
RensR: you shouldn't need this as it usus the foundry profile that already specifies via-ir

contracts/scripts/native_solc_compile_all_workflow
```diff
@@ -12,16 +12,17 @@ CONTRACTS_DIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; cd ../ && pwd -P )"
...
-  command="forge build $CONTRACTS_DIR/src/v0.8/$PROJECT/"$1.sol" \
-   --root $CONTRACTS_DIR \
-   --extra-output-files bin abi \
-   -o $CONTRACTS_DIR/solc/$PROJECT/$contract"
-  $command
+  env FOUNDRY_PROFILE="$profile" forge build \
+  "$CONTRACTS_DIR/src/v0.8/$PROJECT/$src_rel.sol" \
+  --root "$CONTRACTS_DIR" \
+  --extra-output-files bin abi \
+  -o "$CONTRACTS_DIR/solc/$PROJECT/$src_rel"
```
RensR, line 24: note that this used to be $contract, not $src_rel. Is this intentional?

eutopian: yup! intentional as I moved it to sub folders and want the compiled code to be separate as well

.github/workflows/solidity-foundry.yml
eutopian: I imagine I shouldn't either, saw it in a test failure at some point but maybe it was picking up the wrong profile at the time

contracts/foundry.toml
eutopian: :D valid suggestions, though, this contract is much smaller. Recent ran got `Solc 0.8.24 finished in 1.86s` so I guess that might shave off a second

ibrajer (dismissed):

pavel-raykov (dismissed):

contracts/foundry.toml
RensR: You will develop more, including more tests which will add a lot of time in the future, right? Your choice either way

RensR:
To fix HardHat you might want to exclude it in HH config 

`contracts/hardhat.config.ts`

```
// prune forge style tests from hardhat paths
subtask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS).setAction(
  async (_, __, runSuper) => {
    const paths = await runSuper()
    const noTests = paths.filter((p: string) => !p.endsWith('.t.sol'))
    return noTests.filter(
      (p: string) => !p.includes('src/v0.8/vendor/forge-std'),
    )
  },
)
```
It would probably make sense to move this to an include list instead of an exclude list. Could you do that in this PR? It should only the paths for projects that currently still use HH

contracts/foundry.toml
eutopian: Is this still applicable though if we need the tag due to too many local vars in one of the function

eutopian:
@RensR updated! all checks passing now as well

pavel-raykov (approved):

contracts/hardhat.config.ts
```diff
@@ -16,14 +17,18 @@ const COMPILER_SETTINGS = {
   },
 }
 
+const EXCLUDE_DIRS = ['src/v0.8/vendor/forge-std', 'src/v0.8/workflow']
```
RensR, line 20: Clean!

RensR (approved):

ibrajer (approved):

PR #73, opened 2025-05-26
Linking workflow owners in Workflow Registry v2

ibrajer:
This PR adds a set of new features to the Workflow Registry v2 that allow the workflow owner to link their off-chain account by submitting ownership proof. Upon successful verification on the Workflow Registry side, if the proof is valid and if it originated from the trusted signing entity, it will be accepted and the owner address will be linked. Once the owner's address is linked, it means this owner is allowed to register and update workflows. However, if the linking is not completed, the owner's address will not be able to interact with the Workflow Registry contract at all.

Jira tickets: [DEVSVCS-1663](https://smartcontract-it.atlassian.net/browse/DEVSVCS-1663), [DEVSVCS-1664](https://smartcontract-it.atlassian.net/browse/DEVSVCS-1664), [DEVSVCS-1665](https://smartcontract-it.atlassian.net/browse/DEVSVCS-1665)

TODOs:
- [ ] Removal of workflows on workflow owner unlink, once the delete function is available
- [ ] Unit tests for the unlink operation that removes workflows

[DEVSVCS-1663]: https://smartcontract-it.atlassian.net/browse/DEVSVCS-1663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEVSVCS-1664]: https://smartcontract-it.atlassian.net/browse/DEVSVCS-1664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEVSVCS-1665]: https://smartcontract-it.atlassian.net/browse/DEVSVCS-1665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+
+  mapping(address signer => bool allowed) private s_allowedSigners;
+  mapping(address owner => bytes32 proof) private s_ownersProofs;
+  EnumerableSet.AddressSet private s_ownersProofsSet;
+
+  event AllowedSignersUpdatedV1(address[] signers, bool allowed);
+  event OwnershipProofSubmittedV1(address indexed owner, bytes32 indexed proof);
+  event OwnershipProofRevokedV1(address indexed owner, bytes32 indexed proof);
+
+  error RequestExpired(address caller, uint96 validityTimestamp);
```
DeividasK: Nit:
```suggestion
  error RequestExpired(address caller, uint96 currentTime, uint96 expiryTimestamp);
```

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+abstract contract OwnershipProof is Ownable2StepMsgSender {
+  using EnumerableSet for EnumerableSet.AddressSet;
+
+  mapping(address signer => bool allowed) private s_allowedSigners;
+  mapping(address owner => bytes32 proof) private s_ownersProofs;
+  EnumerableSet.AddressSet private s_ownersProofsSet;
+
+  event AllowedSignersUpdatedV1(address[] signers, bool allowed);
+  event OwnershipProofSubmittedV1(address indexed owner, bytes32 indexed proof);
+  event OwnershipProofRevokedV1(address indexed owner, bytes32 indexed proof);
```
DeividasK: Nit:
```suggestion
  event SignersUpdated(address[] signers, bool added);
  event OwnershipLinkUpdated(address indexed owner, bytes32 indexed proof, bool added);
```

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+
+  error RequestExpired(address caller, uint96 validityTimestamp);
+  error OwnershipProofAlreadySubmitted(address owner);
```
DeividasK: General naming comment: I would focus on owner links and not proofs

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// If the transaction sender is truly the owner of the claimed address, the proof will be acccepted if the
+  /// signature is valid and recoverable signer matches any of the allowed signers.
+  function submitOwnershipProof(uint96 validityTimestamp, bytes32 proof, bytes calldata signature) external {
+    if (uint96(block.timestamp) > validityTimestamp) {
+      revert RequestExpired(msg.sender, validityTimestamp);
+    }
+
+    if (s_ownersProofs[msg.sender] != bytes32(0)) {
+      revert OwnershipProofAlreadySubmitted(msg.sender);
+    }
```
DeividasK: Note: This prevents a 1-step transfer of wallet to a different org.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /// @notice Returns the signer of the recovered signature or revert.
+  /// @param validityTimestamp The validity timestamp of the ownership proof.
+  /// @param proof The ownership proof.
+  /// @param signature The signature of the ownership proof metadata.
+  /// @return The signer of the recovered signature.
+  /// @dev The function tries to re-generate the message digest based on the provided parameters and by following
+  /// EIP-191. The it will try to recover the signer address. The function will revert if the signature is invalid.
+  function _recoverSigner(
```
DeividasK: Use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// @param validityTimestamp Validity of the ownership proof.
+  /// @param proof The ownership proof to be submitted.
+  /// @param signature The signature of the ownership proof metadata.
+  /// @dev The ownership proof is a hash of the ownership proof metadata, which is signed by the allowed signer.
+  /// The ownership proof metadata is a combination of the claimed owner address, validity timestamp, and the proof hash.
+  /// Request will be rejected if the validity timestamp has expired or if the proof hasn't been submitted by this
+  /// transaction sender (can't revoke someone else's proof).
+  /// If the transaction sender is truly the owner of the claimed address, the proof will be revoked if the
+  /// signature is valid and recoverable signer matches any of the allowed signers.
+  function revokeOwnershipProof(uint96 validityTimestamp, bytes32 proof, bytes calldata signature) external {
```
DeividasK: Should check for active workflows and error if any are found

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+
+/// @title OwnershipProof
+/// @notice This contract allows the transaction sender to submit or revoke ownership proof.
+/// THIS IS A WORK IN PROGRESS AND IS NOT READY FOR REVIEWS YET.
+abstract contract OwnershipProof is Ownable2StepMsgSender {
+  using EnumerableSet for EnumerableSet.AddressSet;
+
+  mapping(address signer => bool allowed) private s_allowedSigners;
+  mapping(address owner => bytes32 proof) private s_ownersProofs;
+  EnumerableSet.AddressSet private s_ownersProofsSet;
```
DeividasK: A set is sufficient

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// @param batchSize The size of the batch to return (be vary of using a reasonable value here)
+  /// @return A list of ownership proofs for the given range
+  /// @dev The function returns a list of ownership proofs for the given range of stored owners.
+  /// The function will return a list of addresses of owners who have submitted ownership proofs in the
+  /// order in which they were submitted.
+  /// The function will return an empty list if the start index is greater than the total number of owners.
+  /// @notice List of owners may change between subsequent calls to this function. It is recommend to call this
+  /// function by anchoring a specific block number to ensure the list of owners is immutable between subsequent calls.
+  function getOwnersWithSubmittedProof(uint256 start, uint256 batchSize) external view returns (address[] memory) {
+    uint256 total = s_ownersProofsSet.length();
```
DeividasK:
```suggestion
    return s_ownersProofsSet.values();
```

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
+// SPDX-License-Identifier: BUSL 1.1
+pragma solidity 0.8.24;
+
+import {Ownable2StepMsgSender} from "../../shared/access/Ownable2StepMsgSender.sol";
+import {EnumerableSet} from "../../vendor/openzeppelin-solidity/v5.0.2/contracts/utils/structs/EnumerableSet.sol";
```
RayXpub:
```suggestion

import {EnumerableSet} from "../../vendor/openzeppelin-solidity/v5.0.2/contracts/utils/structs/EnumerableSet.sol";
```

RayXpub: https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/STYLE_GUIDE.md#imports

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
RayXpub: +1

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+
+/// @title OwnershipProof
+/// @notice This contract allows the transaction sender to submit or revoke ownership proof.
+/// THIS IS A WORK IN PROGRESS AND IS NOT READY FOR REVIEWS YET.
+abstract contract OwnershipProof is Ownable2StepMsgSender {
+  using EnumerableSet for EnumerableSet.AddressSet;
+
+  mapping(address signer => bool allowed) private s_allowedSigners;
```
RayXpub: suggestion: we could also use a set here, this would enable exposing enumeration through a getter function.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /// @notice Sets the allowed signers for ownership proofs.
+  /// @param signers The addresses of the signers
+  /// @param allowed The boolean value indicating whether the signer is allowed or not
+  /// @dev Ownership proofs can only be signed by approved group of signers.
+  /// When submitting signed proof to this contract, if recovered signature doesn't match any of the signers,
+  /// it will be rejected.
+  function updateAllowedSigners(address[] calldata signers, bool allowed) external onlyOwner {
+    for (uint256 i = 0; i < signers.length; i++) {
```
RayXpub:
```suggestion
    for (uint256 i = 0; i < signers.length; ++i) {
```

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// @return The signer of the recovered signature.
+  /// @dev The function tries to re-generate the message digest based on the provided parameters and by following
+  /// EIP-191. The it will try to recover the signer address. The function will revert if the signature is invalid.
+  function _recoverSigner(
+    uint96 validityTimestamp,
+    bytes32 proof,
+    bytes calldata signature
+  ) internal view returns (address) {
+    // Follow EIP-191 for recoverable signatures
+    bytes32 messageHash = keccak256(abi.encodePacked(msg.sender, block.chainid, validityTimestamp, proof));
```
RayXpub: suggestion: we might want to consider adding the contract address here to prevent signature replay across different contract versions.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// @param validityTimestamp Validity of the ownership proof.
+  /// @param proof The ownership proof to be submitted.
+  /// @param signature The signature of the ownership proof metadata.
+  /// @dev The ownership proof is a hash of the ownership proof metadata, which is signed by the allowed signer.
+  /// The ownership proof metadata is a combination of the claimed owner address, validity timestamp, and the proof hash.
+  /// Request will be rejected if the validity timestamp has expired or if the proof has already been submitted (by this
+  /// transaction sender or a different one).
+  /// If the transaction sender is truly the owner of the claimed address, the proof will be acccepted if the
+  /// signature is valid and recoverable signer matches any of the allowed signers.
+  function submitOwnershipProof(uint96 validityTimestamp, bytes32 proof, bytes calldata signature) external {
```
RayXpub: nit: you can use `uint256` for the timestamp here and all other instances since it is not actually stored in state and does not offer any variable packing.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: That happens in the Workflow Registry, not here. Workflow Registry has to wrap these functions. I was waiting on some pending PRs to stitch both of them together.
Also, we've discussed that we want to remove all workflows once unlinking is requested. So we need to determine which path is more desirable from the UX perspective. I see some potential UX difficulties if we revert in case active workflows are found. But on the other hand, it is safer for the user, because the user must understand that unlinking is a destructive action. I'll start a thread on this topic - I think this is also a product question.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Are you suggesting no pagination at all? What about EVM stack limits, tx gas limits? Would this still work flawlessly if we had 10k, 100k, up to one million entries in this mapping?

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Thanks! Do you know if there is a formatter that I can use that enforces everything from our style guide?

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Is it okay to have a list of signers publicly available? 🤔 Just asking if we should disclose this information or not. I don't see any potential attack vectors, though.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Only one Workflow Registry contract version can be active at a time, but it's still a good callout. I don't think it will hurt (or complicate things) if we added this in the message digest. 👍

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: You mean, having just `s_ownersProofs` is sufficient?

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: It does. So you want me to add this comment (note) to the code?

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Let's discuss naming because right now it is quite confusing 😅 
We have "workflow ownership proof" vs "linking" everywhere from design docs, to Go code, to Solidity

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
RayXpub: You can run solhint which is already configured to use the `chainlink-solidity` plugin: https://github.com/smartcontractkit/chainlink-evm/blob/f7e9b906d4d0e1f974cbc3b5ee3916b3a7779240/contracts/.solhint.json#L3

However I don't believe there is currently a rule for the above so this is still a manual check.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
RayXpub: We can still chose wether to expose it or not with a set, we don't have to implement a getter.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
+// SPDX-License-Identifier: BUSL 1.1
+pragma solidity 0.8.24;
```
eutopian: pragma solidity 0.8.26;

we should use instead

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
```diff
@@ -0,0 +1,221 @@
...
+  /// @dev The function tries to re-generate the message digest based on the provided parameters and by following
+  /// EIP-191. The it will try to recover the signer address. The function will revert if the signature is invalid.
+  function _recoverSigner(
+    uint96 validityTimestamp,
+    bytes32 proof,
+    bytes calldata signature
+  ) internal view returns (address) {
+    // Follow EIP-191 for recoverable signatures
+    bytes32 messageHash = keccak256(abi.encodePacked(msg.sender, block.chainid, validityTimestamp, proof));
+    bytes32 prefixedMessageHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", messageHash));
```
RayXpub: nit: you can use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dbb6104ce834628e473d2173bbc9d47f81a9eec3/contracts/utils/cryptography/MessageHashUtils.sol#L30

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: It wasn't there when the PR was originally created, but I will update it now according to the new Foundry profile for v2

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Renamed to "links"

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: I had to leave both in the end because the unlink operation must strictly check that the incoming proof in the signature is in its original form (to prevent a certain vector of attack). I explained this in the code comments.

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
```diff
@@ -0,0 +1,276 @@
...
+  /// EIP-191. The it will try to recover the signer address. The function will revert if the signature is invalid.
+  function _recoverSigner(
+    address owner,
+    uint256 validityTimestamp,
+    bytes32 proof,
+    bytes memory signature
+  ) internal view returns (address) {
+    // Follow EIP-191 for recoverable signatures
+    bytes32 prefixedMessageHash = MessageHashUtils.toEthSignedMessageHash(
+      keccak256(abi.encodePacked(owner, block.chainid, address(this), validityTimestamp, proof))
```
RayXpub: suggestion: I'd recommend using `abi.encode` instead of `abi.encodePacked` for pre image hash generation. The current implementation is fine but if at some point we introduce dynamic types we increase the risk of hash collisions.

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
```diff
@@ -0,0 +1,276 @@
...
+      revert UnlinkOwnerRequestExpired(owner, block.timestamp, validityTimestamp);
+    }
+
+    // This prevents unlinking of an owner address that is not even linked to this contract
+    if (!s_linkedOwners.contains(owner)) {
+      revert OwnershipLinkDoesNotExist(owner);
+    }
+
+    // It's very important to check that proof matches the one that was originally submitted from the owner
+    // (where msg.sender == owner). For unlinking, we are not enforcing that the caller is the owner, but we
```
RayXpub, line 166: q: why are we not enforcing that the caller is the owner?

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
```diff
@@ -0,0 +1,276 @@
...
+    // are expecting that whoever is calling this function must have access to the original proof signed by the
+    // trusted entity. The original proof will be provided only to the ones who has access to the account that
```
RayXpub: Isn't the owner to proof info emitted in this event and therefore publicly available?https://github.com/smartcontractkit/chainlink-evm/blob/a10462645ae74cba7193c2724c74a1f7aa3ae529/contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol#L134

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
```diff
@@ -0,0 +1,276 @@
...
+  /// in a revert, then the ownership proof is valid and the owner address can be unlinked.
+  function unlinkOwner(
+    address owner,
+    uint256 validityTimestamp,
+    bytes32 proof,
+    bytes calldata signature
+  ) public virtual {
+    tryUnlinkOwner(owner, validityTimestamp, proof, signature);
+
+    s_ownerProofs[owner] = bytes32(0);
```
RayXpub:
```suggestion
    delete s_ownerProofs[owner];
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -115,4 +124,83 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param proof The ownership proof to be submitted.
+  /// @param signature The signature of the ownership proof metadata.
+  function tryUnlinkOwner(
+    address owner,
+    uint256 validityTimestamp,
+    bytes32 proof,
+    bytes calldata signature
+  ) public view override {
+    // set removeWorkflows to false, to simulate the original tryUnlinkOwner
+    this.tryUnlinkOwner(owner, validityTimestamp, proof, signature, PreUnlinkAction.NONE);
```
RayXpub: nit: we usually try to avoid function name overloading due to geth wrappers generation, this generates `tryUnlinkOwner` and `tryUnlinkOwner0`: https://github.com/smartcontractkit/chainlink-evm/blob/5107196420472998b3a98a18230e1d476d0c6dd4/gethwrappers/workflow/generated/workflow_registry_wrapper_v2/workflow_registry_wrapper_v2.go#L398

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
```diff
@@ -0,0 +1,184 @@
...
+
+    vm.startPrank(owner);
+    op = new OwnershipLinkTestable();
+    address[] memory signers = new address[](1);
+    signers[0] = allowedSigner;
+    op.updateAllowedSigners(signers, true);
+    vm.stopPrank();
+  }
+
+  function test_WhenThereAreNoLinkedOwners() external view {
```
RayXpub: nit: test names should be: `test_FunctionName_Description` see [foundry guide](https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/FOUNDRY_GUIDE.md#test-naming)

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
```diff
@@ -0,0 +1,184 @@
+// SPDX-License-Identifier: BUSL 1.1
+pragma solidity 0.8.26;
+
+import {LinkingUtils} from "../../testhelpers/LinkingUtils.sol";
+import {OwnershipLinkTestable} from "../../testhelpers/OwnershipLinkTestable.sol";
+
+import {Test} from "forge-std/Test.sol";
+
+contract OwnershipLinkGetLinkedOwners is Test {
```
RayXpub: nit: test contract name is usually `ContractName_functionName`

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
ibrajer: Do you know how to force `bulloak` to generate names like this? I can update this by hand, but `bulloak check` would fail afterward (I think?)

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
ibrajer: Sure, this probably can be done by changing the top name in the `.tree` file

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Gotcha, in that case, I can either:
* introduce new functions such as `linkWorkflowOwner` and `unlinkWorkflowOwner`, and have them call inherited functions (in terms of gas, not sure much extra cost would that add)
* push the entire `OwnershipLink` code into `WorkflowRegistry` (so not using abstract contract anymore, but this could uglify the code a bit, since everything will be in a single file)

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
ibrajer: Actually, I think there might be a time window when this could be abused, depending on the validity timestamp. So we might have to add an additional flag to prevent this. I need to try this out.

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
RayXpub: > Do you know how to force bulloak to generate names like this? 

afaik there isn't.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: @RayXpub what do you think of adding an internal postlink function in the abstract, something like
```
function _afterUnlinkOwner(address owner, bytes32 proof) internal virtual {
  // noop by default
}
```

where in the abstract we can have 

```
function unlinkOwner(
  address owner,
  uint256 validityTimestamp,
  bytes32 proof,
  bytes calldata signature
) public virtual {
  tryUnlinkOwner(owner, validityTimestamp, proof, signature);
  s_ownerProofs[owner] = bytes32(0);
  s_linkedOwners.remove(owner);
  emit OwnershipLinkUpdatedV1(owner, proof, false);
  // Call post-unlink hook
  _afterUnlinkOwner(owner, proof);
}
```


and then override this `_afterUnlinkOwner` function in the derived function in whatever logic that makes sense (might not even be deleting workflows if it's not a workflow registry contract)


not sure if this is done in solidity best practices but it would allow future derived contracts to do cleanup work without overriding public functions

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -5,14 +5,77 @@ import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
...
+  /// address and to the off-chain account, may be able to generate a valid signature signed by the trusted entity. Once this
+  /// valid signature is submitted to this contract, it can be verified and used to link or unlink the owner address.
+  mapping(address owner => bytes32 proof) private s_ownerProofs;
+
+  // ================================================================
+  // |                         Events                               |
+  // ================================================================
+
+  event AllowedSignersUpdatedV1(address[] signers, bool allowed);
+  event OwnershipLinkUpdatedV1(address indexed owner, bytes32 indexed proof, bool indexed added);
```
DeividasK, line 38: Why do we have these "V1" event suffixes?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -5,14 +5,77 @@ import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
...
 
   enum WorkflowStatus {
     ACTIVE,
     PAUSED
   }
 
+  enum PreUnlinkAction {
+    NONE, //              No action prior to unlinking.
+    REMOVE_WORKFLOWS, //  Remove all workflows owned by the owner prior to unlinking.
+    PAUSE_WORKFLOWS //    Pause all workflows owned by the owner prior to unlinking.
```
DeividasK, line 65: Hmmm, not sure I'm a fan of pausing. How could workflows progress from this state?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -115,4 +178,254 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param validityTimestamp Validity of the ownership proof.
+  /// @param proof The ownership proof to be submitted.
+  /// @param signature The signature of the ownership proof metadata.
+  /// @dev This function is used to verify if the ownership proof is valid without actually linking the owner address.
+  /// The ownership proof metadata is a combination of the claimed owner address, validity timestamp, and the proof hash.
+  /// Request will be rejected if the validity timestamp has expired, owner addres is already linked, if the proof does
+  /// not match the one that was originally submitted, or if the signature is not valid (for different reasons).
+  /// @dev Verification will use msg.sender as the owner address, which means caller must be the owner of the address,
+  /// otherwise the verification will fail with invalid signature error.
+  function tryLinkOwner(uint256 validityTimestamp, bytes32 proof, bytes calldata signature) public view virtual {
```
DeividasK: Nit:
```suggestion
  function canLinkOwner(uint256 validityTimestamp, bytes32 proof, bytes calldata signature) public view virtual {
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Historically, I wanted to have a way to differentiate between events originating from different contract versions that contain differences in event payloads. If an event payload change occurs, the version tag increases. It's just a semantic labeling. E.g. say you have `OwnershipLinkUpdatedV1` in Workflow Registry V2 and then in the Workflow Registry V3, `proof` changes to `bytes`. You would indicate that change by creating a new event called `OwnershipLinkUpdatedV2` rather than being mixed with the previous event name. If the event remained the same, then it would still be called `OwnershipLinkUpdatedV1` in Workflow Registry V3.

My thoughts were - it would be easier for off-chain handling if you had to create different data unpackers for different versions of this event. I was investigating event versioning and one option was embedding the version label in the event payload (but that increases the gas cost when emitting the event).

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: They can't, they're just paused forever in case the customer wants to keep them in registry history (unless the workflow owner is linked again, then they can be reactivated). Do you think there's a legit use case for that?

Here's the scenario that can happen and I wasn't sure how the customer would react:
* customer initiates unlinking
* customer doesn't want any additional pre-unlink action
* registry detects active workflows and reverts the transaction, the customer is instructed to "clean up"
* customer decides to remove some workflows, but pause some of the other ones, overall there are no more active workflows
* customer initiates unlinking again, doesn't want any additional pre-unlink action
* registry detects no active workflows, but there are a few paused ones - should it proceed or should it revert again and ask the customer to remove all workflows?

DeividasK: IMHO, we should push for removal. The only legit use case where pausing might be desirable is if a user wants to transfer the owner address to another org, which, I think, should be supported off-chain. WDYT?

Btw, this isn't a blocker.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: IMHO, event parsing is always attached to the contract address, so off-chain, you first get an address and then fetch events, which are by default attached to that address. If you want to do some smart switching, you could always grab the `typeAndVersion`, too. Personally, I don't think a version suffix to events is that useful 🤷‍♂️ not a blocker.

ibrajer: Got it - I think that for now, I'll align with other upcoming changes on the Workflow Registry. It would be even more confusing if event names on the same contract used a different naming strategy. In other words, if the version tag should be dropped, it should be dropped simultaneously for all existing contract events.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: > The only legit use case where pausing might be desirable is if a user wants to transfer the owner address to another org, which, I think, should be supported off-chain.

Right now, we request explicit unlinking first and then linking the owner to a different org, because the proof hash is generated based on the org details.

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
ibrajer: Done

contracts/src/v0.8/workflow/dev/v2/test/OwnershipLink/OwnershipLink.getLinkedOwners.t.sol
ibrajer: Done

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I've decided to merge `OwnershipLink` with `WorkflowRegistry` to keep things simple and use linking functions directly rather than overriding them. I don't expect anyone to reuse `OwnershipLink` any time soon.

contracts/src/v0.8/workflow/dev/v2/OwnershipLink.sol
ibrajer: It is, but as discussed in person, proof is not sensitive information, and due to added precautions on the Workflow Registry side (verifying if the signature contains the original proof, embedding request type in the message digest), malicious attackers can't use it.

infiloop2 (approved):
stamping to unblock further dependencies implementation. Contract is not final and there might be followup PRs

DeividasK (approved):

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
eutopian: If that's the case, then why add the set one and not just use the mapping?

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
eutopian: This is the only place where the s_ownersProofsSet is used right? Would it make sense to just use the s_ownersProofs if that one is needed?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -115,4 +178,254 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  function _recoverSigner(
+    uint8 requestType,
+    address owner,
+    uint256 validityTimestamp,
+    bytes32 proof,
+    bytes calldata signature
+  ) internal view returns (address) {
+    // Follow EIP-191 for recoverable signatures
+    bytes32 prefixedMessageHash = MessageHashUtils.toEthSignedMessageHash(
+      keccak256(abi.encode(requestType, owner, block.chainid, address(this), validityTimestamp, proof))
```
RayXpub, line 421: nit: since we merged the 2 contracts we could also add the `typeAndVersion` to the pre image hash.

RayXpub (approved):

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: This is old code you're referring to, but this set is mainly intended for use in the `canUnlinkOwner` to extract the stored proof. It's a bit unfortunate having to use both, but we can't protect users against attacks during unlinking without cross-comparing proofs.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Honestly, I've changed this code so many times I don't even know anymore. Let me revisit it one more time.

contracts/src/v0.8/workflow/dev/OwnershipProof.sol
ibrajer: Will revisit: https://github.com/smartcontractkit/chainlink-evm/pull/73#discussion_r2138199971

ibrajer: Okay, so `s_ownerProofs` (mapping) is used when we need a quick way to retrieve proof based on the owner's address. We need this in the main linking and unlinking functions. On the other hand, `s_linkedOwners` (address set) is used in view functions to iterate through a list of linked owner addresses (in that case, proofs are not relevant). Since we can't iterate through a mapping, the only way is to store these addresses in a separate data structure.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I will merge this PR now to unblock @eutopian and then address this comment in the follow-up PR!

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -115,4 +178,254 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    // was used for linking the owner address.
+    bytes32 storedProof = s_ownerProofs[owner];
+    if (storedProof != proof) {
+      revert OwnershipLinkProofDoesNotMatch(owner, proof, storedProof);
+    }
+
+    // Request type prevents replay attacks, since the same proof can be used for both linking and unlinking
+    address signer = _recoverSigner(uint8(LinkingRequestType.UNLINK_OWNER), owner, validityTimestamp, proof, signature);
+    if (!s_allowedSigners[signer]) {
+      revert InvalidOwnershipLink(owner, validityTimestamp, proof, signature);
```
timweri, line 322: Should we have a separate event here? `InvalidOwnershipUnlink`?

timweri: Maybe also `OwnershipUnlinkProofDoesNotMatch`

PR #74, opened 2025-05-26
[CR-403] Move core/chains from the main repo to the chainlink-evm repo.

pavel-raykov:
tron used to be the only subfolder of the core/chains/evm folder, so in the new repo I moved it one level higher - to pkg/chains/tron instead of pkg/chains/evm/tron.

Supports https://github.com/smartcontractkit/chainlink/pull/17888


pkg/chains/tron/txm.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-evm/pkg/config"
+	"github.com/smartcontractkit/chainlink-evm/pkg/config/toml"
+	"github.com/smartcontractkit/chainlink-evm/pkg/keys"
+	tronkeystore "github.com/smartcontractkit/chainlink-tron/relayer/keystore"
+	tronclient "github.com/smartcontractkit/chainlink-tron/relayer/sdk"
+	trontxm "github.com/smartcontractkit/chainlink-tron/relayer/txm"
+)
+
+func ConstructTronTxm(logger logger.Logger, cfg *config.ChainScoped, nodes []*toml.Node, keystore keys.Store) (*trontxm.TronTxm, error) {
```
jmank88: Can probably drop `Tron` from the name given we are in `package tron`:
```suggestion
func ConstructTxm(logger logger.Logger, cfg *config.ChainScoped, nodes []*toml.Node, keystore keys.Store) (*trontxm.TronTxm, error) {
```
Who was the original author? Can we include a comment here about what is going on exactly?

pavel-raykov: @KodeyThomas could you please suggest a comment explaining this code?

KodeyThomas:
```suggestion
// Creates the Transaction Manager used for Tron-based chains. Uses an adapter of the EVM Keystore to sign.
// Doesn't currently support multinode writes
```

pavel-raykov: Thanks, I have updated the method name and the comment.

go.mod
```diff
@@ -192,3 +206,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/fbsobreira/gotron-sdk => github.com/smartcontractkit/chainlink-tron/relayer/gotron-sdk v0.0.5-0.20250422175525-b7575d96bd4d
```
jmank88, line 210: Where does this version come from? It looks like `relayer/v0.0.10` is `d7d29f6f0fc8a25943b01245a2a5ebc49d9c16aa`: https://github.com/smartcontractkit/chainlink-tron/releases/tag/relayer%2Fv0.0.10

pavel-raykov: I am not sure. I took it from the main repo 

https://github.com/smartcontractkit/chainlink/blob/b5e5f8bccfdc764ccba4ce8f87ce28223426c667/go.mod#L389

It was also added by @KodeyThomas

pavel-raykov: @KodeyThomas , could you please say why the tron version is pinned to specific 

https://github.com/smartcontractkit/chainlink-tron/commit/b7575d96bd4d746444cfc7320b1509f82fb90c45 

?

KodeyThomas: No particular reason, originally why we didn't was due to the structure we'd have to create multiple tags for each release version:
- `relayer/gotron-sdk/v.xyz`
- `relayer/v.xyz`

Hoping to remove this dependency soon anyway when we start on the CCIP 1.6 work for Tron

jmank88: I think we just dropped the local replace, but the version that we are pointing to here still has it. The way to reproduce that local replace is to reference the same hash. Core was probably stale, because this has to be maintained with every bump of chainlink-tron. However, it is a very stale module, and so this likely doesn't actually make a difference.

jmank88 (approved):

KodeyThomas (approved):

PR #76, opened 2025-05-27
LogPoller: Add reporting of missing blocks errors for hyperliquid chain

reductionista:
JIRA: [PLEX-1453](https://smartcontract-it.atlassian.net/browse/PLEX-1453)

#Description 

The specific error on hyperliquid we need to handle is "invalid block range", but the general error condition it represents is a state where the rpc server is permanently missing one or more blocks requested in an eth_getLogs range request.

This is required for hyperliquid, but could also be useful for any rpc server which prunes logs or blocks. If they've been permanently pruned, we want to continue. Adding for hyperlqiuid now, but making it configurable will make it easy to add support for any other chains we see this issue come up on.  Both immediately for the short term (via a custom toml config regex) and for the longer term by adding another regex to this `MissingBlocks` error & appropriate error code.

[PLEX-1453]: https://smartcontract-it.atlassian.net/browse/PLEX-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/logpoller/log_poller.go
+++ b/pkg/logpoller/log_poller.go
@@ -137,6 +137,7 @@ type logPoller struct {
 	// LogPoller keeps running in infinite loop, so whenever the invalid state is removed from the database it should
 	// recover automatically without needing to restart the LogPoller.
 	finalityViolated           atomic.Bool
+	missingBlocksErrorCount    atomic.Uint64
 	countBasedLogPruningActive atomic.Bool
 }
 
@@ -536,6 +537,9 @@ func (lp *logPoller) Healthy() error {
 	if lp.finalityViolated.Load() {
 		return commontypes.ErrFinalityViolated
 	}
+	if errCount := lp.missingBlocksErrorCount.Load(); errCount > 2 {
+		return pkgerrors.Errorf("rpc servers reported missing blocks %d times in a row", errCount)
+	}
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


jmank88 (dismissed):

DylanTinianov (approved):

pkg/logpoller/log_poller.go
```diff
@@ -901,6 +905,16 @@ func (lp *logPoller) backfill(ctx context.Context, start, end int64) error {
 
 		gethLogs, err := lp.latencyMonitor.FilterLogs(ctx, lp.Filter(big.NewInt(from), big.NewInt(to), nil))
```
prashantkumar1982, line 906: There's another place this same call is made, in pollAndSaveLogs().
So we should add this same check there too?

reductionista: This "invalid range" error can only come back in response to a range request. The other FilterLogs call requests logs by block hash, so I don't think it would be possible to get this specific error back.

It would require some more testing and/or looking through node op logs to determine whether it's possible to get an error back in that case, and if so what the error would look like and whether we can always assume it's due to a permanent issue. But it shouldn't matter much since that one is only done for unfinalized logs. If there is an error there, it will get retried until the block becomes finalized at which point it will use a block range request to get it, and be handled properly when it gets back the "invalid block range" error we do look for.

prashantkumar1982 (approved):

PR #78, opened 2025-05-27
pkg/keys: s/SignRawUnhashedBytes/Sign

jmank88:
Supports:
- https://github.com/smartcontractkit/chainlink/pull/17904

### DIFF
--- a/pkg/keys/keystore.go
+++ b/pkg/keys/keystore.go
@@ -24,7 +24,7 @@ type Store interface {
 	RoundRobin
 	MessageSigner
 	Locker
-	RawUnhashedSigner
+	Signer
 }
 
 // ChainStore extends Store with methods that require a chain ID.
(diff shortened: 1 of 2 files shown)


pkg/keys/keystore.go
```diff
@@ -46,11 +46,8 @@ type AddressLister interface {
 	EnabledAddresses(ctx context.Context) (addresses []common.Address, err error)
 }
 
-// RawUnhashedSigner is an interface for signing raw bytes without hashing them first.
-// This is useful for signing transactions or other data that needs to be signed as-is (required for TRON).
-// The interface effectively exposes the Sign method from the keystore
-type RawUnhashedSigner interface {
```
jmank88, line 52: I don't think these extra qualifiers and explanation are adding value here. This is simply a passthrough `Sign` call :shrug: no need to complicate it.

pavel-raykov (dismissed):

pkg/keys/keystore.go
KodeyThomas: Not sure i agree..

Given this keystore is primarily used for EVM chains calling `Sign` will **not** produce a valid signature. This needs to be documented if we want to change the naming

KodeyThomas (changes requested):

pkg/keys/keystore.go
krebernisak: I'd suggest to consider `RawSigner.SignRaw`, but `Signer` is also ok if we document this is just a passthrough to `ks.Sign` vs `MessageSigner` which is expected to hash the input

krebernisak (dismissed):

pkg/keys/keystore.go
jmank88: > Given this keystore is primarily used for EVM chains calling Sign will not produce a valid signature.

What is your basis for this claim? All that this does is expose the underlying `Sign` method. This is not related to the other adapter at all.

KodeyThomas (approved):

pavel-raykov (approved):

pkg/keys/keystore.go
Darkknight-web: 👍🏻

PR #81, opened 2025-05-30
Add LogPoller Test ORM

amit-momin:
Created a test ORM for methods only needed by tests. The Solana CCIP E2E test needs to read the LogPoller DB to determine when filters for certain events are registered before proceeding

Core PR: https://github.com/smartcontractkit/chainlink/pull/17939

### DIFF
--- a/pkg/logpoller/testing/test_orm.go
+++ b/pkg/logpoller/testing/test_orm.go
@@ -4,17 +4,18 @@ import (
 	"context"
 
 	"github.com/ethereum/go-ethereum/common"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/sqlutil"
 )
 
 type TestDSORM struct {
-	ds      sqlutil.DataSource
+	ds sqlutil.DataSource
 }
 
 // NewTestORM creates a test DSORM which contains method only used by tests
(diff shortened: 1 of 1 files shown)


pkg/logpoller/testing/test_orm.go
```diff
@@ -0,0 +1,32 @@
...
+func NewTestORM(ds sqlutil.DataSource) *TestDSORM {
+	return &TestDSORM{
+		ds:      ds,
+	}
+}
+
+// HasFilterByEventSig checks if a filter exists for the provided event sig
+func (o *TestDSORM) HasFilterByEventSig(ctx context.Context, chainID string, eventSig common.Hash) (bool, error) {
+	query := `SELECT COUNT(1) FROM evm.log_poller_filters
+		WHERE evm_chain_id = $1 AND event = $2 LIMIT 1`
```
reductionista: I think we should add `address` as a param here too, searching for (address, event_sig) pair. Just searching for any events on chain with that sig could confuse events emitted from different smart contracts. Maybe in this particular test, there is only one contract deployed which can emit that type of event. But I could see writing slight variations of it where there are multiple contracts of the same type deployed.

reductionista (approved):

huangzhen1997 (approved):

PR #83, opened 2025-05-30
OEV-221: Fixes inconsistencies between beholder metrics & onchain txs for the purged Txs

cl-efornaciari:
- Purged txs ToAddress gets updated to 0x0 when writing to chain but beholder metrics for the same tx maintain the original ToAddress
- This change updated purged tx to uses the 0x0 address for beholder metrics to maintain consistency between onchain tx and beholder metrics


pkg/txm/txm_test.go
dimriou, line 1: Some thoughts about the tests:
The metrics are completely abstracted from the TXM. As far as the TXM goes, it only makes a call to the `EmitTx` method, but if the TXM "lives" outside the Core node, this method is considered a NOOP. The TXM already tests the standard path as well as the nonce gap. Here, we're copy-pasting the same code, which adds some noise.
I would propose two alternatives for these tests.
- Instead of repeating the same code, add the Beholder check on the existing test cases as an additional step.
- Create a new `metrics_test.go` file and specifically test the EmitTx method via a unit test.

cl-efornaciari: Makes sense to me, I had a similar thought but opted to test the critical path we were investigating in this ticket. I'll refactor to test the `EmitTx` since that's whats actually in question

dimriou (dismissed):

pkg/txm/metrics_test.go
```diff
@@ -0,0 +1,153 @@
...
+				var actual svrv1.TxMessage
+				if err := proto.Unmarshal(b, &actual); err != nil {
+					return false
+				}
+
+				return actual.FromAddress == fromAddress.String() &&
+					actual.ToAddress == expectedToAddress.String() &&
+					actual.Nonce == strconv.FormatUint(expectedNonce, 10) &&
+					actual.ChainId == expectedChain.String() &&
+					actual.FeedAddress == ""
```
cll-gg: Nitpick:
If e.g. the nonce has an unexpected value, does this test fail by printing the unexpected value, or does the test fail because Emit was called with a different byte array than expected?
If the latter is the case, then it'd be better if we could capture the byte array, unmarshal it and assert on its fields separately.

cll-gg (dismissed):

cll-gg (approved):

dimriou (approved):

PR #85, opened 2025-06-01
Add SenderAddress for Dynamic Gas Limit Estimation

ma33r:
### What
- This PR adds a configurable Estimation Sender Address for dynamic gas estimation
- This value is set in the OCR config and called `SenderAddress`. It is empty and ignored by default, but if it is set to a nonempty string, this address will be used for the "from" field of the `callMsg` for dynamic gas estimation 

### Why
- This field is needed for dynamic gas estimation to work correctly for CCIP. More information and context can be found [here](https://smartcontract-it.atlassian.net/wiki/spaces/CCIP/pages/1159069831/CCIP+Dynamic+Gas+Estimation)


pkg/config/toml/defaults/fallback.toml
```diff
@@ -53,6 +53,7 @@ FeeCapDefault = '100 gwei'
 TipCapDefault = '1'
 TipCapMin = '1'
 EstimateLimit = false
+EstimationSenderAddress = ''
```
jmank88: Does repeating Estimation add value here?
```suggestion
SenderAddress = ''
```

jmank88 (changes requested):

pkg/config/toml/defaults/fallback.toml
ma33r: Thank you; you're correct - I'll update this to SenderAddress instead

ma33r: Actually, on second thought, just naming this "SenderAddress" could lead to a misunderstanding of the config, as this is not the actual sender address of the transaction - it is only the sender address used for gas estimation. I'll change it back to "EstimationSenderAddress"

pkg/gas/models.go
```diff
@@ -362,13 +362,23 @@ func (e *evmFeeEstimator) estimateFeeLimit(ctx context.Context, feeLimit uint64,
 	if !e.geCfg.EstimateLimit() {
 		return providedGasLimit, nil
 	}
+
+	// If EstimationSenderAddress string is not empty, then we unwrap the config value and set the pointer. Otherwise, it remains nil
+	var estimationSenderAddress *common.Address
+	if len(*e.geCfg.EstimationSenderAddress()) != 0 {
```
amit-momin: Think we should check for nil here before dereferencing for safety
```suggestion
	if e.geCfg.EstimationSenderAddress() != nil && len(*e.geCfg.EstimationSenderAddress()) != 0 {
```

pkg/gas/helpers_test.go
```diff
@@ -122,20 +122,21 @@ func (m *MockBlockHistoryConfig) TransactionPercentile() uint16 {
...
+	TipCapDefaultF           *assets.Wei
+	TipCapMinF               *assets.Wei
+	PriceMaxF                *assets.Wei
+	PriceMinF                *assets.Wei
+	PriceDefaultF            *assets.Wei
+	FeeCapDefaultF           *assets.Wei
+	LimitMaxF                uint64
+	ModeF                    string
+	EstimateLimitF           bool
+	EstimationSenderAddressF *string
```
amit-momin: For the mock config it might be easier to just return string directly here if it's always set to '' so you don't have to worry about setting it to a pointer in the tests every time.

pkg/gas/models.go
ma33r: Thank you Amit! I have just added this check

pkg/gas/helpers_test.go
ma33r: I could not set this to directly return a string here as MockGasEstimatorConfig implements GasEstimatorConfig, and there EstimationSenderAddress() returns `*string`

pkg/gas/models.go
```diff
@@ -362,13 +362,23 @@ func (e *evmFeeEstimator) estimateFeeLimit(ctx context.Context, feeLimit uint64,
...
+
 	// Create call msg for gas limit estimation
 	// Skip setting Gas to avoid capping the results of the estimation
 	callMsg := ethereum.CallMsg{
 		To:   toAddress,
 		Data: calldata,
 	}
-	if fromAddress != nil {
+	if estimationSenderAddress != nil {
+		callMsg.From = *estimationSenderAddress
```
amit-momin: Sorry to add another nit but looking at this a little longer I realized we're doing the same check twice. Could we just replace this with something like below and drop `estimationSenderAddress` 
```suggestion
	if e.geCfg.EstimationSenderAddress() != nil && len(*e.geCfg.EstimationSenderAddress()) != 0 {
		callMsg.From = common.HexToAddress(*e.geCfg.EstimationSenderAddress())
```

pkg/config/toml/defaults/fallback.toml
jmank88: But the fully qualified name is the thing to consider - we are already inside of the `GasEstimator` table here: `EVM.GasEstimator.SenderAddress` vs. `EVM.GasEstimator.EstimationSenderAddress`

pkg/gas/models.go
ma33r: Thank you Amit; this makes sense, and I've just updated the PR to reflect this

pkg/config/toml/defaults/fallback.toml
ma33r: Thank you Jordan - that makes sense; I've updated the parameter name

pkg/config/toml/config.go
```diff
@@ -761,6 +762,9 @@ func (e *GasEstimator) setFrom(f *GasEstimator) {
 	if v := f.EstimateLimit; v != nil {
```
amit-momin, line 762: Sorry missed this earlier too since the github diff cut off this part of the file but think we should add a validation in the GasEstimator `ValidateConfig` method above. If the `SenderAddress` is not nil and has len > 0, we should use `common.IsHexAddress()` to verify it's a properly formatted address so we can catch any misconfigs early.

ma33r: Thank you Amit, I've just added this validation

amit-momin (dismissed):

pkg/config/toml/docs.toml
```diff
@@ -205,6 +205,8 @@ LimitMultiplier = '1.0' # Default
 LimitTransfer = 21_000 # Default
 # EstimateLimit enables estimating gas limits for transactions. This feature respects the gas limit provided during transaction creation as an upper bound.
 EstimateLimit = false # Default
+# SenderAddress is optional and can be set to a specific sender address for dynamic gas estimation. If it is not set, the actual "from" address for the transaction is used if available.
+SenderAddress = '' # Default
```
jmank88: What does a default value of empty string mean? I think we want an example here:
```suggestion
SenderAddress = '0xabcd1234' # Example
```

pkg/config/config.go
```diff
@@ -151,6 +151,7 @@ type GasEstimator interface {
 	Mode() string
 	PriceMaxKey(gethcommon.Address) *assets.Wei
 	EstimateLimit() bool
+	SenderAddress() *string
```
jmank88: Let's return an address instead of a string. Then we get validation once up-front during initialization and the callers don't have to worry about it.
```suggestion
	SenderAddress() *gethcommon.Address
```

pkg/config/toml/config.go
```diff
@@ -669,6 +670,7 @@ type GasEstimator struct {
 	LimitTransfer   *uint64
 	LimitJobType    GasLimitJobType `toml:",omitempty"`
 	EstimateLimit   *bool
+	SenderAddress   *string
```
jmank88: You can use a `types.EIP55Address` here too:
```suggestion
	SenderAddress   *types.EIP55Address
```

pkg/config/toml/docs.toml
ma33r: This is an empty string by default because it does not have to be set. If it is not set (i.e. set to the empty string), then it is ignored, and the "from" address for the transaction is used for dynamic gas estimation instead. Also, setting this has no effect when `EstimateLimit` is set to false (which it is by default), so this should not be set by default

pkg/config/config.go
ma33r: That makes sense, but if we were to return an address instead of a string, then there would be no way to leave the parameter un-set if we don't want to set a custom estimation sender address (like we are doing now with an empty string value)

pkg/config/toml/docs.toml
jmank88: Please add an example value.

pkg/config/config.go
jmank88: We use pointers for every field so that we can tell if it was not set.

pkg/config/toml/docs.toml
ma33r: Understood; I have added an example value.

pkg/config/config.go
ma33r: Thank you for your guidance here Jordan, I've updated the type of the parameter in the PR

jmank88 (approved):

amit-momin (approved):

PR #87, opened 2025-06-03
Workflow Registry v2

eutopian:
### 📦 New & Refactored Features

1. **Configurable DON & User Limits**  
   - **`setDONLimit`** / **`getMaxWorkflowsPerDON`**  
     Owner can enable/disable a global cap on active workflows per DON.  
   - **`setUserDONOverride`** / **`getMaxWorkflowsPerUserDON`**  
     Owner may override that cap for a specific user—but never above the global DON limit.

2. **Workflow Lifecycle**  
   - **`upsertWorkflow`**: single entry point for create/update based on `(owner, name, tag)` key.  
   - **Status Management**:  
     - Single: `pauseWorkflow` / `activateWorkflow`  
     - Batch: `pauseWorkflows` / `activateWorkflows`
     - Admin: `adminPauseWorkflow`, `adminBatchPauseWorkflows`, `adminPauseAllByUser`, `adminPauseAllByDON`  
   - **Deletion**: `deleteWorkflow` + internal `_applyDelete` cleans up all indices.

3. **Events log**  
   - Internal `EventRecord[] s_events` log for history when events are emitted (`WorkflowRegistered`, `WorkflowUpdated`, `WorkflowPaused`, `WorkflowActivated`, DONLimitSet`).


contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -338,22 +420,16 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
     bytes calldata signature,
-    PreUnlinkAction /* preUnlinkAction */
+    PreUnlinkAction action
   ) external {
-    // TODO: if preUnlinkAction is not NONE, assume that unlinkWorkflowOwner() will remove or pause workflows
-    // TODO: if preUnlinkAction is NONE, verify if there are any active workflows, if yes, revert
-
-    canUnlinkOwner(owner, validityTimestamp, signature, PreUnlinkAction.NONE);
+    _validateOwnershipLinks(owner, validityTimestamp, signature);
+    _handlePreUnlinkAction(owner, action);
```
ibrajer: @eutopian Can you switch places here? If there are any active workflows, this transaction should revert before any ownership link validations.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +499,569 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    // Iterate from the back since EnumerableSet.remove() swaps-and-pops.
+    while (active.length() > 0) {
+      bytes32 rid = active.at(active.length() - 1);
+      if (action == PreUnlinkAction.PAUSE_WORKFLOWS) {
+        _pauseWorkflow(rid);
+      } else {
+        _deleteWorkflow(rid);
+      }
+    }
+  }
```
ibrajer, line 668: @eutopian Can I add unit tests for this case by contributing to this PR? I would only touch the test files.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -34,12 +40,52 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   /// enforce an invariant on proofs.
   mapping(bytes32 proof => bool used) private s_usedProofs;
 
+  /// @dev workflowRID is a hash of (owner ∥ name ∥ label). It functions as the primary index for the workflow storage.
+  mapping(bytes32 workflowRID => WorkflowMetadata workflowMetadata) private s_workflows;
```
ibrajer: Sorry, what does RID stand for?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -76,26 +135,17 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
-  constructor() {
-    // Intialize with default limits for Config.
-    s_cfg.defaultMaxPerDON = 500;
-    s_cfg.defaultMaxPerUserDON = 200;
+    bytes32 donLabel; //         Label for the DON (32 chars limit) in bytes32.
+    string workflowName; //      Human readable string (64 chars limit).
+    string binaryURL; //         URL to the wasm binary (200 chars limit).
+    string configURL; //         URL to the config (200 chars limit).
+    string tag; //               Unique per (owner, workflowName) human readable identifier (32 chars limit)
+    bytes attributes; //         Arbitrary bytes for additional workflow details.
```
ibrajer, line 149: What is the difference between `tag` and `attributes`? Why do we need to have both? And when would someone add an attribute to the workflow (and why)? Where would these attributes be reflected?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -108,33 +158,43 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
 
-  // Our single config instance
   Config private s_cfg;
 
-  // ─────────────────────────────────────────────────────────────────────────
-  // Limits Config - External Setters: Owner can set defaults and individual overrides
-  // ─────────────────────────────────────────────────────────────────────────
+  constructor() {
+    s_cfg.defaultMaxPerDON = 500;
+    s_cfg.defaultMaxPerUserDON = 200;
```
ibrajer: Who provided these defaults? Is it based on some kind of load test and/or capacity planning projections?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +499,569 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param workflowName  Human‑readable name (≤64 chars)
+  /// @param workflowID    Deterministic hash computed off‑chain (must be unique)
+  /// @param donLabel      Label of the DON (32 chars limit)
+  /// @param status        Initial status (ACTIVE / PAUSED)
+  /// @param binaryURL     URL of the wasm binary (required)
+  /// @param configURL     URL of the config (optional)
+  /// @param attributes    Arbitrary bytes for additional workflow details (optional)
+  /// @param keepAlive     Boolean flag that determines whether existing workflows with the same
+  /// workflowName, workflowOwner combination should be paused or kept active.
+  function upsertWorkflow(
```
ibrajer, line 822: Why not `deployWorkflow`?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +499,569 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    string calldata workflowName,
+    string calldata tag,
+    bytes32 workflowID,
+    WorkflowStatus status,
+    bytes32 donLabel,
+    string calldata binaryURL,
+    string calldata configURL,
+    bytes calldata attributes,
+    bool keepAlive
```
ibrajer, line 831: But what if the customer wants to remove the old version instead? Right now, is it only possible to pause it?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +499,569 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  ) internal {
+    /* ───────────────────────── 0. VALIDATION ─────────────────────────── */
+    _requireOwnerLinked(msg.sender);
+    _validateWorkflowName(bytes(workflowName).length);
+    _validateWorkflowTag(bytes(tag).length);
+    _validateWorkflowID(workflowID);
+    _validateWorkflowURLs(bytes(binaryURL).length, bytes(configURL).length);
```
ibrajer: There is no validation for `donLabel` and `attributes`, is that on purpose? What would happen with `_enforceLimits` below if someone were to send an invalid (unregistered) DON label? And `attributes` is a blob of arbitrary data which can blow up in size, so just wondering if we were planning on restricting that or not?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: reference ID. let me add that! I thought it was in the comments but maybe it got deleted

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: tag is part of the identification and is unique to the (workflow name, owner, tag) combo. It is used to generate the internal primary key RID. attributes are more like metadata/json globs for future use (just a bytes non typed) in case we do want to add something but don't know what yet. It is more of additional information on a workflow we haven't anticipated yet now. 

tag - identifier
attributes - workflow data/info

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: no these were just random. they can be changed. open to any kind of comments on what the defaults should be I just put them as placeholders

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: I would defer the attributes question to @DeividasK 
Hmm. At the moment if someone were to send an invalid DON then the function would return the default limit. So I guess the additional question is whether or not we want to check the donLabels have true ConfigValues or not and in that way allowlist dons instead of having it be open.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: I don't think this is a feature we open up to the user yet. The current behavior will be the CLI will set it to false and only advance users with use cases like blue green will enable it via keepAlive. 

So if they want to delete it they can manually delete it.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: I guess "deploy" feels like the term we use in the CLI but at the data level it's an upsert (create or update)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: I switched them just now but was thinking over this and it seems checking ownership links is not that expensive to do first. We need to check both anyway. If we do the bottom one first I would have to separate out the function so that the 0 check and the deletes are separate, otherwise we could potentially loop through a lot of workflows and have a revert if the ownership link fails

ibrajer: Discussed on Slack - switching to the original makes more sense for now

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -34,12 +42,52 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
   mapping(bytes32 proof => bool used) private s_usedProofs;
 
+  /// @dev workflowRID (reference ID) is a hash of (owner ∥ name ∥ label). It functions as the primary index for the workflow storage.
+  mapping(bytes32 workflowRID => WorkflowMetadata workflowMetadata) private s_workflows;
+  /// @dev workflowKey is a hash of (owner ∥ name ) and maps to a set of RIDs. It is used for filtering workflows based on just
+  /// the name, as there may be multiple workflows with the same name under a different tag.
+  mapping(bytes32 workflowKey => EnumerableSet.Bytes32Set) private s_workflowKeyToRIDs;
+  /// @dev workflowID ⇒ workflowRID. This mapping lets us enforce global uniqueness of workflowID while also allows us to lookup
+  /// workflows through the workflowID.
+  mapping(bytes32 workflowID => bytes32 workflowRID) private s_idToRID;
```
ibrajer: Just to clarify, workflow ID is considered a unique identifier from the perspective of a single Workflow DON, but workflow RID only exists on the Workflow Registry for tracking purposes?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -34,12 +42,52 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
+  // Secondary indices for iteration / queries
+  mapping(address owner => EnumerableSet.Bytes32Set workflowIDs) private s_activeOwnerWorkflowRIDs; // owner -> workflowRID set
+  mapping(bytes32 donLabel => EnumerableSet.Bytes32Set workflowIDs) private s_activeDONWorkflowRIDs; // donLabel -> workflowRID set
+
+  /// @dev Every workflow ever registered under `donLabel`. Pruned only on delete.
+  mapping(bytes32 => EnumerableSet.Bytes32Set) private s_allDonRIDs;
+  /// @dev Every workflow ever registered for an owner. Pruned only on delete.
+  mapping(address => EnumerableSet.Bytes32Set) private s_allOwnerRIDs;
```
ibrajer: Nit - named types:
```
mapping(bytes32 donLabel => EnumerableSet.Bytes32Set workflowRIDs) private s_allDonRIDs;
mapping(address owner => EnumerableSet.Bytes32Set workflowRIDs) private s_allOwnerRIDs;
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -34,12 +42,52 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @dev workflowKey is a hash of (owner ∥ name ) and maps to a set of RIDs. It is used for filtering workflows based on just
+  /// the name, as there may be multiple workflows with the same name under a different tag.
+  mapping(bytes32 workflowKey => EnumerableSet.Bytes32Set) private s_workflowKeyToRIDs;
+  /// @dev workflowID ⇒ workflowRID. This mapping lets us enforce global uniqueness of workflowID while also allows us to lookup
+  /// workflows through the workflowID.
+  mapping(bytes32 workflowID => bytes32 workflowRID) private s_idToRID;
+
+  // Secondary indices for iteration / queries
+  mapping(address owner => EnumerableSet.Bytes32Set workflowIDs) private s_activeOwnerWorkflowRIDs; // owner -> workflowRID set
+  mapping(bytes32 donLabel => EnumerableSet.Bytes32Set workflowIDs) private s_activeDONWorkflowRIDs; // donLabel -> workflowRID set
```
ibrajer: Should be `workflowRIDs`, not `workflowIDs`, right?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -34,12 +42,52 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
   // ================================================================
   // |                         Events                               |
   // ================================================================
 
-  event AllowedSignersUpdatedV1(address[] signers, bool allowed);
-  event OwnershipLinkUpdatedV1(address indexed owner, bytes32 indexed proof, bool indexed added);
+  event AllowedSignersUpdated(address[] signers, bool allowed);
+  event OwnershipLinkUpdated(address indexed owner, bytes32 indexed proof, bool indexed added);
+  event WorkflowRegistered(
```
ibrajer, line 73: So we dropped the workflow artifact URLs from emitted events and now the Workflow DON will call the view function afterwards to fetch the URLs?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -177,24 +249,54 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  function setDONRegistry(address registry, uint256 chainSelector) external onlyOwner {
+    address old = s_donRegistry.registry;
+    s_donRegistry = DONRegistryConfig({registry: registry, chainSelector: chainSelector});
+    emit DONRegistryUpdated(old, registry, chainSelector);
+  }
+
+  /// @dev Reverts unless `account` is in the linked-owners set.
+  /// @param account Address to verify (usually msg.sender).
+  function _requireOwnerLinked(
+    address account
```
ibrajer: Nit: can we use `owner` instead? Just to stick to the naming convention used throughout the contract

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param  action    NONE / PAUSE_WORKFLOWS / REMOVE_WORKFLOWS (expects action ≠ NONE)
+  function _handlePreUnlinkAction(address owner, PreUnlinkAction action) internal {
+    EnumerableSet.Bytes32Set storage active = s_activeOwnerWorkflowRIDs[owner];
+    // ------------- PAUSE or DELETE -------------
+    // Iterate from the back since EnumerableSet.remove() swaps-and-pops.
+    while (active.length() > 0) {
+      bytes32 rid = active.at(active.length() - 1);
+      if (action == PreUnlinkAction.PAUSE_WORKFLOWS) {
+        WorkflowMetadata storage rec = _getWorkflowFromStorage(msg.sender, rid);
+        _applyPause(rid, rec);
```
ibrajer, line 663: Why not `_pauseWorkflow` if there's `_deleteWorkflow` already?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param  owner     The owner being unlinked.
+  /// @param  action    NONE / PAUSE_WORKFLOWS / REMOVE_WORKFLOWS (expects action ≠ NONE)
+  function _handlePreUnlinkAction(address owner, PreUnlinkAction action) internal {
+    EnumerableSet.Bytes32Set storage active = s_activeOwnerWorkflowRIDs[owner];
+    // ------------- PAUSE or DELETE -------------
+    // Iterate from the back since EnumerableSet.remove() swaps-and-pops.
+    while (active.length() > 0) {
+      bytes32 rid = active.at(active.length() - 1);
+      if (action == PreUnlinkAction.PAUSE_WORKFLOWS) {
+        WorkflowMetadata storage rec = _getWorkflowFromStorage(msg.sender, rid);
```
ibrajer: Warning: Please use `owner` instead of `msg.sender` - anyone can call this function and it doesn't have to be the actual workflow owner

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    /* ───────────────────────── 4. UPDATE INDICES ─────────────────────── */
+    s_workflowKeyToRIDs[wKey].add(rid);
+    s_idToRID[workflowID] = rid;
+    s_allDonRIDs[donLabel].add(rid);
+    s_allOwnerRIDs[msg.sender].add(rid);
+
+    if (status == WorkflowStatus.ACTIVE) {
+      s_activeDONWorkflowRIDs[donLabel].add(rid);
+      s_activeOwnerWorkflowRIDs[msg.sender].add(rid);
+      s_userDONCount[msg.sender][donLabel] += 1;
```
ibrajer: Why are we incrementing only for active workflows? Isn't the limit set to a number of workflows in general?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
+    /* ─────── 2. PRIMARY-KEY REMAP ─────── */
+    delete s_idToRID[rec.workflowID];
+    s_idToRID[newWorkflowID] = rid;
+
+    /* ─────── 3. FIELD PATCHES ─────── */
+    bytes32 oldWorkflowID = rec.workflowID;
+    rec.workflowID = newWorkflowID;
+    if (!Strings.equal(rec.binaryURL, newBinaryURL)) rec.binaryURL = newBinaryURL;
+    if (!Strings.equal(rec.configURL, newConfigURL)) rec.configURL = newConfigURL;
```
ibrajer: Why this check instead of simply setting the values?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    _requireOwnerLinked(msg.sender);
+    _activateWorkflow(s_idToRID[workflowID]);
+  }
+
+  function batchPauseWorkflows(
+    bytes32[] calldata workflowIDs
+  ) external {
+    uint256 n = workflowIDs.length;
+    if (n == 0) revert EmptyUpdateBatch();
+    if (n > MAX_UPDATE_BATCH_COUNT) revert BatchTooLarge(n, MAX_UPDATE_BATCH_COUNT);
```
ibrajer: Do we really want to use a hardcoded limit if the user is willing to pay for gas to pause more workflows than this number? Maybe a batch size override can be an optional parameter?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
+    _requireOwnerLinked(msg.sender);
+
+    for (uint256 i; i < n; ++i) {
+      bytes32 rid = s_idToRID[workflowIDs[i]];
```
ibrajer: What if someone injects invalid workflow IDs? Then RID would be empty bytes, maybe you should check that before attempting to pause the workflow?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  ///         modified.
+  ///
+  /// @param  workflowIDs  Array of workflowIDs to activate.
+  ///                      Already-active IDs are silently ignored (no-ops).
+  function batchActivateWorkflows(
+    bytes32[] calldata workflowIDs
+  ) external {
+    uint256 n = workflowIDs.length;
+    if (n == 0) revert EmptyUpdateBatch();
+    if (n > MAX_UPDATE_BATCH_COUNT) revert BatchTooLarge(n, MAX_UPDATE_BATCH_COUNT);
```
ibrajer: See question above: https://github.com/smartcontractkit/chainlink-evm/pull/87/files#r2150504133

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
+    _requireOwnerLinked(msg.sender);
+
+    // ----- per-DON deltas
+    bytes32[] memory dons = new bytes32[](n);
+    uint32[] memory inc = new uint32[](n);
+    uint256 unique = 0;
+
+    for (uint256 i; i < n; ++i) {
+      bytes32 rid = s_idToRID[workflowIDs[i]];
```
ibrajer: Beware of invalid workflow IDs here

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        dons[unique] = don;
+        inc[unique] = 1;
+        ++unique;
+      }
+    }
```
ibrajer: Can we replace this with `EnumerableMap.Bytes32ToUintMap`?
```
for (uint256 i; i < n; ++i) {
      bytes32 rid = s_idToRID[workflowIDs[i]];
     WorkflowMetadata storage rec = _getWorkflowFromStorage(msg.sender, rid);
      if (rec.status == WorkflowStatus.ACTIVE) continue;
      (bool exists, uint256 counter) = map.tryGet(rec.donLabel);
      if (!exists) {
        map.set(rec.donLabel, 1);
      } else {
        map.set(rec.donLabel, ++counter);
      }
 }
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    if (nameLen == 0) revert WorkflowNameRequired();
+    if (nameLen > MAX_WORKFLOW_NAME_LENGTH) revert WorkflowNameTooLong(nameLen, MAX_WORKFLOW_NAME_LENGTH);
+  }
+
+  /// @notice Ensures the given workflowID is unique and non-zero.
+  /// @param workflowID The workflowID to validate and consume.
+  function _validateWorkflowID(
+    bytes32 workflowID
+  ) private view {
+    if (workflowID == bytes32(0)) revert InvalidWorkflowID();
```
ibrajer: Maybe add `workflowID` in the error?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: not for tracking purposes only, but for storage and primary key. Otherwise, every update will have to be a full new workflow storage write. If we use workflowID directly, then we no longer have a concept of "update" or of changing only the fields that changed. all changes would have to be full workflow deploys.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: Yup! I believe with the new storage service the URL might even be optional in some cases. I think the BinaryURL required was removed here in v2 compared to v1 as well

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: the difference between the two is that applyPause is a direct update without any checks. Since we already had to do a lot of validation the functions are split into 3 internal/private ones so that only the necessary steps are called depending on from which function and no redundant checks are performed without sacrificing too much readability (as opposed to just having no helpers and everything directly custom for each, which is itself potentially error prone is you miss calling a step)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: _getWorkflowFromStorage checks that it's the owner. Otherwise we can do a direct s_workflow[rid] to retrieve, but the helper checks for valid addresses owned by the workflow owner

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: I think if we only incremented for all workflows it can become problematic especially if we're just pausing old ones for people. As far as the don and engines are concerned, only active ones take up resources as well and need to be limited.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: sure, I just didn't want this to blow up if they sent too many. I can add an override param to this though

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: `_pauseWorkflow` calls `_getWorkflowFromStorage` which checks that the workflow exists before it can get it, so an invalid workflowID should revert with `WorkflowDoesNotExist`

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: ^same as above

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: but it's bytes32(0) :D ... maybe I'll just change the name of this error. Otherwise, error parameters also add to the contract size (error with no params are the smallest)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: prevents unnecessary storage writes that cost significantly more gas than just checking here

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I mean, you have to change it to this:
```
WorkflowMetadata storage rec = _getWorkflowFromStorage(owner, rid);
```

Unlinking works in such a way that anyone may unlink as long as they provide the correct signature that contains the original proof. The transaction doesn't have to be initiated by the real workflow owner. This is just a countermeasure in case the workflow owner's private key has been lost or compromised.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -427,4 +509,682 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  ) internal view returns (WorkflowMetadata storage workflow) {
+    workflow = s_workflows[rid];
+
+    if (workflow.owner == address(0)) revert WorkflowDoesNotExist();
+    if (workflow.owner != sender) revert CallerIsNotWorkflowOwner(sender);
+
+    return workflow;
+  }
+
+  function pauseWorkflow(
```
ibrajer, line 941: Another ask: can we group external functions in one section and internal ones in another section? What does the style guide say about it?

eutopian: It actually said not to do that and to keep them in logic sections 😅 https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/STYLE_GUIDE.md#comments

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: ah I see what you're saying this is the helper function

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: attributes are also limited to 512 bytes I believe. This was also just a somewhat reasonable number I picked but we can change this to 1024 if people feel like it should allow for more

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: We can’t declare a mapping(bytes32=>uint32) or an EnumerableMap.Bytes32ToUintMap in memory, so we would have to create a permanent storage to use for this if we want to avoid the manual counters

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -5,17 +5,24 @@ import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
...
 
-  // ================================================================
-  // |                 Internal variables                            |
-  // ================================================================
   string public constant override typeAndVersion = "WorkflowRegistry 2.0.0-dev";
+  uint8 private constant MAX_WORKFLOW_NAME_LENGTH = 64;
+  uint8 private constant MAX_WORKFLOW_TAG_LENGTH = 32;
+  uint8 private constant MAX_URL_LENGTH = 200;
+  uint8 private constant MAX_PAGINATION_LIMIT = 100;
+  uint16 private constant MAX_ATTRIBUTES_LENGTH = 512; // 0.5 kilobyte
```
ibrajer: Q: Have we considered having these as configurable parameters?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    );
+
+    emit WorkflowPaused(workflowId, owner, donLabel, workflowName);
+  }
+
+  /// @notice Returns a page of events along with the total event count.
+  /// @param start Zero-based index of the first event to include in the page.
+  /// @param limit  Maximum number of events to return in this page.
+  /// @return list  Array of events in the requested window.
+  function getEvents(uint256 start, uint256 limit) external view returns (EventRecord[] memory list) {
```
ibrajer, line 780: Q: I fail to see the purpose of storing state updates as events on the contract side? We are emitting events for every state change (and we should) and any off-chain service can reconstruct the historical trail of state changes based on these events, by simply filtering them out from RPC data.

Or better to ask: who will use this function? What is its purpose?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -368,36 +468,24 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  function getLinkedOwners(uint256 start, uint256 limit) external view returns (address[] memory owners) {
     uint256 total = s_linkedOwners.length();
-    if (start >= total) {
-      return new address[](0);
-    }
-
-    uint256 end = (start + batchSize > total) ? total : start + batchSize;
-    uint256 length = end - start;
-    address[] memory owners = new address[](length);
+    (uint256 offset, uint256 count) = _getPageBounds(total, start, limit);
```
ibrajer: Warning: Please don't change the original behavior. This has far larger implications on the off-chain system. I don't want this function to be limited by `MAX_PAGINATION_LIMIT`. It's not the same if you're returning an array of addresses (primitive type) and an array of workflow metadata (complex struct). If we limit this function, the off-chain service that relies on it will struggle to keep up with updates. The caller should impose the batch limit, and if the limit hits the EVM constraints or RPC call limitations, that's up to the caller to resolve/fix.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  }
+
+  function _enforceLimits(address owner, bytes32 donHash, string memory donLabel) internal view {
+    uint32 donLimit = s_cfg.donLimit[donHash].value;
+    uint32 userLimit = donLimit;
+    if (s_cfg.userDONOverride[owner][donHash].enabled) {
+      userLimit = s_cfg.userDONOverride[owner][donHash].value;
+    }
+
+    if (s_activeDONWorkflowRids[donHash].length() > donLimit) {
```
ibrajer: Q: Is this necessary if the user limit will always be smaller or equal to the DON limit? You are already setting `userLimit = donLimit`, which will work even when this particular owner address doesn't have a user override set.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Another note: Going back to this topic. Networking is always the bottleneck - we all know that by now. So when the registry will have tens or hundreds of thousands of workflows, it will become difficult for a new node to join the Workflow DON (or even on node restart, the same would happen), because the way the node catches up is by relying on view functions. I'm thinking if we should generally throw this responsibility to the caller? Wdyt?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @dev    Performs the following steps in order:
+  ///            1. Load the workflow and validate existence & ownership via `_getWorkflowFromStorage`.
+  ///            2. Revert if its status is already `PAUSED`, preventing redundant calls.
+  ///            3. Apply the pause change by calling `_applyPause`.
+  /// @param  rid  The workflow reference ID of the workflow record to pause.
+  /// @param  owner The address of the owner of the workflow.
+  function _pauseWorkflow(address owner, bytes32 rid) internal {
+    WorkflowMetadata storage rec = _getWorkflowFromStorage(owner, rid);
+    if (rec.status == WorkflowStatus.PAUSED) {
+      revert WorkflowAlreadyInDesiredStatus();
```
ibrajer: Q: Do we really want to revert here? If this is coming from a batch function, maybe it will be a mix of workflows in different states.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    _activateWorkflow(msg.sender, s_idToRid[workflowId]);
+  }
+
+  /// @notice Pauses multiple workflows owned by `msg.sender`.
+  /// @dev    There is no enforced batch size limit here.
+  ///         **User Risk:** Submitting a very large array of `workflowIds`
+  ///         may cause the transaction to run out of gas and revert.
+  ///         Clients should cap the array length to a safe value based on the current gas limits.
+  /// @param  workflowIds Array of workflow IDs to pause; must not be empty.
+  function pauseWorkflows(
```
ibrajer: Q: Below you have `batchActivateWorkflows` - consider sticking to one naming convention :D

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+      uint32 add = inc[k];
+      string memory donStr = donStrs[k];
+
+      uint32 donLimit = s_cfg.donLimit[donHash].value;
+      uint32 userLimit = donLimit;
+      if (s_cfg.userDONOverride[msg.sender][donHash].enabled) {
+        userLimit = s_cfg.userDONOverride[msg.sender][donHash].value;
+      }
+
+      if (s_activeDONWorkflowRids[donHash].length() + add > donLimit) {
```
ibrajer: Q: Again, is comparing with the DON limit really necessary if the user limit can never be above the DON limit?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+        revert MaxWorkflowsPerDONExceeded(donStr);
+      }
+      if (s_userDONCount[msg.sender][donHash] + add > userLimit) {
+        revert MaxWorkflowsPerUserDONExceeded(msg.sender, donStr);
+      }
+    }
```
ibrajer, line 1077: Q: I see, so this is "all or nothing" approach, either the entire batch must succeed, but if it's one workflow too much, it will revert? Maybe you should make this more explicit in the comment above the function.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param  workflowIds Array of workflow IDs to activate; must not be empty.
+  function batchActivateWorkflows(
+    bytes32[] calldata workflowIds
+  ) external {
+    uint256 n = workflowIds.length;
+    if (n == 0) revert EmptyUpdateBatch();
+
+    _requireOwnerLinked(msg.sender);
+
+    // ----- per-DON deltas
```
ibrajer, line 1023: Nit: Consider either wrapping these segments into internal functions with appropriate names or adding more comments to describe the flow here. It does take a bit of cognitive load to read this code and understand what's going on.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Okay, I guess, unless you want to be very strict, just like in the case of `batchActivateWorkflows`- it's "all or nothing" approach

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Another Q: Why is there no "batch remove" function?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    delete s_idToRid[rec.workflowId];
+
+    emit WorkflowDeleted(rec.workflowId, rec.owner, rec.donLabel, rec.workflowName);
+    delete s_workflows[rid];
+  }
+
+  // ================================================================
+  // |                       Admin Workflow                         |
+  // ================================================================
+  function adminPauseWorkflow(
```
ibrajer, line 1256: Q: Can you add a comment above to explain why we need this function (emergency suspension)? Also, should we consider pausing more than one workflow ID at once? In case there really is a s***storm?

ibrajer: Nvm for the batch function, I see you added several of them below ;)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+      adminPauseWorkflow(workflowIds[i]);
+    }
+  }
+
+  /// @notice Pauses *all* active workflows for a given user, as an administrator.
+  /// @dev    - Only the contract owner may call this (`onlyOwner`).
+  ///         - Iterates from the end of the user’s active-workflow set and directly pauses each workflow.
+  /// @param  user The address whose active workflows should be paused.
+  function adminPauseAllByUser(
+    address user
```
ibrajer: Q: Should we call it `user` or `owner`? We kinda mix both in different contexts 😅

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  ///          (`s_workflowKeyToRids`).  When `limit` is 0 or exceeds
+  ///          `MAX_PAGINATION_LIMIT`, the call clamps it to that constant.
+  ///          Does **not** revert on out-of-range pagination: if `start` is
+  ///          beyond the end of the set the function returns an empty array.
+  /// @param   owner         Address that registered the workflows.
+  /// @param   workflowName  Case-sensitive name string (≤ 64 chars).
+  /// @param   start         Zero-based index into the RID set.
+  /// @param   limit         Max #records to return (clamped to 100).
+  /// @return  list          Array of `WorkflowMetadata`.
+  function getWorkflowMetadataListByOwnerName(
```
ibrajer: Q: `...ByOwnerName` is a bit misleading - owner's name? Maybe `OwnerAndName`?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    _validateWorkflowURLs(bytes(binaryUrl).length, bytes(configUrl).length);
+    _validateAttributesLength(attributes.length);
+
+    bytes32 wKey = _workflowKey(msg.sender, workflowName);
+    bytes32 donHash = _donHash(donLabel);
+
+    /* ───────────────────────── 1. HOUSEKEEPING ───────────────────────── */
+    // we need to do this first, or there may be extra workflows occupying the limit
+    if (!keepAlive) {
+      _pauseAllActiveWorkflowsByWorkflowKey(wKey, donHash);
```
ibrajer: Warning: If someone attempts to register a workflow on a non-existent DON, they can do so by setting the `status` to `PAUSED`, and then we'll have an `s_allDONRids` index that contains this DON. We should ideally prevent this scenario by introducing a validation check for the `donLabel` parameter, just like we did for all the others.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,940 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    }
+
+    /* ───────────────────────── 2. LIMIT CHECKS ───────────────────────── */
+    if (status == WorkflowStatus.ACTIVE) {
+      _enforceLimits(msg.sender, donHash, donLabel);
+    }
+
+    /* ───────────────────────── 3. WRITE PRIMARY RECORD ───────────────── */
+    s_workflows[rid] = WorkflowMetadata({
```
ibrajer: Q: You said that RID is a reference ID made out of workflow owner, workflow name, and label. We are intentionally not taking DON labels into account and making sure that RID is globally unique, in a way that it is not allowed to run the same workflow on different DONs? Is that correct? Question also directed at @DeividasK

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: > If someone attempts to register a workflow on a non-existent DON, they can do so by setting the status to PAUSED

true. but they would never be able to activate it. so the question is would it matter? `_enforceLimits` will ensure that all active workflows are for valid dons

another question for @DeividasK to weigh in on

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: So the reason why I left this function completely inline and other functions with internals is because this logic never gets reused, so more internal functions would just be overhead. But I can definitely add in more comments

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: True, and I think that's especially fine for a batchActivate. If you want to activate 10 workflows, we should check ahead of time that there are 10 spaces in the DON for it. If not, we shouldn't just activate 2 and then reject the rest if they don't get to pick which of the 2.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: good point! I think that when we didn't explicitly say user limits can't go over the DON it made sense but in the current scenario no. Let me remove them

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: the functions are called in different places for different uses. If you are just updating the status we do want this check to prevent unnecessary writes. However the batch functions are already using an active set of workflows so we don't need to make this check, and also why those functions use the `_applyPause` (to perform an underlying pause with zero validations because we've already validated elsewhere). 

I can see if the comments are more clear about this but each of the different internal state changing functions have comments on when they should be used

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: are you suggesting we remove the pagination entirely?

eutopian: I'm okay with removing the MAX_PAGE limit for all of them if that works better. I moved these to a helper because we had a lot of pagination with the exact same bound calculations and this made it easier to review the logic so bugs can stay limited to one place (if there was an index issue etc) instead of everywhere there is a pagination, as there could also be copy and paste errors.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: On the one hand I see it could be nice to configure. On the other hand these also seem like reasonable limits that I'm not sure when we couldn't work around it and absolutely had to go over.

DeividasK: We need to make these configurable. E.g., the attributes length certainly looks too small for me. But I don't know how big we'll need to get to :)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: > any off-chain service can reconstruct the historical trail of state changes based on these events

We want to store on chain so we don't have to rely on archive nodes for these events.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,918 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @param tag   Unique tag for this workflow variant.
+  /// @return rid   Keccak256 hash of `owner`, `name`, and `tag`.
+  function _workflowRid(address owner, string memory name, string memory tag) internal pure returns (bytes32 rid) {
+    return keccak256(abi.encode(owner, name, tag));
+  }
+
+  /// @notice Hashes an arbitrary-length DON label string into a fixed-size bytes32.
+  /// @param str   DON label as a UTF-8 string.
+  /// @return hash  Keccak256 hash of `bytes(str)`.
+  function _donHash(
```
DeividasK:
```suggestion
  function _hash(
```
This is generic 🤷‍♂️

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: > true. but they would never be able to activate it. so the question is would it matter?

I have to disagree with this because it certainly matters from the customer's perspective. By allowing the user to register the workflow, you are setting an expectation that this workflow may be activated in the future. When that doesn't happen, it will only lead to frustration because the only thing the user can do in that case is remove the workflow. If we can add a few lines of code to prevent this bad UX, I think we should do it.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: This is up to you, but note that this function largely differs from all the other ones, because you used a lot of internal functions everywhere, and then all of a sudden, here you don't. Almost feels like this code was written by a different person. 😅

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,918 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    WorkflowMetadata storage rec = _getWorkflowFromStorage(msg.sender, rid);
+
+    /* ─────── 2. PRIMARY-KEY REMAP ─────── */
+    delete s_idToRid[rec.workflowId];
+    s_idToRid[newWorkflowId] = rid;
+
+    /* ─────── 3. FIELD PATCHES ─────── */
+    bytes32 oldWorkflowId = rec.workflowId;
+    rec.workflowId = newWorkflowId;
+    if (!Strings.equal(rec.binaryUrl, newBinaryUrl)) rec.binaryUrl = newBinaryUrl;
```
DeividasK:
```suggestion
    if (_hash(rec.binaryUrl) != _hash(newBinaryUrl)) rec.binaryUrl = newBinaryUrl;
```
To avoid a Strings dependency

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I agree with using the shared pagination function and it makes sense for the reasons you listed above, I'm asking if we should ditch the `MAX_PAGINATION_LIMIT` completely and let the caller decide?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,918 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    string calldata configUrl,
+    bytes calldata attributes,
+    bool keepAlive
+  ) external {
+    bytes32 rid = _workflowRid(msg.sender, workflowName, tag);
+    if (s_workflows[rid].owner == address(0)) {
+      _createWorkflow(rid, workflowName, tag, workflowId, status, donLabel, binaryUrl, configUrl, attributes, keepAlive);
+    } else {
+      _updateWorkflow(rid, workflowName, workflowId, binaryUrl, configUrl, attributes);
+    }
```
DeividasK: I would inline this logic. This will allow you to reduce lines of code by pulling out shared operations (like validation)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Okay, I don't have a full context on that, but is this supposed to be "Capacity changing events for deterministic workflow assignment to be added to the workflow registry" mentioned in the doc?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: If you bake it in, it's game over as soon as your understanding of reasonable limits falls short. I always prefer making these settings configurable rather than shooting myself in the foot a couple of months down the line. 😅 And having a generic `setConfig` on this contract shouldn't be a big lift, right?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: Yes, we should NOT include DON labels in workflow ID calculations. Workflows should be freely reassignable across DON labels (without any change to the workflow ID or other associated data). I'd go as far as to say we need a separate function that simply allows reassigning a workflow to a different DON label.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: I wouldn't revert here. If I want to pause and the workflow is already paused, it is not an error.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,918 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+
+  /// @notice Internal helper to pause a single workflow, enforcing ownership.
+  /// @dev    Performs the following steps in order:
+  ///            1. Load the workflow and validate existence & ownership via `_getWorkflowFromStorage`.
+  ///            2. Revert if its status is already `PAUSED`, preventing redundant calls.
+  ///            3. Apply the pause change by calling `_applyPause`.
+  /// @param  rid  The workflow reference ID of the workflow record to pause.
+  /// @param  owner The address of the owner of the workflow.
+  function _pauseWorkflow(address owner, bytes32 rid) internal {
+    WorkflowMetadata storage rec = _getWorkflowFromStorage(owner, rid);
```
DeividasK: This is doing owner checks that are also done by the functions that call `_pauseWorkflow`. I would inline `_pauseWorkflow` to reduce this.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -426,4 +514,918 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+    emit WorkflowUpdated(oldWorkflowId, newWorkflowId, msg.sender, rec.donLabel, workflowName);
+  }
+
+  /// @dev Reverts if the record does not exist or is not owned by `sender`.
+  /// @param sender  Expected owner of the workflow (typically `msg.sender`).
+  /// @param rid     Workflow reference-ID (hash of owner ∥ name ∥ tag).
+  /// @return workflow  Storage pointer to the requested `WorkflowMetadata`.
+  /// @custom:reverts WorkflowDoesNotExist        If no record is found for `rid`.
+  /// @custom:reverts CallerIsNotWorkflowOwner    If `sender` ≠ `workflow.owner`.
+  function _getWorkflowFromStorage(
```
DeividasK: I wouldn't create this level of abstractions. Getting a workflow should simply be
```
workflow = s_workflows[rid];
```

Associated checks would be performed contextually (right now, they are often duplicated).

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -33,12 +39,49 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @dev workflowId ⇒ workflowRid. This mapping lets us enforce global uniqueness of workflowId while also allows us to lookup
+  /// workflows through the workflowId.
+  mapping(bytes32 workflowId => bytes32 workflowRid) private s_idToRid;
+
+  // Secondary indices for iteration / queries
+  mapping(address owner => EnumerableSet.Bytes32Set workflowRids) private s_activeOwnerWorkflowRids; // owner -> workflowRid set
+  mapping(bytes32 donHash => EnumerableSet.Bytes32Set workflowRids) private s_activeDONWorkflowRids; // donHash -> workflowRid set
+
+  /// @dev Every workflow ever registered under `donLabel` as donHash. Pruned only on delete.
+  mapping(bytes32 donHash => EnumerableSet.Bytes32Set workflowRids) private s_allDONRids;
```
DeividasK, line 52:
```suggestion
  mapping(bytes32 donLabelHash => EnumerableSet.Bytes32Set workflowIds) private s_DONLabelHashToWorkflows;
```

You don't need `rid` that includes the label when you are storing `label => workflowID` associations.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -33,12 +39,49 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
   mapping(bytes32 proof => bool used) private s_usedProofs;
 
+  /// @dev workflowRid (reference ID) is a hash of (owner ∥ name ∥ label). It functions as the primary index for the workflow storage.
+  mapping(bytes32 workflowRid => WorkflowMetadata workflowMetadata) private s_workflows;
+  /// @dev workflowKey is a hash of (owner ∥ name ) and maps to a set of Rids. It is used for filtering workflows based on just
+  /// the name, as there may be multiple workflows with the same name under a different tag.
+  mapping(bytes32 workflowKey => EnumerableSet.Bytes32Set) private s_workflowKeyToRids;
+  /// @dev workflowId ⇒ workflowRid. This mapping lets us enforce global uniqueness of workflowId while also allows us to lookup
+  /// workflows through the workflowId.
+  mapping(bytes32 workflowId => bytes32 workflowRid) private s_idToRid;
```
DeividasK, line 45:
```suggestion
  mapping(bytes32 workflowId => bool isActive) private s_activeWorkflowIds;
```
Wouldn't this be sufficient?

DeividasK: This could potentially be nested in WorkflowMetadata

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -33,12 +39,49 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   /// enforce an invariant on proofs.
   mapping(bytes32 proof => bool used) private s_usedProofs;
 
+  /// @dev workflowRid (reference ID) is a hash of (owner ∥ name ∥ label). It functions as the primary index for the workflow storage.
+  mapping(bytes32 workflowRid => WorkflowMetadata workflowMetadata) private s_workflows;
```
DeividasK, line 39:
```suggestion
  mapping(bytes32 workflowId => WorkflowMetadata workflowMetadata) private s_workflows;
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: (Looks like workflow status is already there 🤷‍♂️ )

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: this can happen when dons get removed too though

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: > I'd go as far as to say we need a separate function that simply allows reassigning a workflow to a different DON label.

this was the last question I had. I left this out of the update just as I did for the status due to the calculations but I think it's simple enough to add a separate `updateDONLabel` function

timothyF95 (dismissed):

timothyF95 (approved):

ibrajer (approved):
Approval to unblock other work - let's get code coverage up to 93/95 in the next PR

PR #95, opened 2025-06-09
Capabilities Registry v2

DeividasK:
**IMPORTANT:** See commits for a diff by change.

## Core changes
- Added the concept of a DON family with getters and setters.
- Added DON name and generic config.
  - Added Update/Get/Remove `DONByName` functions.
- Added `getHistoricalDONInfo` to fetch historical configurations.
  - Moved `isPublic`, `f`, `name`, and `config` variables into this preserved configuration.
- Renamed `DONCapabilityConfig` to `MutableDONConfig` to better reflect the purpose.
- [Replaced addDON function with addDONs function](https://github.com/smartcontractkit/chainlink-evm/pull/95/commits/03cf3f334a0e5b2fbd007aa4571eb6ad3aa4d216)
- Replaced [hashedCapabilityId with capabilityId](https://github.com/smartcontractkit/chainlink-evm/pull/95/commits/98edf29ed4f4cf2a7adc1e95eda9cfcc2f1517f0) for debugging and readability improvements.
- [Removed enums from Capabilities structs in favour of generic bytes metadata](https://github.com/smartcontractkit/chainlink-evm/pull/95/commits/77d99366398201dcb49e793f1cefa5e379f4cc4c)
- [Refactored updateDON function params into a struct](https://github.com/smartcontractkit/chainlink-evm/pull/95/commits/ca27c443de0bdee8c86e5cf0345dd43c585a22ef)

## Notes
- [Associated design document](https://docs.google.com/document/d/1tRZFmhl66-5psRlo8_pkeoRc5EDZXss596p6L5AJKKw/edit?tab=t.0#heading=h.rjds689xtlkq).


kylesmartin (commented):
s_nextDONId still manages the DonID that ultimately gets passed to beforeCapabilityConfigSet, leading to the issues outlined here: https://docs.google.com/document/d/1C-Dp9ljVIFdPP5dTcs-SGqG0KvKA_ddMQ04qOfWtW-U/edit?tab=t.0#heading=h.4e7ng0tekbkc

Would be helpful to have this resolved, where DonID can be some arbitrary uint32 instead of being tied to a nonce.

DeividasK:
> s_nextDONId still manages the DonID that ultimately gets passed to beforeCapabilityConfigSet, leading to the issues outlined here: https://docs.google.com/document/d/1C-Dp9ljVIFdPP5dTcs-SGqG0KvKA_ddMQ04qOfWtW-U/edit?tab=t.0#heading=h.4e7ng0tekbkc
> 
> Would be helpful to have this resolved, where DonID can be some arbitrary uint32 instead of being tied to a nonce.

You can call `addDONs` with unique DON names and then call `updateDONByName`. I thought this would resolve the issue you linked. Am I missing something? `name` is effectively a predictable ID that is enforced to be unique in the registry.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -158,24 +113,39 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
-    /// @notice Mapping from hashed capability IDs to configs
-    mapping(bytes32 capabilityId => bytes config) capabilityConfigs;
+    /// @notice True if the DON is public. A public DON means that it accepts
+    /// external capability requests
+    bool isPublic;
+    /// @notice The f value for the DON.  This is the number of faulty nodes
+    /// that the DON can tolerate. This can be different from the f value of
+    /// the OCR instances that capabilities spawn.
+    uint8 f;
+    /// @notice The name of the DON. Can be empty. If not empty, must be unique
```
krehermann, line 149: why allow it to be empty? it will cause a lot of headaches in deployments because the desire will be to find by name. will be much simpler to make the name required

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -158,24 +113,39 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
   struct CapabilityConfiguration {
     /// @notice The capability Id
-    bytes32 capabilityId;
+    string capabilityId;
     /// @notice The capability config specific to a DON.  This will be decoded offchain
     bytes config;
   }
 
-  struct DONCapabilityConfig {
+  struct MutableDONConfig {
```
krehermann, line 134: code comment please. `mutable` meaning is not clear to. is it the don that's mutable or this config or both?

isn't obvious that don's are mutable, what does the prefix communicate?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -228,8 +201,35 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
+    CapabilityConfiguration[] capabilityConfigurations;
+    bytes32[] nodes;
+    uint8 f;
+    bool isPublic;
+    bool acceptsWorkflows;
+  }
+
+  /// @notice UpdateDONParams is a struct that holds the parameters for updating a DON.
+  struct UpdateDONParams {
+    string name;
```
krehermann, line 225: how can this work if name is optionally empty for multiple dons? seems like there is a gap in the logic

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -158,24 +113,39 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
+    /// @notice Metadata for the capability. This is used to store additional information about the capability.
+    bytes metadata;
   }
 
   /// @notice CapabilityConfiguration is a struct that holds the capability configuration
   /// for a specific DON
   struct CapabilityConfiguration {
     /// @notice The capability Id
-    bytes32 capabilityId;
+    string capabilityId;
```
krehermann, line 126: is the value of capabilityId now user defined as opposed to derived?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -185,19 +155,12 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
     uint32 configCount;
-    /// @notice The f value for the DON.  This is the number of faulty nodes
-    /// that the DON can tolerate. This can be different from the f value of
-    /// the OCR instances that capabilities spawn.
-    uint8 f;
-    /// @notice True if the DON is public. A public DON means that it accepts
-    /// external capability requests
-    bool isPublic;
     /// @notice True if the DON accepts Workflows. A DON that accepts Workflows
     /// is called Workflow DON and it can process Workflow Specs. A Workflow
```
krehermann: nit: `process Workflow Specs` -> `execute Workflows`

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -216,6 +179,16 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
     bool acceptsWorkflows;
     /// @notice List of member node P2P Ids
     bytes32[] nodeP2PIds;
+    /// @notice The family of the DON. A DON family is a group of DONs that
+    /// are connected with each other. Can be empty if the DON is not part of
+    /// any family.
+    string donFamily;
```
krehermann: what is responsible for managing the possible values of donFamily? is there a way to list the known set to the  registry? if not, let's add it because deployments will want to validate the adding a don to an existing family and obtain a clear definitive answer from the contract

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -761,87 +824,96 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
-  /// @param f The maximum number of nodes that can fail
-  function updateDON(
-    uint32 donId,
-    bytes32[] calldata nodes,
-    CapabilityConfiguration[] calldata capabilityConfigurations,
-    bool isPublic,
-    uint8 f
-  ) external onlyOwner {
+  /// @param updateDONParams The parameters for the DON to update
+  function updateDON(uint32 donId, UpdateDONParams calldata updateDONParams) external onlyOwner {
```
krehermann, line 799: what is the flow to use this if i only know the don name?

DeividasK: There's an `updateDONByName` for this purpose.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: The value was always user-defined. We only had a consistent way to generate a hash from the capability ID. Now we would use the capability ID directly instead of going through a hashing process. Internally, the hash is still used.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: If the DON doesn't have a name, then `updateDONByName` won't find the DON. However, you can `updateDON` (by ID) and add a name, which is why `UpdateDONParams` has the name.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: The idea here was not to force DON names on people. This would help with the setup, as you wouldn't have to come up with unique DON names and could continue managing DONs by IDs if you wanted to. Making it a requirement seemed overly strict. You can convince me otherwise if you feel strongly about this.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: > is there a way to list the known set to the registry?

There wasn't. I will add it.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -853,35 +970,81 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
+    // There is no point in erroring out as this is a no-op and the erroring
+    // would not provide any value to the user.
+    if (_hash(s_donIdToDonFamily[donId]) == newDONFamilyHash) return;
+
+    // Set the DON family name hash to the new family name hash so it can be
+    // retrieved by the hash later. We do not need to clean up the old family
+    // name hash as it is only used to retrieve the family name by the hash.
+    s_donFamilyHashToDonFamily[newDONFamilyHash] = donFamily;
+
+    _removeDONFromFamily(donId);
```
ibrajer: I'm a bit confused now - according to the docs that you've shared, one DON can be a member of one or more DON families at the same time. But judging by this code, that is not the case?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -853,35 +970,81 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
+    if (configCount == 0) revert DONDoesNotExist(donId);
+    return _getDON(donId, configCount);
+  }
+
+  /// @notice Gets a historical DON's info
+  /// @param donId The DON ID
+  /// @param configCount The config count of the DON
+  /// @return DONInfo The DON's parameters
+  function getHistoricalDONInfo(uint32 donId, uint32 configCount) external view returns (DONInfo memory) {
+    if (s_dons[donId].configCount == 0) revert DONDoesNotExist(donId);
```
ibrajer: Nit: Some gas savings if you store `s_dons[donId].configCount` into a local variable

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -911,54 +1075,179 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
     // as not all DONs will run OCR instances.
-    if (donParams.f == 0 || donParams.f + 1 > nodes.length) revert InvalidFaultTolerance(donParams.f, nodes.length);
+    if ((!i_canAddOneNodeDONs && donParams.f == 0) || donParams.f + 1 > nodes.length) {
+      revert InvalidFaultTolerance(donParams.f, nodes.length);
+    }
+
+    MutableDONConfig storage prevDONConfig = don.config[donParams.configCount - 1];
+
+    // Check if the DON name is changing
+    if (keccak256(bytes(prevDONConfig.name)) != keccak256(bytes(donParams.name))) {
```
ibrajer: Nit: I thought you wanted to use `_hash(...)` everywhere?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -228,8 +209,42 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
+  /// @notice NewDONParams is a struct that holds the parameters for a new DON.
+  struct NewDONParams {
+    string name;
+    string donFamily;
+    bytes config;
+    CapabilityConfiguration[] capabilityConfigurations;
+    bytes32[] nodes;
+    uint8 f;
+    bool isPublic;
+    bool acceptsWorkflows;
```
ibrajer, line 220: `acceptWorkflows=true` means it's the Workflow DON and `acceptWorkflow=false` is Capability DON? Are we anticipating any other types of DONs in the future? Should we use an enum here instead?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.26;
 import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
 import {ICapabilityConfiguration} from "./interfaces/ICapabilityConfiguration.sol";
 
-import {OwnerIsCreator} from "../../../shared/access/OwnerIsCreator.sol";
+import {ConfirmedOwner} from "../../../shared/access/ConfirmedOwner.sol";
```
ibrajer: You can also check out `Ownable2StepMsgSender.sol`- remake of OZ Ownable2Step

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -228,8 +209,42 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
     bool acceptsWorkflows;
     uint8 f;
+    string name;
+    bytes config;
+  }
+
+  /// @notice NewDONParams is a struct that holds the parameters for a new DON.
+  struct NewDONParams {
+    string name;
+    string donFamily;
```
ibrajer: I wonder if it would be efficient to keep a mapping of DON family IDs to actual names somewhere and just store the ID here instead. If the DON family is removed, that doesn't reshuffle the IDs, it just removes the entry from the mapping and continues with the next available ID. I see the same string is used in the `DONInfo` struct above.
Although I see that you defined `mapping(bytes32 donFamilyHash => string donFamily)` below, so you could also place `bytes32` instead and reuse this existing mapping?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -158,24 +118,42 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
     bytes32[] capabilityIds;
-    /// @notice Mapping from hashed capability IDs to configs
-    mapping(bytes32 capabilityId => bytes config) capabilityConfigs;
+    /// @notice True if the DON is public. A public DON means that it accepts
+    /// external capability requests
+    bool isPublic;
+    /// @notice The f value for the DON.  This is the number of faulty nodes
+    /// that the DON can tolerate. This can be different from the f value of
+    /// the OCR instances that capabilities spawn.
+    uint8 f;
```
ibrajer, line 148: Nit: everything has a nice full name and then you throw `f` into the picture. I know it's OCR-specific, but would naming it `faultyNodesTolerance` confuse people even more?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: You are right, DONs can belong to multiple families. This was a miss on my part. I've pushed an update.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: We are not expecting any atm, no. Enum wouldn't help even if we did, as you can't dynamically add new enum values.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: We want an external interface to deal with strings to improve readability and debugging of deployed contracts. Dealing with hashes adds complexity and doesn't save that much in gas costs.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: I'd prefer to stick with `f` as it is already a well-known and prevalent variable name.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
RensR: Main difference: custom errors over string errors.

eutopian: :D I think new contracts should probably use `Ownable2StepMsgSender`... the other ones are only there to support existing ones or contracts that need to be on really old pragmas for some reason

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.26;
 import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
 import {ICapabilityConfiguration} from "./interfaces/ICapabilityConfiguration.sol";
```
eutopian, line 6: This is imported twice (L12)

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -542,15 +587,18 @@ contract CapabilitiesRegistry is INodeInfoProvider, OwnerIsCreator, ITypeAndVers
...
 
       uint32 capabilityConfigCount = storedNode.configCount;
       for (uint256 j; j < capabilityIds.length; ++j) {
-        if (!s_hashedCapabilityIds.contains(capabilityIds[j])) revert InvalidNodeCapabilities(capabilityIds);
-        storedNode.supportedHashedCapabilityIds[capabilityConfigCount].add(capabilityIds[j]);
+        bytes32 hashedCapabilityId = _hash(capabilityIds[j]);
+        if (!s_hashedCapabilityIds.contains(hashedCapabilityId)) {
+          revert InvalidNodeCapabilities(capabilityIds);
+        }
+        storedNode.supportedHashedCapabilityIds[capabilityConfigCount].add(hashedCapabilityId);
```
eutopian, line 601: question on L628
https://github.com/smartcontractkit/chainlink-evm/pull/95/files#diff-52785757607154a33b9f84655208e21d5e3eb50daffc26b94724a01b106fe8c1R628

the i here is referring to `removedNodeP2PIds` is that the same index we want for 
`NodePartOfCapabilitiesDON(uint32(node.capabilitiesDONIds.at(i)), p2pId);`

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
eutopian: @RensR how come the linter doesn't care

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: Good catch, fixed.

eutopian:
Just one more thing on the string for the donFamily, is it okay to not have a limit? or should some kind of hard limit be put in place "just in case" such as 32 or 64 length? I know when it's hashed we'll always have a bytes32 either way, but if the string is being saved 

the old labeledName did have a comment on a cap as well 
```
    string labelledName;
    /// @notice Semver, e.g., "1.2.3"
    /// @dev must be valid Semver + max 32 characters.
```

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: - [x] make the name required.

ibrajer (approved):

eutopian (approved):

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
RensR: linter bad

RensR: idk

PR #101, opened 2025-06-10
Add OZ 5.1.0 vendor contracts for EnumerableMap

eutopian:
Pulled from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.1.0/contracts

New version of EnumerableSet includes
```
 * - `uint256 -> bytes32` (`UintToBytes32Map`) since v5.1.0
 * - `address -> address` (`AddressToAddressMap`) since v5.1.0
 * - `address -> bytes32` (`AddressToBytes32Map`) since v5.1.0
 * - `bytes32 -> address` (`Bytes32ToAddressMap`) since v5.1.0
```
not available in existing vendored versions

Also added updates for:
```
cryptography/EDCSA.sol
cryptography/MessageHashUtils.sol
math/Math.sol
math/SafeCast.sol
math/SignedMath.sol
Strings.sol
```


ibrajer (dismissed):

RayXpub (dismissed):
We typically require a new version to have been available for at least 3 months, so 5.3.0 is a bit too recent. I’d recommend downgrading to 5.2.0, which should include everything you need. cc @RensR

eutopian:
I can even go down to 5.1.0 actually. That's the version that prompted this PR (there's actually a closed one with the wrong branch name due to this)

ibrajer (dismissed):

eutopian:
@RayXpub updated and simplified for now. Will revisit if we go with npm later depending on what DevRel says

ibrajer (approved):

RensR (approved):

PR #102, opened 2025-06-11
OEV-221: Fix Beholder test race condition

cl-efornaciari:
Using [this](https://github.com/smartcontractkit/chainlink-evm/blob/6ffed718392a30d6db3cbf67e67d9f282df07092/pkg/txm/metrics_test.go#L26) following snippet introduced a race condition in the CI

Modified to use the `BeholderTester` object and and removed parallelization to avoid a conflict of a global beholderEmitter object

### DIFF
--- a/pkg/txm/metrics_test.go
+++ b/pkg/txm/metrics_test.go
@@ -5,29 +5,22 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
-
-	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
+	beholderTests "github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
 	"github.com/smartcontractkit/chainlink-evm/pkg/testutils"
 	"github.com/smartcontractkit/chainlink-evm/pkg/txm/types"
 	svrv1 "github.com/smartcontractkit/chainlink-protos/svr/v1"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


jmank88 (dismissed):

pkg/txm/metrics_test.go
```diff
@@ -71,19 +52,12 @@ func TestEmitTxMessage(t *testing.T) {
...
-				}
-				return true
-			}),
-			"beholder_domain", "svr",
-			"beholder_entity", "svr.v1.TxMessage",
-			"beholder_data_schema", "/beholder-tx-message/versions/2",
-		)
+		messages := beholderTester.Messages(t)
+
+		assert.Len(t, messages, 1)
```
dimriou, line 57: Given that the two tests are run in parallel and `beholderTests.Beholder(t)` returns the same global object, I think it would be safer to match the message here exactly using `beholderTester.Len` or some other method I'm not aware of to avoid flakes.

cl-efornaciari: I removed the `t.Parallel()` call so they run serially. From the docs:
```
// Beholder sets the global beholder client as a message collector and returns a tester that provides helper assertion
// functions on received messages.
//
// Beholder affects the whole process, it cannot be used in parallel tests or tests with parallel ancestors.
```

jmank88 (approved):

dimriou (approved):

PR #103, opened 2025-06-11
Follow-ups to Workflow Registry PR #73

ibrajer:
Addressing follow-ups from the previous PR: https://github.com/smartcontractkit/chainlink-evm/pull/73
* append `typeAndVersion` to the signed message hash
* replace mapping and `EnumerableSet` with `EnumerableMap` now when the OZ libs v5.1.0 are available (replace both with `EnumerableMap.AddressToBytes32Map` instead)
* removed unnecessary `virtual` keyword, Workflow Registry should not be inherited from
* turned some functions into `external` ones, because it's not necessary for all of them to be `public` (gas savings)

Additionally:
* simplified linking as suggested by @RayXpub in the comment below
* introduced verifiable invariant for proofs as suggested by @timweri


contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -311,7 +309,7 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
     // are expecting that whoever is calling this function must have access to the original proof signed by the
     // trusted entity. The original proof will be provided only to the ones who has access to the account that
     // was used for linking the owner address.
-    bytes32 storedProof = s_ownerProofs[owner];
+    bytes32 storedProof = s_linkedOwners.get(owner);
     if (storedProof != proof) {
       revert OwnershipLinkProofDoesNotMatch(owner, proof, storedProof);
     }
```
RayXpub: q: why is this step necessary? since both `owner` and `proof` are publicly available, it doesn't look like this is providing any additional security. We may want to consider passing the stored proof as the pre image hash on L320 instead of a user provided proof instead?

ibrajer: In the original implementation, you could've provided a different proof for unlinking and that would still be acceptable, as long as the signature was valid. I think I might have tangled myself here a bit, since the introduction of the "request type" solved the most viable replay attack. So, I think you're right, if we drop the `proof` as a function parameter and simply use `storedProof` instead, that should be more than enough to complete the unlinking, as long as the valid signature is generated based on the original proof.

timweri: @ibrajer Can the unlink signature on the original proof be lost? Can the user refetch it any time from the CLI or the UI? If they can't, then locking the owner address to the stored proof would prevent users from unlinking if they lost the signature.

ibrajer: They fetch it via CLI/UI as soon as they initiate the unlinking request (original proof will be in the response). The only disaster scenario that I can think of is if the DB and all of its backups have been compromised or wiped out. But even then, the proof can be extracted from the transaction history on the chain and the signature can be reconstructed using a combination of different recovery mechanisms. Happy to chat about it in more detail if it's still unclear.

RayXpub (approved):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
RayXpub: @ibrajer we can probably also add a getter function?

ibrajer: You mean, getter function as in "give me the proof for owner X", for example:
```
function getProof(address owner) {
   return s_linkedOwners.get(owner); // omitting error handling for the purpose of demonstration
}
````

eutopian (approved):

PR #107, opened 2025-06-13
Add Botanix and Polygon Katana node configs

joaoluisam:


pavel-raykov (dismissed):

pavel-raykov (approved):

pkg/config/toml/defaults/Botanix_Mainnet.toml
```diff
@@ -0,0 +1,19 @@
+ChainID = '3637'
+FinalityTagEnabled = true
+LinkContractAddress = '0x30e85A5c9525AD9a7A0FA5C74df4Baf0b01aD241'
+
+[GasEstimator]
+EIP1559DynamicFees = true
+TipCapMin = '0 wei'
+TipCapDefault = '0 wei'
+PriceDefault = '100 wei'
+PriceMax = '800 kwei'
```
simsonraj, line 10: Looks way lower than CCIP pricemax?
https://github.com/smartcontractkit/internal-wiki/blob/remove-kroma/Products/CCIP/v1.5/configs/shared-config-chain.toml#L1333
we usually are very strict about the CCIP limits than DF limits, so this seems very restrictive

joaoluisam: I copied this value from the same place but on main branch: https://github.com/smartcontractkit/internal-wiki/blob/main/Products/CCIP/v1.5/configs/shared-config-chain.toml#L1328-L1336 believe that is an old branch?

simsonraj (approved):

PR #108, opened 2025-06-13
Add new LatestSafeBlock method to EVM RPC client

prashantkumar1982:
Add new method on EVM chain client for LatestSafeBlock(), which calls the RPC with the "Safe" tag.


dhaidashenko (dismissed):

CONFIG.md
```diff
@@ -20,6 +20,7 @@ AutoCreateKey = true # Default
 BlockBackfillDepth = 10 # Default
 BlockBackfillSkip = false # Default
 ChainType = 'arbitrum' # Example
+SafeBlockDepth = 0 # Default
 FinalityDepth = 50 # Default
```
jmank88: Would omitting `Block` be more consistent with Finality?
```suggestion
SafeDepth = 0 # Default
FinalityDepth = 50 # Default
```

prashantkumar1982: Yes, that's a good suggestion!

jmank88 (approved):

dhaidashenko (approved):

PR #110, opened 2025-06-16
Use different allowed signer to unlink

ibrajer:


ibrajer:
This PR was just a showcase, closing it now

PR #111, opened 2025-06-17
PLEX-250 - WriteReport initial implementation

pablolagreca:
[PLEX-250](https://smartcontract-it.atlassian.net/browse/PLEX-250)

### Requires
https://github.com/smartcontractkit/chainlink-common/pull/1264
https://github.com/smartcontractkit/chainlink-framework/pull/50

### Supports
https://github.com/smartcontractkit/chainlink/pull/18170
https://github.com/smartcontractkit/capabilities/pull/122

[PLEX-250]: https://smartcontract-it.atlassian.net/browse/PLEX-250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/chains/legacyevm/evm_txm.go
```diff
@@ -80,6 +86,16 @@ func newEvmTxm(
 	return
 }
 
+func validateTxWaitTimeForSubmitTransaction(cfg evmconfig.EVM) error {
+	if cfg.TxMinimumWaitTimeForConfirmation().Seconds() > time.Duration(600).Seconds() {
```
silaslenihan: Can we make `time.Duration(600).Seconds()` a constant

pkg/config/toml/defaults/Ethereum_Mainnet.toml
```diff
@@ -4,6 +4,8 @@ LinkContractAddress = '0x514910771AF9Ca656af840dff83E8264EcF986CA'
 MinContractPayment = '0.1 link'
 OperatorFactoryAddress = '0x3E64Cd889482443324F91bFA9c84fE72A511f48A'
 NoNewFinalizedHeadsThreshold = '9m'
+TxMaximumWaitTimeForConfirmation = '12s'
```
silaslenihan: I think these values are reversed, or maybe Max is supposed to be 12m

pkg/config/toml/docs.toml
```diff
@@ -118,6 +118,10 @@ LogBroadcasterEnabled = true # Default
 #
 # Set to zero to disable.
 NoNewFinalizedHeadsThreshold = '0' # Default
+# TxMinimumWaitTimeForConfirmation minimum time to wait for a TX to get into a block in the blockchain. This is used for the EVMService.SubmitTransaction operation. It won't check for TX status updates until this time passes
```
silaslenihan: Why do we even need a minimum?

pkg/config/toml/defaults/Ethereum_Mainnet.toml
pablolagreca: You are right. I'll make the minimum time 12 seconds which is what takes to create a block and give 20 seconds to be safe that we have enough time before failing.

pkg/config/toml/docs.toml
pablolagreca: We need a minimum because if a block takes 12 seconds to be created it doesn't make sense to be checking if it's confirmed before that time.

pkg/chains/legacyevm/evm_txm.go
pablolagreca: Makes sense.

pkg/config/toml/docs.toml
```diff
@@ -118,6 +118,10 @@ LogBroadcasterEnabled = true # Default
...
+TxMinimumWaitTimeForConfirmation = '60s' # Default
+# TxMaximumWaitTimeForConfirmation time after TxMinimumWaitTimeForConfirmation to wait for a TX to get into a block. After TxMinimumWaitTimeForConfirmation the operation will check every 100 milliseconds to see if the TX gets into a block and will do it for no longer than TxMaximumWaitTimeForConfirmation. Total wait time will be TxMinimumWaitTimeForConfirmation + TxMaximumWaitTimeForConfirmation
+TxMaximumWaitTimeForConfirmation = '300s' # Default
```
jmank88: We have an `[EVM.Transactions]` section, why don't we move these there? We also try to name things so they group together naturally when sorted
```suggestion
# ConfirmationWaitTimeMin minimum time to wait for a TX to get into a block in the blockchain. This is used for the EVMService.SubmitTransaction operation. It won't check for TX status updates until this time passes
ConfirmationWaitTimeMin = '60s' # Default
# ConfirmationWaitTimeMax time after TxMinimumWaitTimeForConfirmation to wait for a TX to get into a block. After TxMinimumWaitTimeForConfirmation the operation will check every 100 milliseconds to see if the TX gets into a block and will do it for no longer than TxMaximumWaitTimeForConfirmation. Total wait time will be TxMinimumWaitTimeForConfirmation + TxMaximumWaitTimeForConfirmation
ConfirmationWaitTimeMax = '300s' # Default
```

pkg/config/toml/docs.toml
jmank88: Are there any existing fields that we could derive this from dynamically instead of having to calibrate it for each chain?

jmank88: Although, FWIW the time between submission and inclusion can be arbitrarily small no matter how long the block period is - it just might happen to show up right before the next block goes out. What is the rationale for `60s`?

pablolagreca: Open to help here!. What do you guys suggest?. Maybe keep TxMinimumWaitTimeForConfirmation very low?. Something like 5 seconds? Less?. And keep TxMaximumWaitTimeForConfirmation to 300 seconds?

pkg/config/toml/defaults/Ethereum_Mainnet.toml
pablolagreca: Fixed.

pkg/config/toml/docs.toml
silaslenihan: I think we don't need the minimum time since as Jordan said the inclusion time could be arbitrarily small. I haven't seen this inclusion minimum poll time pattern used anywhere else

pablolagreca: Then I'll drop the TxMinimumWaitTimeForConfirmation and just keep the TxMaximumWaitTimeForConfirmation which I'll rename to ConfirmationWaitTime per @jmank88  suggestion.  Then I can set a default value of 60 seconds which will work well for most chains and use 5 minutes in EVM which is the maximum it seems it may take. Makes sense?

jmank88: I think that makes sense, although would naming the field `ConfirmationTimeout` make sense in that case?

pkg/chains/legacyevm/evm_txm.go
pablolagreca: Done

pkg/client/evm_client_test.go
```diff
@@ -39,6 +39,8 @@ func TestNewEvmClient(t *testing.T) {
 	noNewFinalizedBlocksThreshold := time.Second * 5
 	finalizedBlockPollInterval := time.Second * 4
 	newHeadsPollInterval := time.Second * 4
+	txMinimumWaitTimeForConfirmation := time.Second * 60
```
silaslenihan: I thought we removed this minimum?

pablolagreca: Yes, forgot to change that line but I already did.

silaslenihan (dismissed):

jmank88 (dismissed):

pkg/config/toml/defaults/Ethereum_Mainnet.toml
```diff
@@ -5,6 +5,9 @@ MinContractPayment = '0.1 link'
 OperatorFactoryAddress = '0x3E64Cd889482443324F91bFA9c84fE72A511f48A'
 NoNewFinalizedHeadsThreshold = '9m'
 
+[Transactions]
```
dhaidashenko, line 8: Should we keep this config in dedicated structure to signal that it's only related to the EVM Chain Capability and not to the TXM in general?

dhaidashenko (commented):
it would be nice to move config to a dedicated section

pkg/config/toml/defaults/Ethereum_Mainnet.toml
pablolagreca: All this config is specific to EVM. It doesn't need the EVM prefix.

dhaidashenko: [EVM.Transactions] defines the config for TXM. `ConfirmationTimeout` is a config option for EVM Chain Capability. We should probably move it to [EVM.ChainCapability] to make it clearer.

dhaidashenko (approved):

jmank88 (approved):

jmank88 (approved):

PR #114, opened 2025-06-18
Refactor dual broadcast clients

dimriou:


dimriou:
Closed by accident

pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/smartcontractkit/chainlink-evm/pkg/client"
+	"github.com/smartcontractkit/chainlink-evm/pkg/keys"
+	"github.com/smartcontractkit/chainlink-evm/pkg/txm"
+)
+
+func SelectClient(lggr logger.Logger, client client.Client, keyStore keys.ChainStore, url *url.URL, chainID *big.Int) txm.Client {
+	urlString := url.String()
+	switch {
+	case strings.Contains(urlString, "flashbots"):
```
jmank88, line 18: Is is possible to make this more specific to avoid incidental false positives? Like by comparing to the url fields, or including more characters at the start & end?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,206 @@
...
+const (
+	timeout = time.Second * 300
+)
+
+type MetaClientKeystore interface {
+	SignMessage(ctx context.Context, address common.Address, data []byte) ([]byte, error)
+	SignTx(ctx context.Context, fromAddress common.Address, tx *evmtypes.Transaction) (*evmtypes.Transaction, error)
+}
+
+type MetaClient struct {
```
jmank88, line 129: What does `Meta` mean in this context? We are already in package `dualbroadcast`, can this just be a `dualbroadcast.Client`?

jmank88: Er, maybe just some docs would help

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -20,25 +20,25 @@ import (
 	"github.com/smartcontractkit/chainlink-evm/pkg/txm/types"
 )
 
-type DualBroadcastClient struct {
+type FlashbotsClient struct {
 	c         client.Client
 	keystore  keys.MessageSigner
 	customURL *url.URL
```
cll-gg, line 29: Since we're refactoring this anyway, shall we rename `customURL` to `url`, since it's not custom in the scope of Flashbots?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,206 @@
...
+	lggr      logger.SugaredLogger
+	c         client.Client
+	ks        MetaClientKeystore
+	customURL *url.URL
```
cll-gg, line 133:
```suggestion
	url *url.URL
```
or 
```suggestion
	endpoint *url.URL
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send POST request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
cll-gg, line 286: Do we want to allow also other 2xx status codes, e.g. 202?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,206 @@
...
+	ChainID      *hexutil.Uint64 `json:"chainId"`
+	ToAddress    common.Address  `json:"adapter"`
+	Payload      hexutil.Bytes   `json:"updatePayload"`
+	ER           bool            `json:"earlyReturn"`
+	FromAddress  common.Address  `json:"bundlerEoa"`
+	MaxFeePerGas *hexutil.Big    `json:"maxFeePerGas"`
+	Signature    hexutil.Bytes   `json:"signature"`
+}
+
+type requestResponse struct {
```
cll-gg: Slightly more descriptive:
```suggestion
type metaResponse struct {
```

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: The name needs to stay consistent with the original Node config, and CustomURL seems good enough to me to avoid confusion with the standard RPC URLs that we already use in the Node.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Unless we can confirm that the API actually uses/returns these values, I don't see the benefit. We should try to be explicit.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: In the scope of FlashbotsClient it seems unexpected to have a `customURL`, but fine with me

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Fair enough

cll-gg:
Please fill in the PR description a bit, helpful for traceability

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -64,7 +64,7 @@ func (d *DualBroadcastClient) SendTransaction(ctx context.Context, tx *types.Tra
 			return err
 		}
 		params := ""
-		if meta.DualBroadcastParams != nil {
+		if meta.DualBroadcast != nil {
```
cll-gg: Is this change intended? Seems unexpected to do a `nil` check for `DualBroadcast` but then in the if statement body use `DualBroadcastParams`

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -20,25 +20,25 @@ import (
 	"github.com/smartcontractkit/chainlink-evm/pkg/txm/types"
 )
```
cll-gg, line 23: Might be good to add a type assertion, e.g.:

```suggestion
var _ txm.Client = (*FlashbotsClient)(nil)
```

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -20,25 +20,25 @@ import (
...
 
-type DualBroadcastClient struct {
+type FlashbotsClient struct {
```
cll-gg, line 26: While we're renaming, can we make it unexported to limit the API surface?
```suggestion
type flashbotsClient struct {
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,614 @@
...
+  }
+]`
+)
+
+type MetaClientKeystore interface {
+	SignMessage(ctx context.Context, address common.Address, data []byte) ([]byte, error)
+	SignTx(ctx context.Context, fromAddress common.Address, tx *evmtypes.Transaction) (*evmtypes.Transaction, error)
+}
+
+type MetaClient struct {
```
cll-gg, line 129: Unexport?
```suggestion
type metaClient struct {
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,614 @@
...
+    "type": "function"
+  }
+]`
+)
+
+type MetaClientKeystore interface {
+	SignMessage(ctx context.Context, address common.Address, data []byte) ([]byte, error)
+	SignTx(ctx context.Context, fromAddress common.Address, tx *evmtypes.Transaction) (*evmtypes.Transaction, error)
+}
+
```
cll-gg, line 122: Nice to add type assertion:
```suggestion
var _ txm.Client = (*MetaClient)(nil)
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,614 @@
...
+	return a.c.PendingNonceAt(ctx, address)
+}
+
+func (a *MetaClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	meta, err := tx.GetMeta()
+	if err != nil {
+		return err
+	}
+
+	if meta != nil && meta.DualBroadcast != nil && *meta.DualBroadcast && !tx.IsPurgeable && meta.DualBroadcastParams != nil && meta.FwdrDestAddress != nil {
```
cll-gg, line 161: Shall we log something in case this `if` doesn't hold?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -0,0 +1,614 @@
...
+
+	return &metacalldata, nil
+}
+
+func (a *MetaClient) SendOperation(ctx context.Context, tx *types.Transaction, attempt *types.Attempt, meta MetacalldataResponse) error {
+	if tx.Nonce == nil {
+		return fmt.Errorf("failed to create attempt for txID: %v: nonce empty", tx.ID)
+	}
+
+	// TODO: fastest way to avoid overpaying, but might require additional checks.
```
cll-gg, line 438: To do for now or for later?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg, line 1: A unit test would be good. Maybe an LLM can generate one at least for the business logic in `SendTransaction()` where we're doing 3 calls in a row.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Exporting it makes our lives easier during testing because we can import it locally and run our tests outside the Core node.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: No, nice catch.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Same reasoning as the other client.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Makes sense 👍

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Core has been notorious for excessive logging. We can infer the information from the other logs, more importantly, from TXM.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Not required atm, it's for later.

jmank88 (dismissed):

silaslenihan (dismissed):

amit-momin (dismissed):

Farber98 (dismissed):

ilija42 (dismissed):

DylanTinianov (approved):

silaslenihan (approved):

PR #115, opened 2025-06-18
EVM Bindings Generator

silaslenihan:
This PR implements the [EVM Bindings Generator Design](https://docs.google.com/document/d/1AZaJFNM8vhlEYqKYkuFhv_e-xCLcq3oFt22qoaPyKdg/edit?tab=t.0).

It includes:
- A forked [abigen](https://github.com/smartcontractkit/chainlink-evm/tree/PLEX-1500-bindings-generator/pkg/bindings/abigen) (removed tests) to allow custom templates.
- The template to [generate bindings](https://github.com/smartcontractkit/chainlink-evm/blob/PLEX-1500-bindings-generator/pkg/bindings/sourcecre.go.tpl).
- [Binding generation function](https://github.com/smartcontractkit/chainlink-evm/blob/PLEX-1500-bindings-generator/pkg/bindings/bindgen.go).
- [testdata](https://github.com/smartcontractkit/chainlink-evm/tree/PLEX-1500-bindings-generator/pkg/bindings/testdata) including a [sample contract ](https://github.com/smartcontractkit/chainlink-evm/blob/PLEX-1500-bindings-generator/pkg/bindings/testdata/DataStorage.sol) and associated [generated bindings file](https://github.com/smartcontractkit/chainlink-evm/blob/PLEX-1500-bindings-generator/pkg/bindings/testdata/bindings.go).
- [Unit tests](https://github.com/smartcontractkit/chainlink-evm/blob/PLEX-1500-bindings-generator/pkg/bindings/bindings_test.go) that generate and test the bindings for the sample DataStorage contract.


pkg/bindings/common.go
```diff
@@ -0,0 +1,99 @@
+package bindings
+
+import (
+	"fmt"
+	"math/big"
+
+	evmcappb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/v2/chain-capabilities/evm"
+	chain_common "github.com/smartcontractkit/chainlink-common/pkg/loop/chain-common"
+	evmtypes "github.com/smartcontractkit/chainlink-common/pkg/types/chains/evm"
```
pablolagreca: I don't know if I would make the generated code to depend chainlink-common. If we do then users will be sort of aware of chainlink-common and things like versioning of chainlink-common, etc. Can we avoid it?

pkg/bindings/common.go
```diff
@@ -0,0 +1,99 @@
...
+
+const FINALIZED = 0
+const CONFIRMED = 1
+
+type LogTrackingOptions struct {
+	MaxLogsKept   uint64 `protobuf:"varint,1,opt,name=max_logs_kept,json=maxLogsKept,proto3" json:"max_logs_kept,omitempty"`     // maximum number of logs to retain ( 0 = unlimited )
+	RetentionTime int64  `protobuf:"varint,2,opt,name=retention_time,json=retentionTime,proto3" json:"retention_time,omitempty"` // maximum amount of time to retain logs in seconds
+	LogsPerBlock  uint64 `protobuf:"varint,3,opt,name=logs_per_block,json=logsPerBlock,proto3" json:"logs_per_block,omitempty"`  // rate limit ( maximum # of logs per block, 0 = unlimited )
+	//TODO could this be actual values for the indexed log fields instead of hashes?
+	Topic2 []*evmtypes.Hash `protobuf:"bytes,7,rep,name=topic2,proto3" json:"topic2,omitempty"` // list of possible values for topic2
```
pablolagreca: This types should all come from the capabilities.proto generated code from EVM. Why can't you use those?. Is because they are proto buf related?

pkg/bindings/cre_common.go
```diff
@@ -0,0 +1,6 @@
+package bindings
+
+// This function is not EVM specific, it's generic and should be provided by CRE
```
pablolagreca: Can you create a ticket to track pending stuff like plugging in the actual function for generating a report once that's done?

pablolagreca (changes requested):

pkg/bindings/common.go
silaslenihan: Hm, I'm not sure how we can avoid it in this case since the [QueryTrackedLogsRequest](https://github.com/smartcontractkit/chainlink-common/blob/142aef37e6a562044f671a776430334559c90617/pkg/chains/evm/evm.pb.go#L2106-L2113) is dependent on these types.

pkg/bindings/cre_common.go
silaslenihan: https://smartcontract-it.atlassian.net/browse/PLEX-1555

pkg/bindings/testdata/bindings.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+// Structs
+type DataStorageUserData struct {
```
pablolagreca, line 45: This should be named UserData

pkg/bindings/testdata/bindings.go
```diff
@@ -0,0 +1,444 @@
...
+	if options == nil {
+		options = &bindings.ReadOptions{BlockNumber: nil}
+	}
+	return c.evmClient.CallContract(runtime, &evm.CallContractRequest{
+		Call:        &evm.CallMsg{To: c.Address, Data: calldata},
+		BlockNumber: pb.NewBigIntFromInt(options.BlockNumber),
+	}), nil
+}
+
+func (c DataStorage) WriteReportDataStorageUserData(
```
pablolagreca: WriteReportUserData. Maybe this should be WriteUserDataReport?. Maybe we can track this in the TODO ticket as an improvement in the UX.

pablolagreca (changes requested):

pkg/bindings/testdata/bindings.go
silaslenihan: This is tricky because [abigen generates the struct name including the contract name.](https://github.com/smartcontractkit/chainlink-evm/blob/703c00ead7b879c276ad78208015ead711850117/pkg/bindings/abigen/bind.go#L383-L387). I could add additional processing to the abigen fork that removes the contract name if present, but that becomes another thing we need to maintain in the fork.

pkg/bindings/testdata/bindings.go
silaslenihan: Same issue as [above](https://github.com/smartcontractkit/chainlink-evm/pull/115#discussion_r2164084110)

silaslenihan: added to the [TODO ticket](https://smartcontract-it.atlassian.net/browse/PLEX-1555)

pkg/bindings/README.md
```diff
@@ -0,0 +1,42 @@
...
+2. **Solidity compiler**
+   Install `solc` to compile or verify your contracts:
+   ```bash
+   npm install -g solc      # via npm
+   brew install solidity    # macOS (Homebrew)
+   ```
+
+## Usage
+
+You can generate your Go contract bindings in two ways:
```
ilija42: nit, says two ways, but only lists one

pkg/bindings/abigen/bind.go
```diff
@@ -0,0 +1,457 @@
+// Copyright 2016 The go-ethereum Authors
+// This file is part of the go-ethereum library.
```
ilija42, line 2: I think that this warrants a mention under a LGPL license in the LICENSE readme, please double check

pkg/bindings/common.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Logs support
+const FINALIZED = 0
+const CONFIRMED = 1
+
+type LogTrackingOptions struct {
+	MaxLogsKept   uint64 `protobuf:"varint,1,opt,name=max_logs_kept,json=maxLogsKept,proto3" json:"max_logs_kept,omitempty"`     // maximum number of logs to retain ( 0 = unlimited )
+	RetentionTime int64  `protobuf:"varint,2,opt,name=retention_time,json=retentionTime,proto3" json:"retention_time,omitempty"` // maximum amount of time to retain logs in seconds
+	LogsPerBlock  uint64 `protobuf:"varint,3,opt,name=logs_per_block,json=logsPerBlock,proto3" json:"logs_per_block,omitempty"`  // rate limit ( maximum # of logs per block, 0 = unlimited )
+	// TODO could this be actual values for the indexed log fields instead of hashes?
```
ilija42: Left a TODO here. wdym by actual values

pkg/bindings/common.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type FilterOptions struct {
+	BlockHash []byte
+	FromBlock *big.Int
+	ToBlock   *big.Int
+}
+
+func ValidateLogTrackingOptions(opts *LogTrackingOptions) {
+	// TODO: set defaults for opts
```
ilija42: left a TOODO here

pkg/bindings/common.go
```diff
@@ -0,0 +1,109 @@
...
+	BlockNumber *big.Int
+}
+
+type WriteOptions struct {
+	GasConfig  *evmcappb.GasConfig
+	BlockDepth uint16 // 0 means finalized, 1 confirmed, positive numbers block depth - TODO to be defined together with all other operations
+}
+
+// Logs support
+const FINALIZED = 0
```
ilija42: can we import these from common?

pkg/bindings/common.go
```diff
@@ -0,0 +1,109 @@
...
+	Confidence chain_common.Confidence
+	BlockDepth uint64
+}
+
+type ParsedLog[T any] struct {
+	LogData T
+	RawLog  *evm.Log
+}
+
+type FilterOptions struct {
```
ilija42, line 36: why not just use FilterLogsRequest proto in the binding?

pkg/bindings/common.go
```diff
@@ -0,0 +1,109 @@
+package bindings
```
ilija42, line 1: for a lot of these I'm not sure why not just use the underlying proto types that the capability uses anyway

pkg/bindings/sourcecre.go.tpl
```diff
@@ -0,0 +1,367 @@
...
+func (c *{{$contract.Type}}) UnregisterLogTracking{{.Normalized.Name}}(runtime sdk.Runtime) {
+	c.evmClient.UnregisterLogTracking(runtime, &evm.UnregisterLogTrackingRequest{
+		FilterName: "{{.Normalized.Name}}-" + common.Bytes2Hex(c.Address),
+	})
+}
+
+func (c *{{$contract.Type}}) FilterLogs{{.Normalized.Name}}(runtime sdk.Runtime, options *bindings.FilterOptions) (sdk.Promise[*evm.FilterLogsReply]) {
+	if options == nil {
+		options = &bindings.FilterOptions{
+			ToBlock: big.NewInt(123), //TODO we need a enum / constant
```
ilija42: why is this the default?

pkg/bindings/sourcecre.go.tpl
```diff
@@ -0,0 +1,367 @@
...
+
+{{range $contract := .Contracts}}
+var {{$contract.Type}}MetaData = &bind.MetaData{
+	ABI: "{{.InputABI}}",
+	{{- if .InputBin}}
+	Bin: "0x{{.InputBin}}",
+	{{- end}}
+}
+
+// TODO: remove contract name from structs
```
ilija42: left a todo here

ilija42 (commented):
Looks good, left some very small nits. Mostly curios why do we need most of the common file instead of using the generated protos and evm types from common

pkg/bindings/abigen/bind.go
silaslenihan: I added licensing in [this commit](https://github.com/smartcontractkit/chainlink-evm/pull/115/commits/f7cc8d32672c42ea758e376628e2b2bcacdfa923) (ChatGPT assisted). How does that look to you?

pkg/bindings/common.go
silaslenihan: going to remove the TODO and address in a later ticket: https://smartcontract-it.atlassian.net/browse/PLEX-1555

pkg/bindings/sourcecre.go.tpl
silaslenihan: made the default `options.ToBlock`, in the future we want to have a enum/constant for block depth but I will leave it as a TODO in this ticket for now: https://smartcontract-it.atlassian.net/browse/PLEX-1555

pkg/bindings/common.go
silaslenihan: The reason why for this one specifically is that we don't want the bindings user to be able to control the Address and Topics, we need them to use the [default options](https://github.com/smartcontractkit/chainlink-evm/blob/fb7b66467ac1fafceee6ed2e79b9aa32f3eb0dd6/pkg/bindings/testdata/bindings.go#L390-L393). Using the generate proto types might lead to confusion because they'd have access to those fields.

pkg/bindings/common.go
silaslenihan: cleaned this up quite a bit, the remaining ones echo the [reasoning above ](https://github.com/smartcontractkit/chainlink-evm/pull/115#discussion_r2183387783), we don't want the user to be able to control certain fields, and including them in the options would complicate the UX

pablolagreca (approved):

ilija42 (approved):

PR #122, opened 2025-06-23
Use Fixed Energy Value for TRON

HelloKashif:
https://smartcontract-it.atlassian.net/browse/INTAUTO-707

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/smartcontractkit/chainlink-framework/metrics v0.0.0-20250618135814-7e3f79ab707e
 	github.com/smartcontractkit/chainlink-framework/multinode v0.0.0-20250522110034-65c54665034a
 	github.com/smartcontractkit/chainlink-protos/svr v1.1.0
-	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250528121202-292529af39df
+	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250701052953-acaa2d8bb76e
 	github.com/smartcontractkit/libocr v0.0.0-20250328171017-609ec10a5510
 	github.com/stretchr/testify v1.10.0
 	github.com/theodesp/go-heaps v0.0.0-20190520121037-88e35354fe0a
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/smartcontractkit/chainlink-framework/metrics v0.0.0-20250618135814-7e3f79ab707e
 	github.com/smartcontractkit/chainlink-framework/multinode v0.0.0-20250522110034-65c54665034a
 	github.com/smartcontractkit/chainlink-protos/svr v1.1.0
-	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250701052953-acaa2d8bb76e
+	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250701132001-f8be142155b6
 	github.com/smartcontractkit/libocr v0.0.0-20250328171017-609ec10a5510
 	github.com/stretchr/testify v1.10.0
 	github.com/theodesp/go-heaps v0.0.0-20190520121037-88e35354fe0a
--- a/pkg/chains/tron/txm.go
+++ b/pkg/chains/tron/txm.go
@@ -23,10 +23,8 @@ func ConstructTxm(logger logger.Logger, nodes []*toml.Node, keystore keys.Store)
 	}
 
 	return trontxm.New(logger, tronkeystore.NewLoopKeystoreAdapter(keystore), tronClient, trontxm.TronTxmConfig{
-		// From testing, this multiplier ensures all exec messages are fully executed.
-		// Energy estimation doesn't seem to account for more complex smart contract execution.
-		// Given that Tron has static gas prices, we don't expect this to be a problem as this multiplier is sufficiently high.
-		EnergyMultiplier: 3,
+		// Overrides the energy estimator to always use the fixed energy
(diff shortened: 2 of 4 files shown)


pkg/chains/tron/txm.go
```diff
@@ -26,7 +27,7 @@ func ConstructTxm(logger logger.Logger, nodes []*toml.Node, keystore keys.Store)
 		// From testing, this multiplier ensures all exec messages are fully executed.
 		// Energy estimation doesn't seem to account for more complex smart contract execution.
 		// Given that Tron has static gas prices, we don't expect this to be a problem as this multiplier is sufficiently high.
-		EnergyMultiplier: 3,
+		EnergyMultiplier: float64(cfg.EVM().GasEstimator().LimitMultiplier()),
```
KodeyThomas: do we have any nil checks? for tron this would be required might be worth adding defaults as well

KodeyThomas (changes requested):

pkg/chains/tron/txm.go
HelloKashif: @KodeyThomas can you elaborate please? I think the `cfg.EVM().GasEstimator().LimitMultiplier()` always 
returns a hard float32 so it won't be nil. Looks like the cfg.EVM() automatically initializes all the pointers

Are you referring to GasEstimator() being nil?

jmank88: Yeah the tests should have made you fill in a default. The underlying types just use pointers for marshaling to be able to tell what was zero v.s. unset by the user. Defaults fill in the rest.

pkg/chains/tron/txm.go
```diff
@@ -4,14 +4,15 @@ import (
...
+	"github.com/smartcontractkit/chainlink-evm/pkg/config"
 	"github.com/smartcontractkit/chainlink-evm/pkg/config/toml"
 	"github.com/smartcontractkit/chainlink-evm/pkg/keys"
 	tronkeystore "github.com/smartcontractkit/chainlink-tron/relayer/keystore"
 	tronclient "github.com/smartcontractkit/chainlink-tron/relayer/sdk"
 	trontxm "github.com/smartcontractkit/chainlink-tron/relayer/txm"
 )
 
-func ConstructTxm(logger logger.Logger, nodes []*toml.Node, keystore keys.Store) (*trontxm.TronTxm, error) {
+func ConstructTxm(logger logger.Logger, cfg *config.ChainScoped, nodes []*toml.Node, keystore keys.Store) (*trontxm.TronTxm, error) {
```
jmank88: Let's pass a more specific interface, or even just `limitMultiplier` since that is the only field used currently.

jmank88: I think we usually do this:
```suggestion
type TxmConfig interface {
	LimitMultiplier() float64
}

func ConstructTxm(logger logger.Logger, cfg Config, nodes []*toml.Node, keystore keys.Store) (*trontxm.TronTxm, error) {
```
Then the caller can pass `cfg.EVM().GasEstimator()`.

HelloKashif: I was wondering if we might use any other config in the future. @KodeyThomas  if we don't need anything else I can just pass limitMultiplier. Please confirm?

HelloKashif: For now I've updated it to just pass the energyMultiplier directly

KodeyThomas (changes requested):
After discussions we need to pass a fixed gas limit into the tron txm due to quirks with CCIP 1.5

go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	github.com/smartcontractkit/chainlink-framework/metrics v0.0.0-20250618135814-7e3f79ab707e
 	github.com/smartcontractkit/chainlink-framework/multinode v0.0.0-20250522110034-65c54665034a
 	github.com/smartcontractkit/chainlink-protos/svr v1.1.0
-	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250528121202-292529af39df
+	github.com/smartcontractkit/chainlink-tron/relayer v0.0.11-0.20250701052953-acaa2d8bb76e
```
KodeyThomas: this will need to be a commit from main, might need to bump after the chainlink-tron bits are merged

KodeyThomas (changes requested):

go.mod
HelloKashif: Yep, bumped the tests once more and now they are passing. Ptal

pkg/chains/tron/txm.go
```diff
@@ -23,10 +23,8 @@ func ConstructTxm(logger logger.Logger, nodes []*toml.Node, keystore keys.Store)
 	}
 
 	return trontxm.New(logger, tronkeystore.NewLoopKeystoreAdapter(keystore), tronClient, trontxm.TronTxmConfig{
-		// From testing, this multiplier ensures all exec messages are fully executed.
-		// Energy estimation doesn't seem to account for more complex smart contract execution.
-		// Given that Tron has static gas prices, we don't expect this to be a problem as this multiplier is sufficiently high.
-		EnergyMultiplier: 3,
+		// Overrides the energy estimator to always use the fixed energy
+		FixedEnergyValue: 6_000_000,
```
KodeyThomas: thought we said we'd make this configurable?

KodeyThomas (changes requested):
think we should make the value configurable

pkg/chains/tron/txm.go
HelloKashif: oh sorry missed that. Fixed!

KodeyThomas (approved):

athegaul (approved):

DylanTinianov (approved):

george-dorin (approved):

PR #125, opened 2025-06-27
Inline internal functions for enhanced readability

DeividasK:
- Inline internal functions for readability improvement (a non-functional change).
- Simplify Event Record payload, keeping just the necessary data.


eutopian (approved):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -551,7 +539,23 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
       revert CannotUnlinkWithActiveWorkflows();
```
timothyF95, line 539: 👍

timothyF95 (approved):

PR #127, opened 2025-06-30
PLEX-1499: support 'safe' tag for log poller queries

fernandezlautaro:
Jira: https://smartcontract-it.atlassian.net/browse/PLEX-1499

 

### Requires
https://github.com/smartcontractkit/chainlink/pull/18394
https://github.com/smartcontractkit/chainlink-common/pull/1309


CONFIG.md
```diff
@@ -803,7 +803,7 @@ the prices and end up in stale values.
 HistoryDepth = 100 # Default
 MaxBufferSize = 3 # Default
 SamplingInterval = '1s' # Default
-FinalityTagBypass = true # Default
+FinalityTagBypass = false # Default
```
dhaidashenko, line 806: Why do you need this change?

pkg/heads/headstest/simulated_head_tracker.go
```diff
@@ -52,6 +52,14 @@ func (ht *simulatedHeadTracker) LatestAndFinalizedBlock(ctx context.Context) (*e
 	return latest, finalizedBlock, nil
 }
 
+func (ht *simulatedHeadTracker) LatestSafeBlock(ctx context.Context) (safe *evmtypes.Head, err error) {
+	block, err := ht.ec.LatestSafeBlock(ctx)
```
dhaidashenko: I'm not sure that simulated RPC supports finality tags (as result safe tags). We should probably return finalized block from `LatestAndFinalizedBlock` here. Or, if needed in tests pass safetyDepth and calculate using it.

pkg/logpoller/log_poller.go
```diff
@@ -1055,6 +1056,18 @@ func (lp *logPoller) pollAndSaveLogs(ctx context.Context, currentBlockNumber int
...
 		return nil
 	}
+
+	_, safeBlockNumber, err := lp.latestSafeBlocks(ctx)
+	if err != nil {
+		lp.lggr.Warnw("Unable to get latest safe block", "err", err, "currentBlockNumber", currentBlockNumber)
+		return nil
+	}
+	// if safe block is lower than finalized block, it doesn't make sense to use it
+	if safeBlockNumber < latestFinalizedBlockNumber {
```
dhaidashenko: nit: should we move it into `latestSafeBlocks` function?

pkg/logpoller/log_poller_internal_test.go
```diff
@@ -213,6 +213,7 @@ func TestLogPoller_BackupPollerStartup(t *testing.T) {
 	const finalityDepth = 2
 
 	head := &evmtypes.Head{Number: latestBlock}
+	safeHead := &evmtypes.Head{Number: latestBlock - finalityDepth - 1} //forcing safe head to be lower than finalized head
```
dhaidashenko: nit: Should not we test here happy path, where safe block stored by LP is between latest and finalized?

pkg/logpoller/models.go
```diff
@@ -19,6 +19,7 @@ type Block struct {
 	BlockNumber          int64
 	BlockTimestamp       time.Time
 	FinalizedBlockNumber int64
+	SafeBlockNumber      int64
```
dhaidashenko, line 22: If we won't add `default 0` in the sql migration, we need to use `sql.NullInt64` here to support null values

pkg/logpoller/parser.go
```diff
@@ -72,6 +72,9 @@ func (v *pgDSLParser) Confidence(p primitives.Confidence) {
 		v.expression = v.nestedConfQuery(true, 0)
 	case primitives.Unconfirmed:
 		v.expression = v.nestedConfQuery(false, 0)
+	case primitives.Safe:
+		v.expression = v.nestedConfQuery(false, 0)
```
dhaidashenko: You need to modify `nestedConfQuery` to handle the safe tag. Right now, you'll also return logs that were included into latest head.

CONFIG.md
fernandezlautaro: I ran `go test ./...` from root folder to see if everything was working and I got this error:
```bash
                  Diff:
                  --- Expected
                  +++ Actual
                  @@ -805,3 +805,3 @@
                   SamplingInterval = '1s' # Default
                  -FinalityTagBypass = true # Default
                  +FinalityTagBypass = false # Default
                   MaxAllowedFinalityDepth = 10000 # Default
  Test:           TestConfigDocs
  Messages:       CONFIG.md is out of date. Run 'go generate .' to regenerate.
FAIL
```

so I ran `go generate .` which modify this file with it, I can revert/drop it, should I?

pkg/heads/headstest/simulated_head_tracker.go
fernandezlautaro: returned finalized block as requested. 
I also notice this method is not being used anywhere, but I still need it to fulfill's [Client](https://github.com/smartcontractkit/chainlink-evm/blob/e92bf0b88371c48415fefede2ecee63e63edb140/pkg/client/chain_client.go#L63) interface, am I'm missing code using this method then?

pkg/logpoller/log_poller.go
fernandezlautaro: I could, but that implies performing another call to the headTracker to get latest and finalized block inside the function `latestSafeBlocks()`
Should I move it either way and perform the other call?

pkg/logpoller/log_poller_internal_test.go
fernandezlautaro: totally agree, missed it entirely. 
Also, while creating it it threw a bug I had in the ORM layer where I was inserting a wrong value, so thanks for pointing this up.

pkg/logpoller/models.go
fernandezlautaro: ty!, updated it to be consistent with the DML `alter` table thing, but it might get rollbacked once mateusz answer [here (pr for /chainlink)](https://github.com/smartcontractkit/chainlink/pull/18394#discussion_r2175654145)

CONFIG.md
prashantkumar1982: @Unheilbar recently deprecated FinalityTagBypass.
https://github.com/smartcontractkit/chainlink-evm/pull/86

Did you merge latest develop on your branch?

pkg/logpoller/log_poller.go
prashantkumar1982: Agree with @dhaidashenko, it makes this code cleaner.
You don't need to make another headtracker call.
You already fetched the finalized block in the variable latestFinalizedBlockNumber. Just pass it as a param to latestSafeBlocks().

pkg/logpoller/models.go
```diff
@@ -19,6 +20,7 @@ type Block struct {
 	BlockNumber          int64
 	BlockTimestamp       time.Time
 	FinalizedBlockNumber int64
+	SafeBlockNumber      sql.NullInt64
```
prashantkumar1982: My preference would be to keep this as int64, and fill values with 0 during db migration.
But not that strong opinion, so do what's more convenient.

pkg/logpoller/parser.go
prashantkumar1982: +1

pkg/logpoller/log_poller.go
fernandezlautaro: you both are right, not sure why I thought this was part of an API that couldn't be changed.
Fixed it.

pkg/logpoller/models.go
fernandezlautaro: After Prashant comment, I left the `SafeBlockNumber` to be int64 for the time being

pkg/logpoller/models.go
fernandezlautaro: I went with that approach, will be int64 and during migration the `alter` will put 0 as default

CONFIG.md
fernandezlautaro: I just did, and it seems this change is required.

Should I rollback it and see if CI passes?

dhaidashenko (dismissed):

dhaidashenko (dismissed):

pkg/logpoller/log_poller.go
```diff
@@ -1157,6 +1166,24 @@ func (lp *logPoller) latestBlocks(ctx context.Context) (*evmtypes.Head, int64, e
 	return latest, finalizedBN, nil
 }
 
+// Returns information about safe latestSafeBlocks, LatestSafeBlock provided by HeadTracker
+func (lp *logPoller) latestSafeBlocks(ctx context.Context, latestFinalizedBlockNumber int64) (int64, error) {
```
ilija42: why is it blocks and not `latestSafeBlock`, `latestBlocks` returns a linked list of blocks hence the name, here we only return latestSafeBlock

fernandezlautaro: ty, fixed.

ilija42 (approved):

pablolagreca (approved):

PR #136, opened 2025-07-08
LogTrigger bindings

silaslenihan:
This reverts commit 0a01fbb16fd070a439149f4d89b203ae9a0247c0.


pablolagreca (dismissed):

pkg/bindings/sourcecre.go.tpl
```diff
@@ -320,6 +364,20 @@ func (e *{{.Normalized.Name}}) Error() string {
 
 {{range $event := $contract.Events}}
 
+func (c *{{$contract.Type}}) LogTrigger{{.Normalized.Name}}Log(confidence evm.ConfidenceLevel, values []{{.Normalized.Name}}) (sdk.Trigger[*evm.Log, *evm.Log], error) {
```
pablolagreca: Wouldn't make more sense to call the second param filters instead of values?

pablolagreca (changes requested):

pkg/bindings/sourcecre.go.tpl
silaslenihan: fixed

pablolagreca (approved):

fernandezlautaro (approved):

PR #138, opened 2025-07-08
PLEX-1571 Guarantee Observations are Safe/Finalized

dhaidashenko:
Added WithConfidenceMethods to guarantee that observations are finalized.
Required for support of load-balanced RPCs for read consensus. [Ticket](https://smartcontract-it.atlassian.net/browse/PLEX-1571).


pkg/client/rpc_client.go
```diff
@@ -1057,6 +1137,57 @@ func (r *RPCClient) BalanceAt(ctx context.Context, account common.Address, block
...
+// blockNumber - defines block at which call will be executed:
+//
+//	nil - execute at latest block of specified confidence level (e.g. latest finalized, latest safe, etc.);
+//	positive value - execute as specified height with confidence level;
+//
+// confidenceLevel - defines at which confidence level request will be executed:
+//
+//	Unconfirmed - returns the most recent data
+//	Finalized - returned data was finalized;
+//	Safe - returned data is highly unlikely to be reorged;
```
jmank88: I'm a little concerned about the complexity this is introducing, both for the user and the API surface. Do we really need to drift from the geth model, where the user passes these tags in place a block number? Why isn't that good enough? The only case not directly covered is the one where the caller passes a number & a level, but who is actually going to do that? If I want to use a block number, but wait for a confidence level, I would rather wait ahead of time, then make the call. Combining them like this seems unnecessary, and more complex. Instead of adding variants of multiple methods, what about a single method like `WaitForConfidence(ctx, blockNumber, confidence)` that can be called before any method that accepts a block number?

pkg/client/rpc_client.go
```diff
@@ -1360,3 +1532,92 @@ func (r *RPCClient) makeLogValid(log types.Log) (types.Log, error) {
...
+		r.logResult(lggr, err, duration, r.getRPCDomain(), request.Method+"WithConfidence",
+			"result", request.Result,
+		)
+	}()
+
+	referencedHead := new(evmtypes.Head)
+	// BatchElems are copied, so request and blockRequest values wont change, but requests[0] and requests[1] will
+	requests := []rpc.BatchElem{request, {
+		Method: "eth_getBlockByNumber",
+		Args:   []interface{}{ToBackwardCompatibleBlockNumArg(big.NewInt(referencedBlockNumber.Int64()))},
```
Unheilbar: Missed false argument

pkg/client/rpc_client.go
```diff
@@ -1360,3 +1532,92 @@ func (r *RPCClient) makeLogValid(log types.Log) (types.Log, error) {
...
+
+	if !r.finalityTagEnabled {
+		return rpc.LatestBlockNumber, nil
+	}
+
+	return referencedBlockNumber, nil
+}
+
+func (r *RPCClient) referenceHeadToMaxAvailableHeight(confidence primitives.ConfidenceLevel, referenceHeadHeight int64) (int64, error) {
+	if r.finalityTagEnabled {
```
Unheilbar, line 1522: what if some chains support Finalized but not Safe?

Unheilbar (changes requested):

pkg/client/rpc_client.go
dhaidashenko: > The only case not directly covered is the one where the caller passes a number & a level, but who is actually going to do that?

CRE chain capaiblity. CallContract and other calls are performed in two rounds: agree on the block height, agree on data at that height.  If user requested CallContract on latest finalized block, we need to guaranee that the data is acutally finalized. Since some honest nodes may be lagging behind on the finalized state, we need to double check that the data is acutally finalized. `WaitForConfidence` is not good for that purpose since it does not work on load balanced RPCs, which are used in production.

jmank88: > WaitForConfidence is not good for that purpose since it does not work on load balanced RPCs, which are used in production.

I would propose that it work in the same way, and at the same level. Any flaw present would also be present in this alternative version. :shrug: 

> agree on the block height, agree on data at that height.

Agreeing on height is an implicit agreement on block hash though. Querying by hash from that point avoids this issue.

dhaidashenko: > I would propose that it work in the same way, and at the same level. Any flaw present would also be present in this alternative version. 🤷

Could you elaborate? The problem with the `WaitForConfidence` followed by `CallContract` is that there is no guarantee that both calls will be executed by the same RPC. So there's a chance that we'll observe reorged data. BatchCall solves this problem

> Agreeing on height is an implicit agreement on block hash though. Querying by hash from that point avoids this issue.

Mapping height to hash is problematic. Example: Nodes agree that the latest finalized height is 10, Node A's observation was height 8, so it now needs to poll for block 10, guarantee that it's finalized, and be aware of which requests we can resume polling of actual data. It's doable, but it seems more complicated.
Also, we've previously observed cases where nodes were serving reorged data when requested by hash, despite having a new canonical block at that height.

pablolagreca (dismissed):

pkg/client/rpc_client.go
dhaidashenko: Nice catch. Fixed

pkg/client/rpc_client.go
dhaidashenko: Good point. In such cases, I believe we should return an error, as the user should not use the safe tag in this scenario.
If the chain does not support a safe tag (a polygon is the only one that we need to release), it returns an error or, in some cases, a null result. The returned error is self-explanatory. Added specific handling for instances where the returned block is null.

Unheilbar (dismissed):

pkg/client/rpc_client.go
```diff
@@ -658,6 +668,47 @@ func (r *RPCClient) BlockByNumber(ctx context.Context, number *big.Int) (head *e
...
+// Parameters:
+// blockNumber - specifies which block to fetch:
+//   - nil or -2: latest block
+//   - -3: finalized block
+//   - -4: safe block
+//   - positive value: specific block at that height
+//
+// confidence - determines if additional verification is required (only applicable for positive blockNumber values):
+//   - "Unconfirmed" or empty string: no additional verification
+//   - "Finalized": returns error if requested is not finalized
```
ilija42, line 686: nit
```suggestion
//   - "Finalized": returns error if requested block is not finalized
```

ilija42: nit
```suggestion
//   - "Finalized": returns error if requested block is not finalized
```

pkg/client/rpc_client.go
```diff
@@ -658,6 +668,47 @@ func (r *RPCClient) BlockByNumber(ctx context.Context, number *big.Int) (head *e
...
+//   - "Safe": returns error if requested block is not safe
+func (r *RPCClient) HeaderByNumberWithOpts(ctx context.Context, blockNumber *big.Int, opts evmtypes.HeaderByNumberOpts) (*evmtypes.Header, error) {
```
ilija42, line 688: Why opt for using an RPC call to get head data instead of Head Tracker, with head tracker we would have more confidence in the confidence tags being supported

pkg/client/rpc_client.go
```diff
@@ -977,6 +1028,36 @@ func (r *RPCClient) CallContract(ctx context.Context, msg interface{}, blockNumb
 	return
 }
 
+// CallContractWithOpts executes a message call transaction, which is directly executed in the VM of the node,
+// but never mined into the blockchain.
+//
+// blockNumber - defines block at which call will be executed:
```
ilija42, line 1034: The option of -1 to request a pending block number is not mentioned in the comment, but implicitly it is allowed

pkg/client/rpc_client.go
dhaidashenko: RPCClient provides us with the method to read a positive block with specified confidence. Relayer is the one that uses HeadTracker to fetch the latest and finalized block, or routes the request to the RPCClient

pkg/client/rpc_client.go
dhaidashenko: We don't need to disallow using the pending block number. If a caller is aware of what they are doing, it should be fine. At the same time, the argument is not documented as we do not provide any guarantees on the behaviour.

Unheilbar (dismissed):

ilija42 (dismissed):

ilija42 (approved):

Unheilbar (approved):

PR #146, opened 2025-07-11
using now() for ht insert/delete

yashnevatia:


pkg/heads/orm.go
```diff
@@ -64,8 +64,8 @@ func (orm *DbORM) batchInsertHeads(ctx context.Context, heads []*evmtypes.Head)
 	}
```
dhaidashenko: There is no need to set now to the go struct now

dhaidashenko (approved):

pkg/heads/orm.go
```diff
@@ -56,16 +55,11 @@ func (orm *DbORM) batchInsertHeads(ctx context.Context, heads []*evmtypes.Head)
 	if len(heads) == 0 {
 		return nil
 	}
-	// explicitly set created_at to now()
-	// as now() does not work with batch inserts
```
jmank88, line 60: Was this just mistaken? Or did something change?

jmank88 (approved):

pkg/heads/orm.go
yashnevatia: if you run the in memory test, created_at actually does not get populated.
so i thought i need to do that.
but then we saw the existing db dump and seems like its not an issue.
so sticking with using now() (it just behaves differently when writing to the actualy db)

PR #148, opened 2025-07-11
Bindings Generator bug fixes

silaslenihan:
- Update to latest SDK version
- Fix read bindings
- remove getchainID


akhilchainani (dismissed):

pablolagreca (dismissed):

ilija42 (dismissed):

pkg/bindings/bindings_test.go
```diff
@@ -126,6 +128,17 @@ func TestReadMethods(t *testing.T) {
 	ds, err := datastorage.NewDataStorage(client, nil, &bindings.ContractInitOptions{})
 	require.NoError(t, err, "Failed to create DataStorage instance")
 
+	client.EXPECT().LatestAndFinalizedHead(mock.Anything, mock.Anything).Return(
```
ilija42: fyi dmytro removed this function

silaslenihan: this hasn't been reflected in the sdk yet has it?

silaslenihan: I will remove these calls and readd them when the SDK is updated

pkg/bindings/testdata/DataStorage.sol
```diff
@@ -13,7 +13,9 @@ contract DataStorage {
 
     // Custom error for when a key is not found
     error DataNotFound(address requester, string key, string reason);
-
+    // duplicate error for testing
+    error DataNotFound2(address requester, string key, string reason);
```
ilija42, line 17: I don't see this being used anywhere

ilija42 (approved):

pkg/bindings/testdata/DataStorage.sol
silaslenihan: Previously it was causing a compile-time error when the bindings were generated, so including it here and having no errors was enough to test that particular issue

pablolagreca (approved):

PR #154, opened 2025-07-16
Operations approval in Workflow Registry

ibrajer:
Ticket: https://smartcontract-it.atlassian.net/browse/DEVSVCS-2052


contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -59,6 +59,9 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   // When a workflow is paused it is removed from the don family.
   mapping(bytes32 rid => bytes32 donHash) private s_donByWorkflowRid;
 
+  // Tracking approved operations for the owner address, required to enable MSIGs to verify off-chain requests.
+  mapping(address owner => mapping(bytes32 operationDigest => uint256 expiryTimestamp)) private s_approvedOperations;
```
DeividasK:
```suggestion
  mapping(address owner => mapping(bytes32 requestDigest => uint256 expiryTimestamp)) private s_approvedRequests;
```

Just to keep the language aligned, maybe we could use `requestDigest`? This is how the digest would be created: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/jsonrpc2/types.go#L56

DeividasK: Also, maybe `requestsAllowlist` instead of `approved`? Approved implies completion, but the request would still need to be sent to the service.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -85,6 +88,7 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   event WorkflowDonFamilyUpdated(
     bytes32 indexed workflowId, address indexed owner, string oldDonFamily, string newDonFamily
   );
+  event OperationApproved(address indexed owner, bytes32 indexed operationDigest, uint256 expiryTimestamp);
```
DeividasK:
```suggestion
  event RequestAddedToAllowlist(address indexed owner, bytes32 indexed requestDigest, uint256 expiryTimestamp);
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1227,6 +1231,39 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
     }
   }
 
+  // ================================================================
+  // |                  Approved operations                         |
+  // ================================================================
+  /// @notice Approves an operation for a specific owner, allowing MSIGs to verify off-chain requests.
```
DeividasK: It wouldn't be just multisigs, we agreed that all users (EOAs and MSIGs) would follow this path.

contracts/src/v0.8/workflow/dev/v2/test/WorkflowRegistry/WorkflowRegistry.approveOperation.tree
```diff
@@ -0,0 +1,5 @@
+WorkflowRegistry.approveOperation
+├── when the user is not linked
+│   └── it should revert with OwnershipLinkDoesNotExist
+└── when the user is linked
+    └── it should accept the operation digest
```
DeividasK: How do you generate this?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1227,6 +1231,39 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// @custom:revert OwnershipLinkDoesNotExist  If the caller is not a linked owner.
+  /// @dev User flow:
+  /// - User generates the digest of the operation.
+  /// - User calls approveOperation(operationDigest) on the Workflow Registry.
+  /// - Any user can then send the operation payload to the Vault DON.
+  /// - Vault DON checks if the digest is on-chain for verification purposes.
+  function approveOperation(bytes32 operationDigest, uint256 expiryTimestamp) external {
+    if (!s_linkedOwners.contains(msg.sender)) revert OwnershipLinkDoesNotExist(msg.sender);
+
+    s_approvedOperations[msg.sender][operationDigest] = expiryTimestamp;
```
DeividasK: Should we maybe warn if the expiry timestamp is outdated?

contracts/src/v0.8/workflow/dev/v2/test/WorkflowRegistry/WorkflowRegistry.approveOperation.tree
ibrajer: Some ASCII tree VSC plugin suggested by https://github.com/alexfertel/bulloak
Personally, I dislike the tool, but I have to stay aligned with the rest of the tests

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I was just copying from the doc that you provided. Can you give me a better description to put here?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Naming is completely up to you :) I copied from the doc, but I can align with the suggestions above

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: What do you mean by "warn"?

DeividasK: Revert 😅

ibrajer: Yeah that makes sense

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -59,6 +59,9 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   // When a workflow is paused it is removed from the don family.
   mapping(bytes32 rid => bytes32 donHash) private s_donByWorkflowRid;
 
+  // Tracking allowlisted requests for the owner address, required to enable anyone to verify off-chain requests.
+  mapping(address owner => mapping(bytes32 requestDigest => uint256 expiryTimestamp)) private s_requestsAllowlist;
```
shileiwill, line 63: ok, this is add only, no removal. 🤔

shileiwill (approved):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Yes, that's the idea according to the requirements from the ticket

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1227,6 +1232,40 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  }
+
+  /// @notice Checks if a request is allowlisted for a specific owner.
+  /// @dev    - Returns true if the request is allowlisted and not expired.
+  ///         - The request is considered allowlisted if the current block timestamp is less than the expiry timestamp.
+  /// @param owner            The address of the linked owner who allowlisted the request.
+  /// @param requestDigest    Unique identifier for the request (hash of the request payload).
+  /// @return bool            True if the request is allowlisted and not expired, false otherwise.
+  function isRequestAllowlisted(address owner, bytes32 requestDigest) external view returns (bool) {
+    return s_requestsAllowlist[owner][requestDigest] > block.timestamp;
```
eutopian, line 1266: Just thinking about this in terms of the timestamp, what does that mean? That we should only revert when it is added to allow list? But what if someone calls this in the future past that timestamp?

And to Lei's question, what about prunings and removal of approved or expired old requests?

ibrajer: I will refer to @DeividasK who was the original author of this feature.
My understanding is that the request is only allowlisted for a limited amount of time, and you can't complete off-chain verification past the expiry date. If that happens, you must either re-send the original request or send a new one and then submit the request digest to the registry contract.

We could add "cancelRequest" but someone will have to call it with the original request payload. TBH, I don't think anyone will spend gas on it, leaving it to expire is a better option.

Deletions are expensive in terms of gas. Why would we want to have a function that would prune expired requests and why would we (or anyone, it can be permissionless) pay for it?

anirudhwarrier (approved):

PR #158, opened 2025-07-17
workflow registry: update canLinkOwner 

eutopian:
to take the address instead of relying on msgsender. This is because calling this function otherwise in a block explorer with an unlinked wallet will default to address(0) and show the users misleading information


ibrajer (dismissed):
LGTM

ibrajer (approved):

timothyF95 (approved):

PR #166, opened 2025-07-29
Fix Code Owners for components owned by PLEX/Framework

dhaidashenko:


.github/CODEOWNERS
```diff
@@ -73,8 +73,12 @@
...
-go.sum @smartcontractkit/core @smartcontractkit/foundations @smartcontractkit/core-solidity
+go.mod @smartcontractkit/core @smartcontractkit/foundations @smartcontractkit/core-solidity @smartcontractkit/bix-framework
+go.sum @smartcontractkit/core @smartcontractkit/foundations @smartcontractkit/core-solidity @smartcontractkit/bix-framework
 
 pkg/txm @dimriou @smartcontractkit/bix-framework @smartcontractkit/core
 pkg/bindings @smartcontractkit/bix-framework
+pkg/client @smartcontractkit/bix-framework
+pkg/logpoller @smartcontractkit/bix-framework
+pkg/heads @smartcontractkit/bix-framework
+pkg/writetarget @smartcontractkit/bix-framework
```
jmank88, line 84: Should this be the default from the root (line 10) so we don't have to repeat it? Or this + core?

jmank88: Perhaps the go.mod/sum could also be reduced to this default - can drop foundations given core, and why do solidity folks care about go changes?

PR #170, opened 2025-08-03
Prepare repo for Solidity NPM 1.4.1 release

matYang:
✅ updated version to 1.4.1 in package.json
✅ reviewed all PRs and updated CHANGELOG.md to include relevant PRs
✅ forge test with all individual profiles
✅ make wrappers-all
✅ make wrappers-zksync

Solidity changes excluded from CHANGELOG:
1. Contracts still marked with -dev suffix, eg most l2ep contracts
2. Contracts under /dev, e.g. workflow/dev/v2/CapabilitiesRegistry.sol


gethwrappers/zksync/go_generate.go
```diff
@@ -8,4 +8,4 @@ package gethwrappers
 //go:generate go run ../generation/zksync/wrap.go shared WETH9ZKSync weth9_zksync
 //go:generate go run ../generation/zksync/wrap.go shared MockV3Aggregator mock_v3_aggregator_contract
 
-//go:generate go run ../generation/zksync/wrap.go automation MockETHUSDAggregator mock_ethusd_aggregator_wrapper
\ No newline at end of file
+//go:generate go run ../generation/zksync/wrap.go automation MockETHUSDAggregator mock_ethusd_aggregator_wrapper
```
matYang: wrapper gen errors if missing newline at EOF

contracts/CHANGELOG.md
```diff
@@ -1,8 +1,48 @@
 # @chainlink/contracts
 
+## 1.4.1
+
+### Minor Changes
+
+- [#164](https://github.com/smartcontractkit/chainlink-evm/pull/164) [`bf75f8a`](https://github.com/smartcontractkit/chainlink-evm/commit/bf75f8a) - Adding a variety of audited flavours of the CCT token (BurnMintERC20) that implements UUPS and Transparent Proxy upgradeability patterns. Brings proxy-related OZ contracts into vendor.
+
+- [#121](https://github.com/smartcontractkit/chainlink-evm/pull/121) [`e479a8d`](https://github.com/smartcontractkit/chainlink-evm/commit/e479a8d409) - Add BurnMintERC20WithDrip to replace the BurnMintERC677Helper. This should serve as the new standard BurnMintERC20 implementation.
```
RensR: > standard BurnMintERC20 implementation

not `BurnMintERC20` but something like `drippable token` idk, it's just a token for testnet, we don't want customers accidentally deploying a token with a free mint function

contracts/CHANGELOG.md
```diff
@@ -1,8 +1,48 @@
...
+
+- [#117](https://github.com/smartcontractkit/chainlink-evm/pull/117) [`2820442`](https://github.com/smartcontractkit/chainlink-evm/commit/282044242e) - Make a number of functions virtual in BurnMintERC20 and BurnMintERC677.
```
RensR: Do we need this in the changelog? Imo we should not list every PR. It becomes very noisy. Same for most of the other notes below

contracts/CHANGELOG.md
matYang: worded to "This provides a token with a public drip function, allowing users to test CCIP token transfers."

contracts/CHANGELOG.md
matYang: hmm the previous changelog is pretty extensive, if we don't include every PR, then we need a consistent bar on what to / not do include, that's more difficult

using this release as example, maybe 117, 116, 69, 55, and 46 can be excluded?

RensR: Just because we did it badly in the past does't mean we have to stick to it :) I fully agree that we should align on some clear set of rules for the changelogs. 

I would remove all of those and more. Also 118, 130, 3, 39, 41

matYang: hmm I suppose we are centring around changes to actual Solidity code, as opposed to including tooling or minor foundry versions; cleaned up the above except PR 39, it does actually touch [ScrollValidator.sol](https://github.com/smartcontractkit/chainlink-evm/pull/39/files#diff-a539e361893f14744c12cc833f1efe690d441c0f80433b51a6212f083597174b)

RensR (dismissed):

pavel-raykov (dismissed):

Farber98 (dismissed):

amit-momin (dismissed):

RensR (dismissed):

.github/workflows/solidity-foundry.yml
```diff
@@ -203,6 +203,7 @@ jobs:
         uses: baptiste0928/cargo-install@904927dbe77864e0f2281519fe9d5bd097a220b3 # v3.1.1
         with:
           crate: svm-rs
+          version: '0.5.17'
```
matYang, line 206: latest version of svm-rs 0.5.18 released yesterday requires rustc 1.89, but gh default rust is 1.88; between setting gh rust toolchain to 1.89 v.s setting svm-rs to 0.5.17, the latter is more straightforward and reliable

contracts/scripts/native_solc_compile_all_shared
```diff
@@ -12,34 +12,65 @@ FOUNDRY_PROJECT_SUFFIX=""
 CONTRACTS_DIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; cd ../ && pwd -P )"
 export FOUNDRY_PROFILE="$PROJECT"$FOUNDRY_PROJECT_SUFFIX
 
-compileContract() {
+SOLC_VERSION_DEFAULT="0.8.19"
+SOLC_VERSION_026="0.8.26"
```
matYang: only use 0.8.26 for the upgradable contracts to avoid re-compiling many existing contracts

gethwrappers/shared/generation/generated-wrapper-dependency-versions-do-not-edit.txt
```diff
@@ -1,11 +1,18 @@
 GETH_VERSION: 1.15.3
 aggregator_v3_interface: ../../contracts/solc/shared/AggregatorV3Interface/AggregatorV3Interface.sol/AggregatorV3Interface.abi.json ../../contracts/solc/shared/AggregatorV3Interface/AggregatorV3Interface.sol/AggregatorV3Interface.bin 0294226c10dd4b77239d85074fcf05c39c5cbf513b45bb07f966a8733457d2cb
 burn_mint_erc20: ../../contracts/solc/shared/BurnMintERC20/BurnMintERC20.sol/BurnMintERC20.abi.json ../../contracts/solc/shared/BurnMintERC20/BurnMintERC20.sol/BurnMintERC20.bin 161d4fc00d3bd8759a82a11d2d0946653ed06853e67359049ac4176288090d22
-burn_mint_erc20_with_drip: ../../contracts/solc/shared/BurnMintERC20/BurnMintERC20.sol/BurnMintERC20.abi.json ../../contracts/solc/shared/BurnMintERC20/BurnMintERC20.sol/BurnMintERC20.bin 161d4fc00d3bd8759a82a11d2d0946653ed06853e67359049ac4176288090d22
```
matYang, line 4: burn_mint_erc20_with_drip was generated from BurnMintERC20, looks like the bug before

gethwrappers/shared/go_generate.go
```diff
@@ -6,7 +6,7 @@ package gethwrappers
 //go:generate go run ../generation/wrap.go shared ERC677 erc677 latest
 //go:generate go run ../generation/wrap.go shared LinkToken link_token latest
 //go:generate go run ../generation/wrap.go shared BurnMintERC20 burn_mint_erc20 latest
-//go:generate go run ../generation/wrap.go shared BurnMintERC20 burn_mint_erc20_with_drip latest
+//go:generate go run ../generation/wrap.go shared BurnMintERC20WithDrip burn_mint_erc20_with_drip latest
```
matYang, line 9: burn_mint_erc20_with_drip was generated from BurnMintERC20, looks like the bug before

matYang (commented):
reverting the go.mod under v1_4_1 for now, because wrappers have a dependency on 
```
type AbigenLog interface {
	Topic() common.Hash
}
```

it lives inside `generated.go` under generated folder; every wrapper imports it, this import path is baked into abi.gen's AST. 
As a result, if we add a go.mod under v1_4_1, that go.mod must take a dependency on the parent go.mod which isn't published yet and kinda defeats the purpose.
A better solution may be refactoring out AbigenLog into a published mini go module, and update abi.gen's AST to use the mini module instead, however this will produce a diff for all wrappers. This is invasive and should be its own task.

erikburt (dismissed):

RensR (approved):

erikburt (approved):

PR #172, opened 2025-08-04
Updated TRON TXM to use combinedClient

silaslenihan:


pkg/chains/tron/txm.go
```diff
@@ -22,19 +24,28 @@ func ConstructTxm(logger logger.Logger, cfg TxmConfig, nodes []*toml.Node, keyst
 	}
 
 	fullNodeURL := nodes[0].HTTPURLExtraWrite.URL()
-	tronClient, err := tronclient.CreateFullNodeClient(fullNodeURL)
+	solidityNodeURL, err := url.ParseRequestURI(fullNodeURL.String() + "solidity")
```
pavel-raykov: just checking - I am not sure what kind of path you expect here, but commonly url concatenation is done via library functions like https://pkg.go.dev/net/url#JoinPath instead of the string manipulation (+). Is the + necessary here?

silaslenihan: fixed

pkg/chains/tron/txm.go
```diff
@@ -22,19 +23,25 @@ func ConstructTxm(logger logger.Logger, cfg TxmConfig, nodes []*toml.Node, keyst
 	}
 
 	fullNodeURL := nodes[0].HTTPURLExtraWrite.URL()
-	tronClient, err := tronclient.CreateFullNodeClient(fullNodeURL)
+	solidityNodeURL := fullNodeURL.JoinPath("solidity")
+
+	combinedClient, err := sdk.CreateCombinedClient(fullNodeURL, solidityNodeURL)
```
KodeyThomas: tron's endpoints can be path based but i don't know if we can assume that.

for some of our nodes it'll end in `/wallet` for the full node url so this would work but i'm not 100% certain it would always work. we may need another config value

KodeyThomas (changes requested):

pkg/chains/tron/txm.go
silaslenihan: is there a way we can verify whether it will sometimes not be path based? My understanding was that the HTTP API is shared between the `solidity` and `walletsolidity` endpoints which we treat as different clients.

silaslenihan: It seems they can sometimes have a different port

pkg/chains/tron/txm.go
```diff
@@ -22,19 +23,27 @@ func ConstructTxm(logger logger.Logger, cfg TxmConfig, nodes []*toml.Node, keyst
...
-	return trontxm.New(logger, tronkeystore.NewLoopKeystoreAdapter(keystore), tronClient, trontxm.TronTxmConfig{
+	return trontxm.New(logger, tronkeystore.NewLoopKeystoreAdapter(keystore), combinedClient, trontxm.TronTxmConfig{
 		// Overrides the energy estimator to always use the fixed energy
 		FixedEnergyValue: fixedEnergyValue,
 		// Maximum number of transactions to buffer in the broadcast channel.
 		BroadcastChanSize: 100,
 		// Number of seconds to wait between polling the blockchain for transaction confirmation.
 		ConfirmPollSecs: 5,
+		// How long transactions are kept in the txm
+		RetentionPeriod: 8 * time.Minute,
```
amit-momin: Noticed the same retention here period here but this should be easier to update than defaults later so should be ok

amit-momin (dismissed):

KodeyThomas (dismissed):

pkg/chains/tron/txm.go
silaslenihan: I reverted it here as well.

KodeyThomas (approved):

ilija42 (approved):

PR #179, opened 2025-08-07
move shared into versioned folders

RensR:
The main goal of this PR was to split the `shared` gethwrappers into a `latest` folder which will contain in-progress work, and other folders. Each future release will have a versioned folder, that will contain the gethwrappers for all of the contracts that belong to that release. If the contract is not in there, it cannot be assumed to be part of the release. 

This PR also adds a separate go.mod for the `gethwrappers` folder. Some changes have been made to ensure it didn't import chainlink-common or chainlink-evm. This standalone go.mod can now more easily be imported by anyone requiring gethwrappers. 

Due to this package split, CI had to be adjusted to run for the new package as well. There it was found that the root-level and gethwrapper tests didn't even run before the changes had been made! This is fixed now, and all tests are green again.


.github/workflows/build-test.yml
```diff
@@ -3,57 +3,101 @@ name: Build and Test
...
     env:
       # We explicitly have this env var not be "CL_DATABASE_URL" to avoid having it be used by core related tests
       # when they should not be using it, while still allowing us to DRY up the setup
       DB_URL: postgresql://postgres:postgres@localhost:5432/chainlink_test?sslmode=disable
+
+    strategy:
+      fail-fast: false
+      matrix:
+        type:
+          # TODO only run when needed
```
RensR, line 18: Should only run based on changes, out of scope for this PR as this was already not the being done

.github/workflows/build-test.yml
```diff
@@ -3,57 +3,101 @@ name: Build and Test
...
         with:
           name: go-test-results
           path: |
-            pkg/coverage.txt
-            pkg/race.*
+            ${{ matrix.type.path }}/pkg/coverage.txt
+            ${{ matrix.type.path }}/pkg/race.*
+
+
+  build-test:
```
RensR, line 90: Using the same name ensures the required checks still pass

gethwrappers/shared/go_generate.go
```diff
@@ -2,20 +2,20 @@
...
-//go:generate go run ../generation/wrap.go shared BurnMintERC20 burn_mint_erc20_with_drip
-//go:generate go run ../generation/wrap.go shared WERC20Mock werc20_mock
-//go:generate go run ../generation/wrap.go shared ChainReaderTester chain_reader_tester
-//go:generate go run ../generation/wrap.go shared AggregatorV3Interface aggregator_v3_interface
-//go:generate go run ../generation/wrap.go shared MockV3Aggregator mock_v3_aggregator_contract
-//go:generate go run ../generation/wrap.go shared LogEmitter log_emitter
-//go:generate go run ../generation/wrap.go shared VRFLogEmitter vrf_log_emitter
-//go:generate go run ../generation/wrap.go shared ITypeAndVersion type_and_version
-//go:generate go run ../generation/wrap.go shared WETH9ZKSync weth9_zksync
+//go:generate go run ../generation/wrap.go shared BurnMintERC677 burn_mint_erc677 latest
```
RensR, line 5: new wrappers are always stored in the `latest` folder. Only when a version is released are contracts moved from `latest` to their versioned folder.

pkg/config/toml/docs.toml
```diff
@@ -535,7 +535,7 @@ Order = 100 # Default
 # IsLoadBalancedRPC indicates whether the http/ws url above has multiple rpc's behind it.
 # If true, we should try reconnecting to the node even when its the only node in the Nodes list.
 # If false and its the only node in the nodes list, we will mark it alive even when its out of sync, because it might still be able to send txs.
-IsLoadBalancedRPC = false
+IsLoadBalancedRPC = false # Example
```
RensR, line 538: This was missing, but since the test for the docs was not running, it was not caught

matYang (approved):

pavel-raykov (approved):

PR #188, opened 2025-08-12
Add MetaClient verification

dimriou:


pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -174,7 +219,96 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
...
+	if err != nil {
+		return nil, err
+	}
+
+	info := params["info"]
+	if len(info) != 2 {
+		return nil, fmt.Errorf("incorrect size for info: %v", info)
+	}
+	to := common.HexToAddress(info[0]) // metacall address
+	dApp := common.HexToAddress(info[1])
```
LunkRat: Are we sure that an array of strings is the best design here? The worry about this design is that the positional requirements in the array are hard to validate and track across deployment systems and tooling. 

One alternative could be to use a TOML table with named params for `to` and `dApp`, maybe like the below?

```toml
[relayConfig.dualTransmission.meta.info]
to = "0x000000..." # metacall address
dApp = "0x000000..."
```

PR #196, opened 2025-08-25
View function for fetching non-expired allowlisted request digests

ibrajer:


cedric-cordenier (dismissed):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1327,7 +1339,41 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   /// @param requestDigest    Unique identifier for the request (hash of the request payload).
   /// @return bool            True if the request is allowlisted and not expired, false otherwise.
   function isRequestAllowlisted(address owner, bytes32 requestDigest) external view returns (bool) {
-    return s_requestsAllowlist[owner][requestDigest] > block.timestamp;
+    return s_requestsAllowlist[_ownerDigestHash(owner, requestDigest)] > block.timestamp;
```
eutopian: can you inline this like the rest of the hash functions that are only used once inside the function

ibrajer: You mean inline `_ownerDigestHash`? It's used twice, but I can inline if that's a preference

shileiwill (dismissed):
+1'd, will need to fix test tho

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Removed inline function

eutopian (approved):

shileiwill (approved):

PR #197, opened 2025-08-25
Upper bound on the expiration timestamp for allowlisted requests

ibrajer:


contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1312,8 +1330,12 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   /// - User calls allowlistRequest(requestDigest) on the Workflow Registry.
   /// - Any user can then send the request payload to the Vault DON.
   /// - Vault DON checks if the digest is on-chain for verification purposes.
-  function allowlistRequest(bytes32 requestDigest, uint256 expiryTimestamp) external {
+  function allowlistRequest(bytes32 requestDigest, uint32 expiryTimestamp) external {
     if (expiryTimestamp <= block.timestamp) revert RequestExpired(requestDigest, expiryTimestamp);
+    uint32 maxAllowedExpiry = getMaxExpiry();
+    if (expiryTimestamp - block.timestamp > maxAllowedExpiry) {
+      revert RequestOverMaxExpiry(requestDigest, expiryTimestamp, maxAllowedExpiry);
```
shileiwill: I think we can combine the 2 errors: RequestExpired in L1334 and this RequestOverMaxExpiry. We can call them InvalidExpiryTimestamp. `RequestExpired` here could be misleading bc this is where we set expiry, and the timestamp is a invalid one.

Nit, align the comparison <= v.s. <. I recommend using `<` or `>` (without equal) to save a bit gas.

shileiwill (commented):
Left a comment, otherwise LGTM.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -36,21 +36,23 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  /// EXPIRY_SHIFT (32 bits) = 40
   ///
   /// @dev Single-slot override; when s_metadataCaps == 0, all defaults apply.
   /// @dev DEFAULT_PACKED bundles the defaults:
   ///   (uint256(64)  << NAME_SHIFT) │
   ///   (uint256(32)  << TAG_SHIFT)  │
   ///   (uint256(200) << URL_SHIFT)  │
   ///   (uint256(1024)<< ATTR_SHIFT)
-  /// where maxNameLen = 64, maxTagLen = 32, maxUrlLen = 200, maxAttrLen = 1024
+  /// where maxNameLen = 64, maxTagLen = 32, maxUrlLen = 200, maxAttrLen = 1024, maxExpiry = 604800
```
eutopian: The packing doesn't make sense anymore if this is uint256 as that's just a slot on its own. Let's discuss and I think adding this is the right place to moving the config out of the compile time const into a strict

eutopian: The packing doesn't make sense anymore if this is uint256 as that's just a slot on its own. Let's discuss and I think adding this is the right place to moving the config out of the compile time const into a struct

eutopian (changes requested):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I switched to using `uint32` for the expiry timestamp - `uint256` was an overkill
In that case, would you still recommend splitting it or keeping it the way it is now?

ibrajer: Per today's discussion - decided to split the entire config into a separate struct for better readability

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -199,6 +180,15 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   // ================================================================
   // |                         Structs                              |
   // ================================================================
+  /// @dev Struct for Workflow Registry configuration parameters
+  struct Config {
+    uint8 maxNameLen; // Cap for `workflowName` (0 ➜ unlimited)
+    uint8 maxTagLen; // Cap for `tag` (0 ➜ unlimited)
+    uint8 maxUrlLen; // Cap for each URL (0 ➜ unlimited)
+    uint16 maxAttrLen; // Cap for `attributes` (0 ➜ unlimited)
+    uint32 maxExpiryLen; // Cap for every allowlisted request expiration timestamp (0 ➜ unlimited)
```
shileiwill, line 189: what does unlimited mean for the above?

eutopian (approved):
LGTM! Amazing turnaround on the changes!

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1307,7 +1263,13 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
   /// - Any user can then send the request payload to the Vault DON.
   /// - Vault DON checks if the digest is on-chain for verification purposes.
   function allowlistRequest(bytes32 requestDigest, uint32 expiryTimestamp) external {
-    if (expiryTimestamp <= block.timestamp) revert RequestExpired(requestDigest, expiryTimestamp);
+    uint32 maxAllowedExpiry = s_config.maxExpiryLen; // 0 -> unlimited
+    if (
+      expiryTimestamp <= block.timestamp
```
shileiwill, line 1268:
```suggestion
      expiryTimestamp < block.timestamp
```

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1307,7 +1263,13 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+        || (maxAllowedExpiry != 0 && expiryTimestamp - block.timestamp > maxAllowedExpiry)
```
shileiwill, line 1269: maxAllowedExpiry will never be 0 since we set default in L33, no?

shileiwill (approved):
left nit comments. otherwise LG, approved to unblock

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: 0 means unlimited, you can set it to unlimited only via `setConfig`

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: It means there is no upper bound for the parameter (e.g. workflow name can be of any size)

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: No, actually, this is okay, if expiry is in the current block, it will expire as soon as the block is produced, so we can't accept that

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
shileiwill: that is basically the max_value of uint32?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
eutopian: so I guess in this sense, 0 would mean "never" in the old gh tokens that allowed you to set never expires that the fine grained ones now don't

PR #201, opened 2025-08-26
DEVSVCS-2369: Ensure donFamily is always string

eutopian:
using `donFamily` as the input value and use strings
internally on contract `donHash` should refer to the bytes32 key for mappings

this was the only instance when I went through the code again. All other input params take string donFamily


ibrajer (approved):
LGTM

shileiwill (approved):
LGTM

PR #205, opened 2025-08-29
[CRE-774] Correct CapabilitiesRegistry TypeAndVersion & small gas golf

justinkaseman:
### Description
Correction to CapabilitiesRegistry TypeAndVersion which should maintain `-dev` within the `dev` folder 

Additionally, for stability. freezes the current standing gethwrappers. Audit changes will go into new wrappers, then be merged back in.


eutopian (dismissed):

krehermann (changes requested):
we need to maintain the existing geth wrapper so we don't break everything staging as we make these fixes.

if you change the wrapper, all the imports in core will fail against the currently deployed contract.

ideally we generate the wrappers for these in-flight changes to a new place and then switch all the import or move the new wrapper to the existing location after all the code changes are commited

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -447,7 +447,7 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
   // |                 Internal variables                            |
   // ================================================================
 
-  string public constant override typeAndVersion = "CapabilitiesRegistry 2.0.0";
+  string public constant override typeAndVersion = "CapabilitiesRegistry 2.0.0-dev";
```
krehermann, line 450: was the wf registry type and version correct, and already had this -dev suffix?

justinkaseman: Yeah, WF Registry has the correct `-dev` suffix already 😦

gethwrappers/workflow/generated/capabilities_registry_wrapper_v2/README.md
krehermann, line 1: thank you

krehermann (approved):

DeividasK (approved):

PR #207, opened 2025-08-29
[CRE-774] (feat): Prevent duplicate Node Operators in Capabilites Registry

justinkaseman:
### Description
There are no checks to prevent duplicate Node Operators from being added to the Capabilities Registry.

This PR ensures that Node Operators are unique by `hash(admin, name)` by adding a lookup mapping for if the data has been set.


mchain0 (dismissed):

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -539,6 +551,9 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
   ) external onlyOwner {
     for (uint32 i; i < nodeOperatorIds.length; ++i) {
       uint32 nodeOperatorId = nodeOperatorIds[i];
+      NodeOperator memory nodeOperator = s_nodeOperators[nodeOperatorId];
```
ibrajer, line 554: What if someone sends a node operation ID outside of the current range? For example, `s_nodeOperators` contains 10 nodes, but someone sent a list of node operator IDs: 20, 21, 22. Do we want to proceed with deletion if `nodeOperator` comes out empty?

ibrajer (approved):

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: > Do we want to proceed with deletion if `nodeOperator` comes out empty?

Yes, as this would be a no-op. Reverting wouldn't serve much of a purpose here.

DeividasK (approved):

PR #208, opened 2025-08-29
[CRE-774] Prevent node operators from being deleted if still on a node

justinkaseman:
### Description
To prevent NOPs from being deleted if still in use, adds an EnumerableSet of bytes32 P2PIDs to the Node Operator struct.

A new struct `NodeOperatorParams` acts as the Node Operator structure for parameters and query info.


mchain0 (dismissed):

DeividasK (approved):

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -31,6 +31,28 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
...
+
+  struct NodeOperatorInfo {
+    /// @notice The address of the admin that can manage a node operator
+    address admin;
+    /// @notice Human readable name of a Node Operator managing the node
+    /// @dev The contract does not validate the length or characters of the node operator name because
+    /// a trusted admin will supply these names. We reduce gas costs by omitting these checks on-chain.
+    string name;
+    /// @notice The P2P IDs of the nodes that this Node Operator owns
+    bytes32[] nodeP2PIDs;
```
ibrajer, line 55: In case we do something like this:
```
struct NodeOperatorInfo {
  NodeOperatorParams params;
  bytes32[] nodeP2PIDs;
}
```
Would it result in the same storage and memory layout?

ibrajer (approved):

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: It would be a nested structure and, _I think_, it would have the same storage and memory layout, yes. Generally, it seems that the practice is to flatten the structs, but I don't have a strong pref 🤷‍♂️

PR #211, opened 2025-08-29
(test): Add build external test against smartcontractkit/chainlink

justinkaseman:
### Problem
While a contract is under active development, the gethwrappers may often change.

There is risk that once the integration with the core node has happened, that any changes in the contract's gethwrapper will halt bumping the `chainlink-evm` dependency on the core node side until a new contract has been deployed.

This creates uncertainty which slows development process. 

In this time of frequent iteration, freezing the wrappers becomes cumbersome.

### Changes

This PR introduces non-blocking testing to flag any breaking interface changes early. It clones the core repo, updates the `chainlink-evm` dependency and attempts a build.

Note that if this test passes it doesn't mean that the underlying behavior is the same, just that the current usage in Chainlink maintains the same interfaces.

Github workflow lifted from [chainlink-common](https://github.com/smartcontractkit/chainlink-common/blob/main/.github/workflows/build_external.yml)

### DIFF
--- a/.github/scripts/get-refs-from-pr-body.js
+++ b/.github/scripts/get-refs-from-pr-body.js
@@ -1,0 +1,205 @@
+/**
+ * Configuration for repositories and their default refs
+ * Add new repos here to automatically include them in the ref extraction
+ */
+const REPO_CONFIG = {
+  core: {
+    pattern: /core ref:\s*(\S+)/i,
+    defaultRef: "develop",
+    outputKey: "core-ref",
+  },
+  // Add new config here:
+  // <name>: {
+  //   pattern: /<name> ref:\s*(\S+)/i,
+  //   defaultRef: "develop",
+  //   outputKey: "<name>-ref"
+  // }
+};
+
+/**
+ * Validates a Git reference (branch name, tag, or commit SHA).
+ *
+ * These git refs are extracted from PR body text and must be validated for safety.
+ *
+ * @param {string} ref - The Git reference to validate
+ * @returns {string} - The validated reference
+ * @throws {Error} - If the reference is invalid
+ */
+function validateGitRef(ref) {
+  if (!ref) return null;
+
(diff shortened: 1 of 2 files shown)


.github/scripts/get-refs-from-pr-body.js
```diff
@@ -0,0 +1,205 @@
+/**
```
chainchad, line 1: Is this the same script from https://github.com/smartcontractkit/chainlink-common/blob/main/.github/scripts/get-refs-from-pr-body.js ?

If so, would you be up for creating a small composite action on the /.github repo? I can help if needed.

.github/workflows/build-external.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Checkout the chainlink core repo
+        uses: actions/checkout@v4
+        with:
+          repository: smartcontractkit/chainlink
+          ref: ${{ needs.init.outputs.core-ref }}
+          persist-credentials: false
+          path: chainlink
+
+      - name: Setup Go
```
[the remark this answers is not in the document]
justinkaseman: @chainchad What do you make of this?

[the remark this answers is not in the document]
chainchad: I think we can probably dismiss this since we're checking out an internal repo here. If we were bringing in an external repo then we'd have to be more careful.

.github/scripts/get-refs-from-pr-body.js
justinkaseman: This is a good idea. It would be helpful to have this same check in `chainlink-protos` as well.

chainchad: Created DX-1667 to track that 👍

chainchad (approved):

erikburt (approved):

PR #214, opened 2025-09-02
Update FixtureChainID

vreff:
Picking up from: https://github.com/smartcontractkit/chainlink-evm/pull/17

Update FixtureChainID to avoid using an invalid chain ID (0) that causes a panic in newer versions of go-ethereum. This should unblock us from bumping go-ethereum in a subsequent PR.

There are quite a few places where a zero chain ID is implied by `evmChainID` or `chainID` simply being absent from structs we declare. Now the chain ID is overtly declared in these structs to be our `FixtureChainID`.


RensR:
Can we add codeowners for the `pkg` dir? Right now most of it is owned by the root repo owners.

Unheilbar (dismissed):

dhaidashenko (dismissed):

pkg/types/utils.go
```diff
@@ -13,3 +13,5 @@ func MustGetABI(json string) abi.ABI {
 	}
 	return abi
 }
+
+const NullClientChainID = 1399100
```
jmank88, line 20: Why this particular number? Could use a comment explaining.
Do we have a plan for future collisions? Can we just change it in that case?

vreff: I used the same number from https://github.com/smartcontractkit/chainlink-evm/pull/17, and I don't know why that was chosen. I do add a comment over FixtureChainID but I can bring that over this as well.

You'll see that things have been refactored so that yes, we can just change this one value and bump imports downstream, and everything should automatically switch over.

jmank88 (approved):

pavel-raykov (approved):

PR #220, opened 2025-09-08
Prepare repo for Solidity NPM 1.5.0 release

RayXpub:
Replaces 1.4.1 release (https://github.com/smartcontractkit/chainlink-evm/pull/170)

✅ updated version to 1.5.0 in package.json
✅ reviewed all PRs and updated CHANGELOG.md to include relevant PRs
✅ forge test with all individual profiles
✅ make wrappers-all
✅ make wrappers-zksync

Solidity changes excluded from CHANGELOG:

Contracts still marked with -dev suffix, eg most l2ep contracts
Contracts under /dev, e.g. workflow/dev/v2/CapabilitiesRegistry.sol


tools/bin/build_abigen
```diff
@@ -5,16 +5,20 @@
 
 set -e
 
+ROOT_DIR="$(git rev-parse --show-toplevel)"
 # Version of abigen to install. Must be run within chainlink project
-GETH_VERSION=$(go list -json -m github.com/ethereum/go-ethereum | jq -r .Version)
+GETH_VERSION=$(go -C "$ROOT_DIR/gethwrappers" list -json -m github.com/ethereum/go-ethereum | jq -r .Version)
```
RayXpub: I had to update the script logic because previously it was fetching the geth version from root which should now be fetched from the `gethwrappers` sub package. This was causing `make abigen` to fail since current root version is `1.16.2` and `gethwrappers` is `1.15.3`.

contracts/snapshots/DataFeedsCacheGasTest.json
```diff
@@ -0,0 +1,30 @@
+{
```
RensR: This always happens when you run snapshot for data feeds but shouldn't be committed. Can you remove?

contracts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@chainlink/contracts",
-  "version": "1.4.1",
+  "version": "1.5 0",
```
RensR: ` ` instead of `.`

contracts/snapshots/DataFeedsCacheGasTest.json
RayXpub: done

contracts/package.json
RayXpub: fixed

RensR (dismissed):

tools/bin/build_abigen
DeividasK: I would have expected us to want to keep these two versions in sync 🤷‍♂️  in which case this check might actually be doing a hidden "versions are matching" test.

RayXpub: That's a good call out. I updated the script to catch the version drift and error with the correct log.

RensR (approved):

pavel-raykov (approved):

DeividasK (approved):

PR #236, opened 2025-09-15
Fix SVR forwarder

dimriou:


cl-efornaciari (dismissed):

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -172,6 +172,7 @@ func (a *MetaClient) SendTransaction(ctx context.Context, tx *types.Transaction,
 		a.lggr.Info("No bids for transactionID(%d): ", tx.ID)
```
cl-efornaciari: I think this should be `a.lggr.Infof`

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -193,6 +194,9 @@ type RequestResponse struct {
 }
 
 type ResponseResult struct {
+	UO  *UORaw `json:"userOperation"`
+	SOS []*SO  `json:"solverOperations"`
+	DO  *DO    `json:"dAppOperation"`
```
cl-efornaciari: Can unmarshalling these fields fail? is it worth splitting this out this unmarshalling for logging from the critical path?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -309,6 +347,12 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 		return nil, nil
 	}
 
+	r, err := json.MarshalIndent(response.Result, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	a.lggr.Info("Response: ", string(r))
+
```
cl-efornaciari: can we add some tests to ensure the serialization works as expected

cl-efornaciari (commented):
A few comments/thoughts

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: The unmarshalling will occur on a best effort basis, and if the corresponding fields are not found, it will just leave them empty. One way this could fail is if the response is not a valid JSON object at all; however, that means the execution will fail at the next check. Having said that, I agree to simply log the error instead of returning, as it makes things more clear.

cl-efornaciari (approved):

silaslenihan (approved):

PR #237, opened 2025-09-16
Set default FinalityTagEnabled  on polygon and amoy to be false

infiloop2:
Multiple products have faced issue with this config as polygon has sent reorgs on blocks previously sent with finality tags.

Most products override this manually, changing the default for safety in future


pavel-raykov (dismissed):

RensR:
Can you please fix the codeowners for this folder? It should not be the catch-all `root` oweners

ilija42 (approved):

infiloop2:
@RensR addressed comment if you can help review

RensR (approved):

infiloop2:
per conversation on slack https://chainlink-core.slack.com/archives/C01616631EU/p1757413408051909 , aligned that this is good to set as default for products. Merging PR

PR #241, opened 2025-09-19
Default user limits per DON family

ibrajer:
When setting a DON family active workflows limit, ensure that the default user limit is also configured. With a separate function, the contract admin can always override the default user limit, but never above the global DON family limit. The default user limit ensures fairness for all users who depend on having access to a chunk of a DON family capacity.


contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -331,42 +336,63 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+  function setDONLimit(
+    string calldata donFamily,
+    uint32 donLimit,
+    uint32 userDefaultLimit,
+    bool enabled
+  ) external onlyOwner {
     bytes32 donHash = _hash(donFamily);
-    uint32 newCapacity = enabled ? limit : 0;
+    uint32 newDONCapacity = enabled ? donLimit : 0;
+    uint32 newDefaultUserCapacity = enabled ? userDefaultLimit : 0;
```
DeividasK: Nit (I know this isn't what you've introduced): It's a bit weird how this switches from "limit" to "capacity", which is a slightly different term.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -331,42 +336,63 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
     }
 
+    if (newDefaultUserCapacity > newDONCapacity) {
+      // The default user limit must not exceed the global DON limit
+      revert UserDONDefaultLimitExceedsDONLimit();
+    }
+
+    if (newDONCapacity != 0 && newDefaultUserCapacity == 0) {
+      // If the global DON limit is set, it doesn't make sense to have a zero default user limit
+      revert UserDONDefaultLimitIsZero();
```
DeividasK: A default limit of 0 workflows could make sense in a world where we want to only allow a whitelisted set of workflow owners using owner override limit.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -331,42 +336,63 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
-    cfg.limitValue.value = newCapacity;
+    cfg.limitValue.value = newDONCapacity;
+    cfg.defaultUserLimit = newDefaultUserCapacity;
 
     s_events.push(
       EventRecord({
         eventType: EventType.DONCapacitySet,
         timestamp: uint32(block.timestamp),
-        payload: abi.encode(donHash, newCapacity)
+        payload: abi.encode(donHash, newDONCapacity, newDefaultUserCapacity)
```
DeividasK: `newDefaultUserCapacity` doesn't need to be included in the payload, as it wouldn't affect the workflow distribution mechanism.

DeividasK: This could even go as far as skipping the call to `s_events.push` if only the `newDefaultUserCapacity` changes.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1522,8 +1551,13 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
 
-    // Determine the effective cap: start with the global cap, but if this user has an override, use that instead
-    uint32 cap = cfg.userOverride[owner].enabled ? cfg.userOverride[owner].value : cfg.limitValue.value;
+    // Determine the effective cap:
+    // - if this user has no override, use the default user limit set for this DON
+    // - if this user has an override, use that unless it is bigger than the global cap set for this DON
+    uint32 cap = cfg.defaultUserLimit;
+    if (cfg.userOverride[owner].enabled) {
+      cap = cfg.userOverride[owner].value < cfg.limitValue.value ? cfg.userOverride[owner].value : cfg.limitValue.value;
+    }
```
DeividasK, line 1543: What if we want to intentionally limit some users below the don-user limit? A scenario here would be to limit users after pausing their workflows so they cannot reactivate them.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Valid point - I will rename to `newDonLimit` and `newDefaultUserLimit`

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Yes, that's how the global limit of zero workflow was intended to be used. You can set it to zero and it will apply to all users, but as soon as you set a user override, then you'll have this user as allowlisted to use the DON regardless of the global limit being set to zero. Does that align with your expectations?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Gotcha, makes sense, I will update this in the code

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Well, in that case, I think the only way to do that would be to explicitly set the user override to either zero (if we want to disallow deploying workflows completely) or set it to a much lower value than the current defaultUserLimit.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
infiloop2: I think similar to deividas, I find it a bit confusing that a user override can override a DON limit

imo a user override should only override its user capacity. the don capacity should not be overridable by a user

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
infiloop2: as an outsider, who doesn't know about the internal details of workflow distribution mechanism, I find it very hard to understand when something would be needed or not by its algorithm.

Is it bad to leave EventRecord as generic record of events with all the fields? any consumer of this can ignore events which are not relevant to it

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Discussed over a call, decided that global DON capacity must always be respected.

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
DeividasK: > Is it bad to leave EventRecord as generic record of events with all the fields? any consumer of this can ignore events which are not relevant to it

Having additional information that is unused costs extra gas. There's no downside other than that.

contracts/src/v0.8/workflow/dev/v2/test/WorkflowRegistry/WorkflowRegistry.activateWorkflow.t.sol
```diff
@@ -28,7 +28,7 @@ contract WorkflowRegistry_activateWorkflow is WorkflowRegistrySetup {
   function test_activateWorkflow_WhenTheWorkflowExistsButOwnerDoesNotEqualCaller() external whenCallerIsLinked {
     // It reverts with CallerIsNotWorkflowOwner
     vm.prank(s_owner);
-    s_registry.setDONLimit(s_donFamily, 10, true);
+    s_registry.setDONLimit(s_donFamily, 100, 10);
```
DeividasK, line 31: Nit for future (don't change it now) - it helps to readability to have vars for these numbers, e.g., `DON_WORKFLOWS_LIMIT` and `DEFAULT_USER_WORKFLOWS_LIMIT`

DeividasK (dismissed):

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -330,43 +337,48 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
-    cfg.limitValue.value = newCapacity;
+    cfg.limit = donLimit;
+    cfg.defaultUserLimit = userDefaultLimit;
 
     s_events.push(
       EventRecord({
         eventType: EventType.DONCapacitySet,
         timestamp: uint32(block.timestamp),
-        payload: abi.encode(donHash, newCapacity)
+        payload: abi.encode(donHash, donLimit)
```
infiloop2, line 377: just verifying we don't need userDefaultLimit here?

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -381,37 +393,42 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
...
+        emit UserDONLimitSet(user, donFamily, userLimit);
+      } else if (ov.value != userLimit) {
+        // → was ON with a different limit, so update it with a new override
+        ov.value = userLimit;
+        emit UserDONLimitSet(user, donFamily, userLimit);
       } else {
-        // → already ON at exactly this cap, nothing to do
+        // → already ON at exactly this limit, nothing to do
         return;
       }
```
infiloop2: nit. imo a lot of this logic can be simplified by simply doing

```
ov.enabled = true;
ov.value = userLimit;
emit UserDONLimitSet(user, donFamily, userLimit);
```

infiloop2: there's not much extra gas sent when value is written to the same value (if the storage slot doesn't change) and also prevents silent acknowledgement by the contract without any logs

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
```diff
@@ -1519,14 +1541,24 @@ contract WorkflowRegistry is Ownable2StepMsgSender, ITypeAndVersion {
 
   function _enforceLimits(address owner, bytes32 donHash, string memory donFamily, uint32 pending) internal view {
     DonConfig storage cfg = s_donConfigs[donHash];
-    // Global limit must be explicitly enabled (zero or missing = disallowed)
-    if (!cfg.limitValue.enabled) revert DonLimitNotSet(donFamily);
+    // global limit must be explicitly enabled (zero means it is disabled)
+    if (cfg.limit == 0) revert DonLimitNotSet(donFamily);
```
infiloop2, line 1530: just checking, is this function called when a workflow is deleted as well? in that case we would want to allow that

infiloop2 (dismissed):
lgtm. added some small comments for improvements

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: Deividas said it's not needed in one of his previous comments so I removed it

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: No, it is called on upsert, activate, activateBatch and updateDONFamily

contracts/src/v0.8/workflow/dev/v2/WorkflowRegistry.sol
ibrajer: I've simplified it, please take a look

infiloop2 (approved):

DeividasK (approved):

PR #243, opened 2025-09-26
Add pagination to getters with multiple return items

DeividasK:


contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -565,16 +571,18 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
...
       if (nodeOperator.admin == address(0)) revert InvalidNodeOperatorAdmin();
-      uint32 nodeOperatorId = s_nextNodeOperatorId;
-      NodeOperator storage storedNodeOperator = s_nodeOperators[nodeOperatorId];
-      storedNodeOperator.admin = nodeOperator.admin;
-      storedNodeOperator.name = nodeOperator.name;
       bytes32 nodeOperatorDataHash = _nodeOperatorHash(nodeOperator);
       if (s_nodeOperatorDataHashToId[nodeOperatorDataHash] != 0) {
         revert NodeOperatorAlreadyExists(s_nodeOperatorDataHashToId[nodeOperatorDataHash]);
       }
+      uint32 nodeOperatorId = s_nextNodeOperatorId;
```
DeividasK, line 578: Moved data storage to happen after all checks.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -6,8 +6,8 @@ import {ITypeAndVersion} from "../../../shared/interfaces/ITypeAndVersion.sol";
 import {ICapabilityConfiguration} from "./interfaces/ICapabilityConfiguration.sol";
 import {INodeInfoProvider} from "./interfaces/INodeInfoProvider.sol";
 
-import {ERC165Checker} from "@openzeppelin/contracts@4.8.3/utils/introspection/ERC165Checker.sol";
-import {EnumerableSet} from "@openzeppelin/contracts@4.8.3/utils/structs/EnumerableSet.sol";
+import {ERC165Checker} from "@openzeppelin/contracts@5.1.0/utils/introspection/ERC165Checker.sol";
+import {EnumerableSet} from "@openzeppelin/contracts@5.1.0/utils/structs/EnumerableSet.sol";
```
DeividasK, line 10: Updated version to match what's used for the Workflow Registry.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -1174,25 +1197,24 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
...
-      }
+  /// @notice Returns a paginated list of configured DONs
+  /// @dev This operation will copy DONs to memory, which can be quite expensive.
+  /// This is designed to mostly be used by view accessors that are queried without any gas fees.
+  /// Does not revert on out-of-range pagination; it simply returns the largest sub-range that fits.
+  /// @param start Zero-based index at which the page begins
+  /// @param limit Maximum number of DONs to return
+  /// @return DONInfo[] List of DONs for the requested page
+  function getDONs(uint256 start, uint256 limit) external view returns (DONInfo[] memory) {
+    uint256 total = s_activeDONIds.length();
```
DeividasK, line 1208: Used this to avoid the sparse data problem. Otherwise, fetching all active DONs would have been particularly cumbersome (the total active value would have been unclear).

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -1052,28 +1068,27 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
...
-        mstore(nodeOperators, idx)
-      }
+  /// @notice Gets a paginated list of node operators. This operation will copy node operators to memory, which can be
+  /// quite expensive. This is designed to mostly be used by view accessors that are queried without any gas fees. Does
+  /// not revert on out-of-range pagination; it simply returns the largest sub-range that fits.
+  /// @param start Zero-based index at which the page begins
+  /// @param limit Maximum number of node operators to return
+  /// @return NodeOperatorInfo[] List of node operators for the requested page
+  function getNodeOperators(uint256 start, uint256 limit) external view returns (NodeOperatorInfo[] memory) {
+    uint256 total = s_activeNodeOperatorIds.length();
```
DeividasK, line 1078: Used this to avoid the sparse data problem. Otherwise, fetching all active NOPs would have been particularly cumbersome (the total active value would have been unclear).

ibrajer: So when the NOP is removed, it's not deleted from storage? Would it be more efficient to delete the NOP data from storage (even though deleting it can be expensive) so we don't have to track active ones in a separate data structure? The same question applies to the `getDONs` function.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -1109,14 +1124,22 @@ contract CapabilitiesRegistry is INodeInfoProvider, Ownable2StepMsgSender, IType
...
-    for (uint256 i; i < p2pIds.length; ++i) {
-      nodesInfo[i] = getNode(p2pIds[i]);
+  /// @notice Returns a paginated list of nodes
+  /// @dev This operation will copy nodes to memory, which can be quite expensive.
+  /// This is designed to mostly be used by view accessors that are queried without any gas fees.
+  /// Does not revert on out-of-range pagination; it simply returns the largest sub-range that fits.
+  /// @param start Zero-based index at which the page begins
+  /// @param limit Maximum number of nodes to return
+  /// @return NodeInfo[] List of nodes for the requested page
+  function getNodes(uint256 start, uint256 limit) external view returns (NodeInfo[] memory) {
```
ibrajer, line 1134: So NOPs and DONs can be activated/deactivated and we have to track them in a separate "active list", but this doesn't apply to individual nodes?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: The NOP is removed from storage, but the Solidity array length is not updated, e.g., if you delete NOPs with IDs 1 and 3 from [1, 2, 3, 4, 5], the length would still be 5. An enumerable set does update the length, though.

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
DeividasK: `s_nodeP2PIds` is already an active list. Basically, I made all list getters use `EnumerableSet.*` to become active lists.

ibrajer (approved):

justinkaseman (approved):

PR #256, opened 2025-10-02
[CRE-491] Move core relay evm types to chainlink-evm

pavel-raykov:
Supports https://github.com/smartcontractkit/chainlink/pull/19683

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.24.7
 
 require (
 	github.com/ethereum/go-ethereum v1.16.2
+	github.com/fxamacker/cbor/v2 v2.7.0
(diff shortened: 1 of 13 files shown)


jmank88 (approved):

RensR:
Can we please add codeowners for the `pkg` dir? Currently it uses the global fallback which isn't ideal.

pavel-raykov:
> Can we please add codeowners for the `pkg` dir? Currently it uses the global fallback which isn't ideal.

Created a separate PR for that https://github.com/smartcontractkit/chainlink-evm/pull/258 . Could you please have a look?

ettec (approved):

PR #258, opened 2025-10-02
[CRE-491] Set core team as the root owner of the pkg folder.

pavel-raykov:


jmank88 (dismissed):

.github/CODEOWNERS
```diff
@@ -12,6 +12,8 @@
 # Contracts catch all, for files not matched by the more specific patterns below
 /contracts/ @smartcontractkit/core-solidity
 
+/pkg/ @smartcontractkit/core
```
RensR: Should this not be foundations? Core is mostly a random bag of people who have been around for a long time. Maybe both Core and Foundations but Foundations feels like the most appropriate owner imo

pavel-raykov: changed to foundations.

jmank88 (approved):

RensR (approved):

PR #259, opened 2025-10-03
[CRE-491] Move remaining core type code to chainlink-evm

pavel-raykov:
Supports https://github.com/smartcontractkit/chainlink/pull/19714

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -1,0 +1,291 @@
+package config
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


pkg/config/types.go
jmank88, line 1: Are we able to use some of these types instead? https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/types/evm/chain_reader.go
I didn't finish this transition yet, but if we can use some instead of making another copy, that would be nice.

pavel-raykov: Reused `DataWordDetail`, other types are, unfortunately, different.

jmank88 (approved):

Atrax1 (approved):

PR #270, opened 2025-10-08
EVM key store on top of common key store lib

connorwstein:
Notes
- Adds helpers for creating EVM transaction keys and EVM OCR onchain keyrings on top of the new [base keystore library](https://github.com/smartcontractkit/chainlink-common/tree/main/keystore) Includes a test with real libocr to ensure the keys/keyrings function. Example usage:
```
	ks, err := keystore.LoadKeystore(ctx, storage, "test-password", commonks.WithScryptParams(commonks.FastScryptParams))
	require.NoError(t, err)
	testKey, err := evmks.CreateTxKey(ks, "test-tx-key")
	require.NoError(t, err)
	resp, err := testKey.SignTx(ctx, evmks.SignTxRequest{
		ChainID: big.NewInt(1337), // Use a test chain ID
		Tx:      testTransaction,
	})
	require.NoError(t, err)
	require.NoError(t, backend.Client().SendTransaction(ctx, resp.Tx))
``` 
- Points to latest chainlink-common which includes a golang version bump to 1.25, which required the v2 linter version. As per @chainchad's rec switched to using the common/standard one instead so we could remove the custom one here (but we can override it again if needed).


pkg/keystore/evm_tx.go
```diff
@@ -0,0 +1,140 @@
+package keystore
```
connorwstein: keysv2

.github/actions/golangci-lint/action.yml
```diff
@@ -65,11 +65,11 @@ runs:
         fi
 
     - name: Golangci-lint
-      uses: golangci/golangci-lint-action@38e1018663fa5173f3968ea0777460d3de38f256 # v5.3.0
+      uses: golangci/golangci-lint-action@v7.0.0
```
connorwstein: what happened here: 
1. chainlink-common recently bumped to golang 1.25, which bumped this repo to golang 1.25
2. the only version of the linter which supports 1.25 is golangci-lint v2.4.0 or greater. We are currently on v1.x
3. The v2 version of golangci-lint requires: >v7 of the golangci-lint action as well as breaking changes to the CLI arguments and the .golangci.yml itself.

I was mostly able to use the migration tool to move the .golangci.yml to the new version. Core is already on v2. We'd have to deal with it anyways at some point.

.golangci.yml
```diff
@@ -1,3 +1,4 @@
+version: "2"
```
connorwstein: massaging this into the v2 format

.github/actions/golangci-lint/action.yml
jmank88, line 1: WDYT about swapping over to the unified lint action?
```
        uses: smartcontractkit/.github/actions/ci-lint-go@bd2ca3d8fa2dc89f98b49b297e9b72c2e3e68cdc # v3.1.0
        with:
          golangci-lint-version: v2.6.1
```
cc @chainchad any reason no to use this everywhere?

.golangci.yml
jmank88: Did you run the command? Or convert manually?

pkg/keysv2/evm_tx.go
```diff
@@ -0,0 +1,154 @@
+// Package keysv2 provides key management functionality for EVM transactions and OCR2 keyrings.
+package keysv2
```
jmank88: Typically a v2 package is under a `<original>/v2` directory, but with the same name `package <original>`, so that importers reference it with the `/v2` suffix, but the import name doesn't change.
```suggestion
package keys
```
```diff
-import common/pkg/keys
+import common/pkg/keys/v2

keys.Thing
```

.golangci.yml
jmank88: ~@chainchad was this our original source for the shared action?~
nvm that was common IIRC

.github/actions/golangci-lint/action.yml
chainchad: No reason not to use the composite action to try and standardize.

If that action's [default config](https://github.com/smartcontractkit/.github/blob/main/actions/ci-lint-go/files/golangci-default.yml) doesn't work, we can pass another one in via the [golangci-lint-config]() input.

.golangci.yml
connorwstein: migrate command to start but iirc I think there were one or 2 things that didn't work out of the box after I had to tweak

pkg/keys/v2/evm_tx.go
```diff
@@ -0,0 +1,154 @@
...
+	gethtypes "github.com/ethereum/go-ethereum/core/types"
+	gethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/smartcontractkit/chainlink-common/keystore"
+)
+
+const (
+	// EVMPrefix is the prefix for EVM-related keys.
+	EVMPrefix = "evm"
+	// TxKeystorePrefix is the prefix for transaction keys.
+	TxKeystorePrefix = "tx"
```
jmank88: Will these be easier to find if they share a prefix instead of a suffix?
```suggestion
	PrefixEVM = "evm"
	// TxKeystorePrefix is the prefix for transaction keys.
	PrefixTxKeystore = "tx"
```

connorwstein: sure

.github/actions/golangci-lint/action.yml
connorwstein: update - was able to just move away to the default linter so didn't need this anymore

jmank88 (approved):

.github/workflows/golangci-lint.yml
```diff
@@ -26,10 +26,18 @@ jobs:
...
         with:
-          args: --config=${{ github.workspace }}/.golangci.yml
+          golangci-lint-version: v2.6.1
           only-new-issues: true
-          version: v1.64.8
-          working-directory: pkg
\ No newline at end of file
+          working-directory: pkg
+
+      # - name: Run golangci-lint
```
pavel-raykov, line 35: remove comments?

pkg/keys/v2/ocr2_keyring_test.go
```diff
@@ -0,0 +1,490 @@
...
+func (t *helper) Transmit(
+	ctx context.Context,
+	reportContext ocrtypes.ReportContext,
+	report ocrtypes.Report,
+	signatures []ocrtypes.AttributedOnchainSignature,
+) error {
+	t.backendMutex.Lock()
+	defer t.backendMutex.Unlock()
+
+	t.lggr.Info("Transmit", "report", report)
```
pavel-raykov, line 89: wouldn't you be able to reuse https://github.com/smartcontractkit/chainlink-evm/blob/8eb761afcb0f4b1cecaa76afe4bb70aa9cde33bf/pkg/transmitter/dual_contract_transmitter.go#L98 ?

dhaidashenko (approved):

pkg/keys/v2/ocr2_keyring_test.go
connorwstein: doesn't look like it - that struct has many other dependencies (logpoller etc.)

PR #274, opened 2025-10-15
OEV-605: Adds purging unstarted txs to txmv2

cl-efornaciari:
adds purging unstarted txs beyond the cutoff

### DIFF
--- a/pkg/txm/storage/inmemory_store.go
+++ b/pkg/txm/storage/inmemory_store.go
@@ -284,11 +284,10 @@ func (m *InMemoryStore) pruneUnstartedTransactionsWithinDuration(threshold time.
 		delete(m.Transactions, tx.ID)
 		m.UnstartedTransactions[idxTxToRetain] = nil // prevent memory leak
 	}
-	if len(txIDsToPrune) > 0 {
-		m.UnstartedTransactions = m.UnstartedTransactions[idxTxToRetain:]
-	} else {
+	if len(txIDsToPrune) == 0 {
 		return nil
 	}
+	m.UnstartedTransactions = m.UnstartedTransactions[idxTxToRetain:]
 	sort.Slice(txIDsToPrune, func(i, j int) bool { return txIDsToPrune[i] < txIDsToPrune[j] })
 	return txIDsToPrune
 }
--- a/pkg/txm/storage/inmemory_store.go
+++ b/pkg/txm/storage/inmemory_store.go
@@ -21,7 +21,7 @@ const (
 	// pruneSubset controls the subset of confirmed transactions to prune when the structure reaches its max limit.
 	// i.e. if the value is 3 and the limit is 90, 30 transactions will be pruned.
 	pruneSubset              = 3
-	pruneUnstartedTxDuration = 1 * time.Hour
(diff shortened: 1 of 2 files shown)


pkg/txm/storage/inmemory_store.go
```diff
@@ -277,6 +300,11 @@ func (m *InMemoryStore) UpdateUnstartedTransactionWithNonce(nonce uint64) (*type
 		m.lggr.Debugf("Unstarted transactions queue is empty for address: %v", m.address)
 		return nil, nil
 	}
+	prunedTxIDs := m.pruneUnstartedTransactionsWithinDuration(pruneUnstartedTxDuration)
+	if prunedTxIDs != nil {
```
dimriou: It's safer to check like this:
```suggestion
	if len(prunedTxIDs) != 0 {
```
because depending on the implementation, a slice could be `!= nil`, but still have 0 elements. `len` method is safe to use even for nil slices.

pkg/txm/storage/inmemory_store.go
```diff
@@ -269,6 +271,27 @@ func (m *InMemoryStore) UpdateTransactionBroadcast(txID uint64, txNonce uint64,
...
+		m.UnstartedTransactions[idxTxToRetain] = nil // prevent memory leak
+	}
+	if len(txIDsToPrune) == 0 {
+		return nil
+	}
+	m.UnstartedTransactions = m.UnstartedTransactions[idxTxToRetain:]
+	sort.Slice(txIDsToPrune, func(i, j int) bool { return txIDsToPrune[i] < txIDsToPrune[j] })
+	return txIDsToPrune
+}
+
```
dimriou, line 287: Does the work 👍 . Here's a more idiomatic way for future reference:
```suggestion
// Shouldn't call lock because it's being called by a method that already has the lock
func (m *InMemoryStore) pruneUnstartedTransactionsWithinDuration(threshold time.Duration) (txIDsToPrune []uint64) {
	for i, tx := range m.UnstartedTransactions {
		if time.Since(tx.CreatedAt) < threshold {
			m.UnstartedTransactions = m.UnstartedTransactions[i:]
			return txIDsToPrune // you can sort before this if you want to
		}
		txIDsToPrune = append(txIDsToPrune, tx.ID)
		delete(m.Transactions, tx.ID)
		m.UnstartedTransactions[i] = nil // prevent memory leak
	}
	m.UnstartedTransactions = m.UnstartedTransactions[:0]
	return
}
```

pkg/txm/storage/inmemory_store.go
```diff
@@ -277,6 +300,11 @@ func (m *InMemoryStore) UpdateUnstartedTransactionWithNonce(nonce uint64) (*type
 		m.lggr.Debugf("Unstarted transactions queue is empty for address: %v", m.address)
 		return nil, nil
 	}
+	prunedTxIDs := m.pruneUnstartedTransactionsWithinDuration(pruneUnstartedTxDuration)
```
dimriou, line 292: This will panic. Needs to be before the check on line 299.

dimriou (approved):

dhaidashenko (approved):

PR #276, opened 2025-10-16
Oev 618 update max retries logic

dimriou:
This PR implements two new changes:

1. Updates max retries logic: TXM no longer stops rebroadcasting new attempts for transactions that have reached the maximum threshold. Instead, it throws an error, purges old attempts, and retries. You should now rely exclusively on monitoring to catch abnormalities in broadcasting.
2. Updates time-based stuck transaction logic: along with the time-based heuristic, the time-based method also adds a max attempts threshold heuristic, even if the transaction has no successful broadcasts.

### DIFF
--- a/pkg/txm/storage/inmemory_store.go
+++ b/pkg/txm/storage/inmemory_store.go
@@ -16,6 +16,9 @@ import (
 )
 
 const (
+	// MaxAllowedAttempts controls the maximum number of attempts stored per transactions. After this threshold is exceeded
+	// the oldest attempts will be pruned to make room for new attempts.
+	MaxAllowedAttempts = 10
 	// maxQueuedTransactions is the max limit of UnstartedTransactions and ConfirmedTransactions structures.
 	maxQueuedTransactions = 250
 	// pruneSubset controls the subset of confirmed transactions to prune when the structure reaches its max limit.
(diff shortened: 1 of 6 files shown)


jmank88 (approved):

pkg/txm/storage/inmemory_store.go
```diff
@@ -85,8 +88,13 @@ func (m *InMemoryStore) AppendAttemptToTransaction(txNonce uint64, attempt *type
 	}
 
 	attempt.CreatedAt = time.Now()
-	attempt.ID = uint64(len(tx.Attempts)) // Attempts are not collectively tracked by the in-memory store so attemptIDs are not unique between transactions and can be reused.
+	attempt.ID = uint64(tx.AttemptCount) // Attempts are not collectively tracked by the in-memory store so attemptIDs are not unique between transactions and can be reused.
```
cl-efornaciari, line 91: why did this change?

pkg/txm/storage/inmemory_store.go
```diff
@@ -85,8 +88,13 @@ func (m *InMemoryStore) AppendAttemptToTransaction(txNonce uint64, attempt *type
...
 	tx.AttemptCount++
+	// Prune oldest attempt.
+	if len(tx.Attempts) >= MaxAllowedAttempts {
```
cl-efornaciari, line 94: because of the lock I think its not possible for this to exceed at most 1 attempts over the max allowed attempts, is that correct?

pkg/txm/storage/inmemory_store_test.go
```diff
@@ -104,6 +104,27 @@ func TestAppendAttemptToTransaction(t *testing.T) {
...
+
+		var attemptsTried uint16 = 12
+		for range attemptsTried {
+			newAttempt := &types.Attempt{
+				TxID: 1,
+			}
+			require.NoError(t, m2.AppendAttemptToTransaction(nonce, newAttempt))
+		}
+		tx, _ := m2.FetchUnconfirmedTransactionAtNonceWithCount(nonce)
+		fmt.Println("TX:", tx)
```
cl-efornaciari, line 123: nit: do we want this print here?

pkg/txm/storage/inmemory_store_test.go
```diff
@@ -104,6 +104,27 @@ func TestAppendAttemptToTransaction(t *testing.T) {
...
+		assert.Len(t, tx.Attempts, MaxAllowedAttempts)
+		assert.Equal(t, attemptsTried, tx.AttemptCount)
+		assert.Equal(t, uint64(attemptsTried-MaxAllowedAttempts), tx.Attempts[0].ID)
```
cl-efornaciari, line 126: nice

cl-efornaciari (approved):
LGTM, added a few comments/questions

pkg/txm/storage/inmemory_store.go
dimriou: After this patch, the length of the attempts slice is finite, but the total number of attempts can exceed that threshold due to pruning. To accurately count them we need to make this switch.

pkg/txm/storage/inmemory_store.go
dimriou: Indeed, that's just safer.

PR #280, opened 2025-10-17
PLEX-1843: add node balances to beholder

fernandezlautaro:
jira: https://smartcontract-it.atlassian.net/browse/PLEX-1843

Need node balance to be emitted in beholder besides prometheus for CREs' dashboards

requires: 
- https://github.com/smartcontractkit/chainlink-framework/pull/71

### DIFF
--- a/pkg/monitor/balance.go
+++ b/pkg/monitor/balance.go
@@ -94,8 +94,8 @@ func (bm *balanceMonitor) OnNewLongestChain(_ context.Context, _ *evmtypes.Head)
 	}
 }
 
-func (bm *balanceMonitor) updateBalance(ethBal assets.Eth, address common.Address) {
-	bm.promUpdateEthBalance(&ethBal, address)
+func (bm *balanceMonitor) updateBalance(ctx context.Context, ethBal assets.Eth, address common.Address) {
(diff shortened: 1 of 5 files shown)


pkg/monitor/balance.go
```diff
@@ -134,7 +141,7 @@ func (bm *balanceMonitor) promUpdateEthBalance(balance *assets.Eth, from common.
 		return
 	}
 
-	metrics.NodeBalance.WithLabelValues(from.Hex(), bm.chainIDStr, metrics.EVM).Set(balanceFloat)
+	bm.balanceMetrics.RecordNodeBalance(context.Background(), from.Hex(), balanceFloat)
```
dhaidashenko: we should not pass the background context as it won't be closed during the shutdown, so this call may block past the close timeout and break and cause "abnormal termination" that could be misleading.

dhaidashenko: nit: method no longer updates just prom metrics, so should be renamed

dhaidashenko (dismissed):

ilija42 (approved):

dhaidashenko (approved):

PR #281, opened 2025-10-20
Enabled safe tags in rpc client

silaslenihan:


pkg/client/rpc_client.go
```diff
@@ -570,9 +573,17 @@ func (r *RPCClient) HeaderByHash(ctx context.Context, hash common.Hash) (header
...
-	if err != nil {
-		return
+
+	if r.finalityTagEnabled && r.safeTagEnabled {
+		err = r.ethGetBlockByNumber(ctx, rpc.SafeBlockNumber.String(), &head)
+		if err != nil {
+			return
+		}
+	} else {
+		head, err = r.latestFinalizedBlock(ctx)
```
silaslenihan: should this return latest instead? Or are we only expecting this to be called on a chain with finality tags enabled

pkg/client/rpc_client.go
```diff
@@ -570,9 +573,17 @@ func (r *RPCClient) HeaderByHash(ctx context.Context, hash common.Hash) (header
 func (r *RPCClient) LatestSafeBlock(ctx context.Context) (head *evmtypes.Head, err error) {
 	ctx, cancel, _, _, _ := r.acquireQueryCtx(ctx, r.rpcTimeout)
 	defer cancel()
-	err = r.ethGetBlockByNumber(ctx, rpc.SafeBlockNumber.String(), &head)
-	if err != nil {
-		return
+
+	if r.finalityTagEnabled && r.safeTagEnabled {
```
dhaidashenko: IMO, this log should be done [here](https://github.com/smartcontractkit/chainlink-framework/blob/2350c82883e20adf2d0ae5ce482a90993b3e6b74/chains/heads/tracker.go#L404).
RPC should always return what was requested, even if chain does not support that.

CONFIG.md
```diff
@@ -23,6 +23,7 @@ ChainType = 'arbitrum' # Example
 SafeDepth = 0 # Default
 FinalityDepth = 50 # Default
 FinalityTagEnabled = false # Default
+SafeTagEnabled = true # Default
```
prashantkumar1982: nit: Maybe a better name would be SupportsSafeTag or SafeTagSupported, or SafeTagSupportedByChain.
Typically Enabled means if we have enabled this specific config for this chain.
For example, when we say "FinalityTagEnabled", we imply whether we use FinalityTags or not. The chain may suport FinalityTags but we still sometimes explicitly not use it.

For SafeTag, we want to be clear here that this field indicates if the chain supports it.

pkg/client/rpc_client.go
silaslenihan: done

dhaidashenko (dismissed):

prashantkumar1982 (dismissed):

dhaidashenko (approved):

ilija42 (approved):

PR #288, opened 2025-10-27
Add ErrorHandler

dimriou:
This PR adds an ErrorHandler implementation that allows the TXM to revert to the previous nonce if the preliminary check on the MetaClient fails. This avoids excessive empty transactions.
Empty transactions are still used as a fallback mechanism in case the transaction data is not valid, i.e., extreme gas limit values, so the malformed transaction can be purged.

### DIFF
--- a/pkg/txm/clientwrappers/dualbroadcast/error_handler.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/error_handler.go
@@ -20,12 +20,13 @@ func NewErrorHandler() *errorHandler {
 func (e *errorHandler) HandleError(
 	ctx context.Context,
 	tx *types.Transaction,
-	err error,
+	txErr error,
 	txStore txm.TxStore,
 	setNonce func(common.Address, uint64),
 	isFromBroadcastMethod bool,
 ) error {
-	if strings.Contains(err.Error(), NoBidsError) {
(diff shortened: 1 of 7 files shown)


pkg/txm/clientwrappers/dualbroadcast/error_handler.go
```diff
@@ -0,0 +1,37 @@
...
+	if strings.Contains(err.Error(), NoBidsError) {
+		if err := txStore.MarkTxFatal(ctx, tx, tx.FromAddress); err != nil {
+			return err
+		}
+		setNonce(tx.FromAddress, *tx.Nonce)
+		return fmt.Errorf("transaction with txID: %d marked as fatal", tx.ID)
+	}
+
+	return nil
+}
```
cl-efornaciari, line 38: when in the callchain can this get invoked??

cl-efornaciari (commented):
Looks good, but I am lacking context on what this can impact inside and outside the SVR perspective. We may want to get additional reviews

pkg/txm/clientwrappers/dualbroadcast/error_handler.go
dimriou: Right after we broadcast the attempt on the TXM: `t.errorHandler.HandleError( ...`

PR #292, opened 2025-11-03
Extract CRE into its own package

RensR:
This PR extracts Workflow & Keystone from the core contracts folder. 

- Separate NPM package
- Separate gobindings go.mod
- Shared CI to maintain consistency

Existing gethwrappers remain unchanged. This ensures zero impact on offchain systems. That is also the source of the massive diff: duplicated gethwrappers

TODO:
- future PR 1: replace gethwrapper imports with the new ones
- future PR 2: introduce improvements to versioning of wrappers & add bytecode+abi versioned to the npm pacakge.


DeividasK (dismissed):

gethwrappers/generation/wrap.go
```diff
@@ -25,6 +25,11 @@ func main() {
 		rootDir = "../contracts/solc/"
 	}
 
+	// If the project starts with "workflow" we need to adjust the path
+	if len(project) >= 8 && project[0:8] == "workflow" || project == "keystone" {
+		rootDir = "../../solc/"
+	}
```
pavel-raykov: why do you need this part? In this PR there is also cre/gobinginds/generates/wrap.go that can do the same, right ?

Makefile
```diff
@@ -21,7 +21,7 @@ protoc: ## Install protoc
 
 .PHONY: generate
 generate: gomods codecgen mockery protoc modgraph
-	export PATH="$(HOME)/.local/bin:$(PATH)"; gomods -s gethwrappers -go generate ./...
+	export PATH="$(HOME)/.local/bin:$(PATH)"; gomods -s gethwrappers,contracts/cre/ -go generate ./...
```
pavel-raykov, line 24: is the intention that all gethwrappers will be moved under contracts/${product-name} at some point?

gethwrappers/generation/wrap.go
RensR: Ahh yeah good catch, I generated the wrappers from this file originally but that didn't work once I introduced the new go.mod. Fixing now

Makefile
RensR: My intention, yes. I just have to convince everyone the churn is worth it. For now it's just CRE. If this is a success we can start thinking about more extractions.

DeividasK (approved):

contracts/cre/gobindings/generate/wrap.go
```diff
@@ -0,0 +1,105 @@
...
+
+	gethwrappers "github.com/smartcontractkit/chainlink-evm/contracts/cre/gobindings"
+	zksyncwrapper "github.com/smartcontractkit/chainlink-evm/contracts/cre/gobindings/zksync"
+)
+
+var (
+	rootDir = "../../solc/"
+)
+
+func main() {
```
pavel-raykov, line 16: is it also necessary to copy this file?

This logic should be shared among the contracts. Do you want each contracts/{product} to have their own copy?

pavel-raykov (approved):

contracts/cre/gobindings/generate/wrap.go
RensR: Yes, potential follow up but not blocking this

mohamed-mehany (approved):
LGTM.
Assuming no bytecode change for the DF contract so no audit/redeployment required.

matYang (approved):

PR #294, opened 2025-11-07
Oev 671 txm improvements

dimriou:
Each commit corresponds to the following features:
- OEV-676 Sync broadcasting and backfilling
- OEV-680 Trigger instant retransmission of purgeable transactions
- OEV-675 Reduce empty transaction gas limit
- OEV-677 Add Meta Error Handler
- OEV-668 Add agnostic gas bumping
- OEV-679 Add separate pricing for empty transactions

Many of the TODOs of this PR are fixed in:
- https://github.com/smartcontractkit/chainlink-evm/pull/303
Specifically for bumping Purgeable attempts, you can see the updated method in the above PR.

Update: `GetMaxFee` PR merged into this


pkg/txm/txm.go
```diff
@@ -132,9 +132,8 @@ func (t *Txm) startAddress(address common.Address) {
 	triggerCh := make(chan struct{}, 1)
 	t.triggerCh[address] = triggerCh
 
-	t.wg.Add(2)
-	go t.broadcastLoop(address, triggerCh)
-	go t.backfillLoop(address)
+	t.wg.Add(1)
+	go t.loop(address, triggerCh)
```
augustbleeds, line 137: can you remind me what the advantage was of combining these two loops?

pkg/txm/txm.go
```diff
@@ -387,31 +367,31 @@ func (t *Txm) backfillTransactions(ctx context.Context, address common.Address)
...
-		return false, err
+		return err
 	}
 	if unconfirmedCount == 0 {
 		t.lggr.Debugf("All transactions confirmed for address: %v", address)
-		return false, err // TODO: add backoff to optimize requests
+		return err
 	}
 
 	if tx == nil || *tx.Nonce != latestNonce {
```
augustbleeds, line 378: when does this scenario happen? hard for me to think of why the local nonce would be behind the chain nonce if it has to originate from the local txm.

pkg/txm/attempt_builder.go
```diff
@@ -67,15 +67,30 @@ func (a *attemptBuilder) NewAgnosticBumpAttempt(ctx context.Context, lggr logger
...
-			lggr.Errorf("error bumping attempt: %v for txID: %v", err, tx.ID)
-			return attempt, nil
+
+	// bump purge attempts
+	if tx.IsPurgeable {
+		for {
+			// TODO: add better handling
+			bumpedAttempt, err := a.NewBumpAttempt(ctx, lggr, tx, *attempt)
+			if err != nil {
+				return attempt, nil
```
augustbleeds: can we select for that specific error so we reduce possibility of halting on a different error?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -27,8 +27,11 @@ import (
 )
 
 const (
-	timeout = time.Second * 5
-	metaABI = `[
+	timeout                    = time.Second * 5
+	NoBidsError                = "no bids"
+	NoSolverOps                = "no solver operations received"
```
augustbleeds, line 31: ~can you remind me what this error means?~

pkg/txm/txm.go
dimriou: It enables easier nonce management via error handling when broadcasting a transaction.

pkg/txm/txm.go
dimriou: Not behind but in front. Here's an example: you broadcast tx A to the mempool. You restart the TXM, fetch the pending nonce, which is A+1, and start sending more transactions to the mempool. Since you restarted the TXM, transaction A is not being tracked. If it gets dropped by the RPC or the builders, TXM won't be able to recover unless it fills the nonce. That's what this case does.

augustbleeds (commented):
Changes look good to me, just want to clarify the error handling behavior for empty tx gas bumping.

pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+type errorHandler struct{}
+
+func NewErrorHandler() *errorHandler {
+	return &errorHandler{}
+}
+
+func (e *errorHandler) HandleError(ctx context.Context, tx *types.Transaction, txErr error, txStore txm.TxStore, setNonce func(common.Address, uint64), isFromBroadcastMethod bool) error {
+	// If this isn't the first broadcast, don't mark the tx as fatal as other txs might be included on-chain.
```
cll-gg:
```suggestion
	// Only if this is the first broadcast, mark the tx as fatal. In other cases, other txs might be included on-chain 
```

pkg/txm/attempt_builder.go
```diff
@@ -42,13 +53,52 @@ func (a *attemptBuilder) NewAttempt(ctx context.Context, lggr logger.Logger, tx
...
 
+func (a *attemptBuilder) NewAgnosticBumpAttempt(ctx context.Context, lggr logger.Logger, tx *types.Transaction, dynamic bool) (*types.Attempt, error) {
+	attempt, err := a.NewAttempt(ctx, lggr, tx, dynamic)
+	if err != nil {
+		return nil, err
+	}
+
+	// bump purge attempts
+	if tx.IsPurgeable {
+		// TODO: add better handling
```
cll-gg: Would be good to make a separate ticket with an explanation of what would be better here

pkg/txm/attempt_builder.go
```diff
@@ -42,13 +53,52 @@ func (a *attemptBuilder) NewAttempt(ctx context.Context, lggr logger.Logger, tx
...
+		for {
+			bumpedAttempt, err := a.NewBumpAttempt(ctx, lggr, tx, *attempt)
+			if err != nil {
+				if errors.Is(err, fees.ErrConnectivity) {
```
lhmerino: I'm looking at the error choice. I believe this is due to the code located [here](https://github.com/smartcontractkit/chainlink-evm/blob/da6e4a9838e94e56df48a3dd2daf8429a698742e/pkg/gas/fee_history_estimator.go#L388C2-L392C1)

```
if bumpedMaxPriorityFeePerGas.Cmp(priorityFeeThreshold) > 0 {
    return ..., ErrConnectivity
}
```

It feels slightly counterintuitive to return ErrConnectivity for a value check, where ErrBumpFeeExceedsLimit might seem more natural.

Is the intention to treat priorityFeeThreshold as a sanity guardrail (implying the network/estimator is broken if fees get this high) rather than a budget limit (implying the tx just needs to pay more)? If so, adding a small comment explaining why this triggers a 'connectivity' error would be helpful for readability.

pkg/txm/txm.go
```diff
@@ -216,18 +216,19 @@ func newBackoff(minDuration time.Duration) backoff.Backoff {
...
 }
 
-func (t *Txm) broadcastLoop(address common.Address, triggerCh chan struct{}) {
+func (t *Txm) loop(address common.Address, triggerCh chan struct{}) {
 	defer t.wg.Done()
 	ctx, cancel := t.stopCh.NewCtx()
 	defer cancel()
 	broadcastWithBackoff := newBackoff(1 * time.Second)
 	var broadcastCh <-chan time.Time
+	backfillCh := time.After(utils.WithJitter(t.config.BlockTime))
```
lhmerino: `utils.WithJitter` is a [deprecated](https://github.com/smartcontractkit/chainlink-common/blob/731e805e0a2fc5351705f0c24a47c8a8719d39bd/pkg/utils/utils.go#L15) function.
It says to use `timeutils` package instead.

pkg/txm/txm.go
lhmerino: I noticed TXM uses `pending` for the block parameter to capture pending transactions.
Does the Flashbots client behave accordingly (i.e., takes into account inflight transactions)?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -142,7 +145,7 @@ func NewMetaClient(lggr logger.Logger, c MetaClientRPC, ks MetaClientKeystore, c
 	}
```
lhmerino, line 146: Is the MetaClient currently active -- it looks A-specific? 
I noticed a potential dependency in the metrics setup. If beholder is unavailable, will this cause the client to error out as well?

Ideally, we want to ensure the client is decoupled so that a beholder outage doesn't block the main flow. For example, on 10/10 I saw many alerts about beholder potentially having issues.

pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -12,12 +12,16 @@ import (
...
 
-func SelectClient(lggr logger.Logger, client client.Client, keyStore keys.ChainStore, url *url.URL, chainID *big.Int) (txm.Client, error) {
+func SelectClient(lggr logger.Logger, client client.Client, keyStore keys.ChainStore, url *url.URL, chainID *big.Int) (txm.Client, txm.ErrorHandler, error) {
 	urlString := url.String()
 	switch {
 	case strings.Contains(urlString, "flashbots"):
-		return NewFlashbotsClient(client, keyStore, url), nil
+		return NewFlashbotsClient(client, keyStore, url), nil, nil
 	default:
-		return NewMetaClient(lggr, client, keyStore, url, chainID)
```
lhmerino, line 21: This code could then be made cleaner if `NewMetaClient` does not fail.

pkg/txm/storage/inmemory_store.go
```diff
@@ -366,8 +366,15 @@ func (m *InMemoryStore) DeleteAttemptForUnconfirmedTx(transactionNonce uint64, a
...
 
-func (m *InMemoryStore) MarkTxFatal(*types.Transaction) error {
-	return errors.New("not implemented")
+func (m *InMemoryStore) MarkTxFatal(txToMark *types.Transaction) error {
+	m.Lock()
+	defer m.Unlock()
+
+	// TODO: for now do the simple thing and drop the transaction instead of adding it to the fatal queue.
+	delete(m.UnconfirmedTransactions, *txToMark.Nonce)
+	delete(m.Transactions, txToMark.ID)
```
lhmerino, line 375: I have a few safety concerns about deleting these entries:
- Nonce Gaps: Could deleting this nonce create a gap in our internal inflight list? I'm concerned that higher-nonce transactions currently inflight might stall indefinitely waiting for the tx associated with this nonce to fill.
- Panic Risk: Dereferencing *txToMark.Nonce without checking if txToMark.Nonce is not nil can cause a panic.
- Less important but still relevant: could this complicate debugging by deleting this information?

pkg/txm/storage/inmemory_store.go
```diff
@@ -122,6 +121,7 @@ func (m *InMemoryStore) CreateEmptyUnconfirmedTransaction(nonce uint64, gasLimit
 		SpecifiedGasLimit: gasLimit,
```
lhmerino, line 121: Suggestion: We could simplify this by hardcoding the gas limit to 21_000.

Since a standard ETH transfer to an EOA (0x0) has a fixed cost (21_000), passing in gasLimit and maintaining a separate EmptyTxLimitDefault config seems like unnecessary overhead. In addition, it would also remove the need to add `gasEstimatorConfig` parameter to `func NewTxmV2`.

If we were to instead send ETH to/from a smart contract, only then would the gas amount differ.

pkg/txm/txm.go
dimriou: Yes it does.

pkg/txm/attempt_builder.go
dimriou: This functionality is based on TXMv1. Both [FeeHistory](https://github.com/smartcontractkit/chainlink-evm/blob/da6e4a9838e94e56df48a3dd2daf8429a698742e/pkg/gas/fee_history_estimator.go#L388C2-L392C1) and [BlockHistory](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/gas/block_history_estimator.go#L306) estimators use the 90th percentile as a maximum threshold to prevent indefinite gas bumping.
The above was used in a hacky way to access that percentile without changing multiple interfaces in the gas package during testing. We'll have to update the PR and change this before it gets merged.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Both prom and beholder packages are safe and if they're not available a default object will be created with NOOPs.

dimriou: The errors you saw are due to Beholder package not properly exlcuding the bootstrap node from the metrics and sending repeated error messages. This is Beholder-specific.

pkg/txm/storage/inmemory_store.go
dimriou: - Nonce Gaps: it's in the TXM's responsibility to call `MarkTxFatal` in the correct place so it doesn't create a nonce gap, the same way it does for other methods as well. In this case, we call `MarkTxFata` inside the Meta error handler and only if there are no past broadcasts. Finally, if all the above measures for some reason fail, the TXM has a native nonce gap detection mechanism that will fill any untracked nonce gaps and unblock txs with higher nonces.
- Panic Risk: we check for a nil nonce value a few lines before [here](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/txm/txm.go#L343-L345).
- Debugging: the fatal transactions are not currently being utilized in some sort of way. We do log the information of this marking [here](https://github.com/smartcontractkit/chainlink-evm/blob/oev-671_txm_improvements/pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go#L27), but for now there is no effort to do something with the fatal queue, hence the TODO.

pkg/txm/storage/inmemory_store.go
dimriou: Although standard ETH transfer uses 21k we've observed other EVM-based chains using slightly different values. Since there is already a Core config that we can leverage, it's better to future-proof the design than to require a new custom release if something needs to be adjusted.

pkg/txm/attempt_builder.go
dimriou: PR that permanently handles this is here:
- https://github.com/smartcontractkit/chainlink-evm/pull/303

pkg/chains/legacyevm/evm_txm.go
```diff
@@ -60,6 +60,7 @@ func newEvmTxm(
 				logPoller,
 				opts.KeyStore,
 				estimator,
+				cfg.GasEstimator(),
```
pavel-raykov, line 63: why do you need to pass the GasEstimator separately, isn't it already passed in cfg?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -27,8 +27,11 @@ import (
 )
 
 const (
-	timeout = time.Second * 5
-	metaABI = `[
+	timeout                    = time.Second * 5
+	NoBidsError                = "no bids"
```
pavel-raykov: why not `errors.New("no bids")` ? this way you can use `errors.Is` for easy error testing.

pkg/txm/txm.go
```diff
@@ -361,77 +342,77 @@ func (t *Txm) sendTransactionWithError(ctx context.Context, tx *types.Transactio
...
 				tx.PrintWithAttempts())
 		} else {
-			t.metrics.ReachedMaxAttempts(ctx, false)
+			t.Metrics.ReachedMaxAttempts(ctx, false)
 		}
 
-		if tx.LastBroadcastAt == nil || time.Since(*tx.LastBroadcastAt) > (t.config.BlockTime*time.Duration(t.config.RetryBlockThreshold)) {
+		if tx.LastBroadcastAt == nil || time.Since(*tx.LastBroadcastAt) > (t.config.BlockTime*time.Duration(t.config.RetryBlockThreshold)) || tx.IsPurgeable {
 			// TODO: add optional graceful bumping strategy
 			t.lggr.Info("Rebroadcasting attempt for txID: ", tx.ID)
```
pavel-raykov, line 414: you can add more information here to show why rebroadcasting happened: it seems that there are 3 conditions above and it would be useful to understand which triggered.

pkg/chains/legacyevm/evm_txm.go
dimriou: The entire cfg is a superset of all configs, but the expected param interfaces are a small subset of it. Historically, we did that to split the responsibilities of configs and not pass one large EVM config, so I replicated the same behaviour here, but if you ask me it's not a great design choice.

pavel-raykov: Thanks for explaining, let's keep it as is.

pavel-raykov (approved):

amit-momin (approved):

PR #297, opened 2025-11-13
pkg/log: set FromBlock on subscribe

jmank88:


simsonraj (approved):

jmank88:
@simsonraj Can we close this as unnecessary?

PR #300, opened 2025-11-16
Add llo/ChannelConfigStore support for adder roles and multiple channel definition sources

kaleofduty:
The most interesting part is how channel adder authorization works in two steps, setting the allowed channelAdders for a DON, and then the authorized address for a given channelAdder.

allowed channelAdders for a DON:
- Each DON maintains a set of allowed channelAdderIds
- Set by owner via setChannelAdder(donId, channelAdderId, bool)

authorized address for a given channelAdder:
- Each channelAdderId (uint32) maps to exactly one address
- Set by owner via setChannelAdderAddress(channelAdderId, address)


contracts/src/v0.8/llo-feeds/v0.5.0/configuration/ChannelConfigStore.sol
```diff
@@ -2,24 +2,121 @@
...
+  }
+
+  /// @notice Sets the address for a channel adder ID
+  /// @param channelAdderId The channel adder ID
+  /// @param adderAddress The address to associate with the channel adder ID
+  function setChannelAdderAddress(
+    uint32 channelAdderId,
+    address adderAddress
+  ) external onlyOwner {
+    s_channelAdderAddresses[channelAdderId] = adderAddress;
```
akuzni2: Where I see this being integrated in Core when logs are parsed we are using the `channelAdderId` and comparing in someplaces with an expected "OwnerId". I see that OwnerId isn't defined anywhere in this contract. Likely we could fix this at the integration itself to make it clearer - but if that's the path we're taking maybe we should define an OwnerId to prevent any integration issues.

contracts/src/v0.8/llo-feeds/v0.5.0/configuration/ChannelConfigStore.sol
```diff
@@ -2,24 +2,121 @@
...
   }
 
+  /// @notice Allows a channel adder to add channel definitions to the specified DON.
+  /// The DON enforces (in its consensus rules), that the channel definitions provided
+  /// by the channel adder are well-formed, purely additive, and do not overlaod the DON.
+  /// @param donId The DON ID
+  /// @param channelAdderId The channel adder ID
+  /// @param url The URL of the channel definition
+  /// @param sha The SHA hash of the channel definition
+  function addChannelDefinitions(
```
akuzni2, line 63: Should this function check that a `s_channelDefinitionVersions` exists for particular `donId` i.e. we've called `setChannelDefinitions` at least once?

brunotm: to prevent an already setup don to have configured adders added before the owner? while i agree that the order would be correct we can choose to enable a given don to have only adders if necessary.

kaleofduty: Note also that a channel adder cannot perform any actions for a `donId` before the owner has executed `setChannelAdder(donId, channelAdderId, true)`.

contracts/src/v0.8/llo-feeds/v0.5.0/configuration/ChannelConfigStore.sol
```diff
@@ -1,27 +1,124 @@
...
 
   /// @notice The version of a channel definition keyed by DON ID
   // Increments by 1 on every update
   mapping(uint256 => uint256) internal s_channelDefinitionVersions;
 
+  /// @notice Mapping from channel adder ID to its corresponding address
+  mapping(ChannelAdderId => address) internal s_channelAdderAddresses;
+
+  /// @notice Mapping from DON ID to the set of allowed channel adder IDs
+  mapping(uint256 => EnumerableSet.UintSet) internal s_allowedChannelAdders;
```
RensR, line 42: nit: named map

```
 mapping(uint256 donId => EnumerableSet.UintSet allowedChannels)
```

same for the other maps

brunotm (approved):

brunotm (approved):

ro-tex (approved):

PR #302, opened 2025-11-26
OEV-699 Introduce relayConfig.GasLimit to override gas limit on job level

cll-gg:
OEV-699: Make it possible to set the gas limit per feed. 
We want to do this for SVR and L2EP feeds specifically, so that they don't need to conform to the global gas limit. 

There's support on the transmitter side for setting a custom gas limit [here](https://github.com/smartcontractkit/chainlink-evm/blob/da6e4a9838e94e56df48a3dd2daf8429a698742e/pkg/transmitter/util.go#L142-L155). 
This is used for e.g. Automation to set gas limit per Product. 

With this change, we can now also set gas limit per job. 

### Supports
https://github.com/smartcontractkit/chainlink/pull/20386

### Notes
* There's already support for setting a gas limit per job, which was introduced in https://github.com/smartcontractkit/chainlink/pull/6935, but this is not used at the moment.


pkg/transmitter/util.go
```diff
@@ -186,3 +178,19 @@ func generateTransmitterFrom(ctx context.Context, rargs types.RelayArgs, ethKeys
 	}
 	return transmitter, nil
 }
+
+func getGasLimitFrom(chain legacyevm.Chain, opts ConfigTransmitterOpts, relayConfig config.RelayConfig) uint64 {
```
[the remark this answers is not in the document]
cll-gg: The test is in the same package

pkg/transmitter/util.go
```diff
@@ -186,3 +178,19 @@ func generateTransmitterFrom(ctx context.Context, rargs types.RelayArgs, ethKeys
...
+	gasLimit := chain.Config().EVM().GasEstimator().LimitDefault()
+	ocr2Limit := chain.Config().EVM().GasEstimator().LimitJobType().OCR2()
+	if ocr2Limit != nil {
+		gasLimit = uint64(*ocr2Limit)
+	}
+	if opts.PluginGasLimit != nil {
+		gasLimit = uint64(*opts.PluginGasLimit)
+	}
+	if relayConfig.GasLimit != nil {
```
pavel-raykov: what is the reason to choose relayConfig.GasLimit to have the highest precedence? It seems to be the most general one out of the three (ocr2Limit and opts.PluginGasLimit), so it seems more natural to put it first.

cll-gg: If I'm not mistaken, `ocr2Limit`  is set in the node toml config, so it's less specific than `relayConfig.GasLimit`, which originates from a job spec. 
As for `opts.PluginGasLimit`, it might indeed be better to give that higher precedence, it seems conceptually more specific than `relayConfig.GasLimit` indeed.

pkg/transmitter/util.go
```diff
@@ -134,15 +134,7 @@ func generateTransmitterFrom(ctx context.Context, rargs types.RelayArgs, ethKeys
...
-	gasLimit := chain.Config().EVM().GasEstimator().LimitDefault()
-	ocr2Limit := chain.Config().EVM().GasEstimator().LimitJobType().OCR2()
-	if ocr2Limit != nil {
-		gasLimit = uint64(*ocr2Limit)
-	}
-	if opts.PluginGasLimit != nil {
-		gasLimit = uint64(*opts.PluginGasLimit)
-	}
-
+	gasLimit := getGasLimitFrom(chain, opts, relayConfig)
```
jmank88: You can use `cmp.Or` to select the first non-nil value, and then the func maybe isn't necessary:
```suggestion
	gasLimit := chain.Config().EVM().GasEstimator().LimitDefault()
	ocr2Limit := chain.Config().EVM().GasEstimator().LimitJobType().OCR2()
	if gl := cmp.Or(ocr2Limit, opts.PluginGasLimit, relayConfig.GasLimit); gl != nil {
		gasLimit = uint64(*gl)
	}
```

cll-gg: Nice idea! A downside is that the func helps with testability. 
I'm open to both, any preference?

pkg/transmitter/util.go
jmank88: I think I agree on the precedence, but then let's consider if the callers that currently pass a value via `opts` still make sense.

cll-gg: The only caller that currently passes a value via `opts` is Automation/Keepers [here](https://github.com/smartcontractkit/chainlink/blob/ce4687bdf727b84ddbb139be54bd39b1eb0feae0/core/services/relay/evm/ocr2keeper.go#L90-L91). 

I assume this still makes sense to keep. 

So then I propose I'll change the precedence to be from low to high: `ocr2limit` -> `relayConfig.GasLimit` -> `opts.PluginGasLimit`. 

Let me know if any of you disagrees!

cll-gg: This has been updated

pavel-raykov: You can still use `cmp.Or` inside this function (if you want to keep it for testing)

cll-gg: Good point, done

pkg/transmitter/util_test.go
```diff
@@ -0,0 +1,136 @@
...
+			})
+
+			if !assert.Equal(t, tt.expectedGasLimit, gasLimit, tt.description) {
+				t.Errorf("expected gas limit: %d, got: %d", tt.expectedGasLimit, gasLimit) // to print in decimal format
+			}
+		})
+	}
+}
+
+type mockJobTypeConfig struct {
```
pavel-raykov: could you please add LimitJobType to the .mockery file and use the standard mock instead of writing a custom one?

cll-gg: Done!

pkg/transmitter/util.go
jmank88: If you want to keep it, then I think it should be simplified. e.g. it doesn't need the whole `legacyevm.Chain` - it just needs the gas estimator interface.

jmank88: Don't need the whole RelayConfig either

cll-gg: Updated!

jmank88 (approved):

pavel-raykov (approved):

PR #310, opened 2025-12-09
Clean up gas estimator after GetMaxDynamicFee method was added

dimriou:
For SVR purposes we had to add a `GetMaxFee` method to the general gas estimator interface. Due to the strict deadline we implemented the underlying `GetMaxDynamicFee` method but not the `GetMaxLegacyGas` where applicable. This PR implements that method, adds some tests for `GetMaxDynamicFee` and does a small clean up regarding configs and deprecated methods.

### DIFF
--- a/pkg/gas/arbitrum_estimator.go
+++ b/pkg/gas/arbitrum_estimator.go
@@ -143,6 +143,12 @@ func (a *arbitrumEstimator) GetLegacyGas(ctx context.Context, calldata []byte, l
 	return
 }
 
+// GetMaxLegacyGas returns the result of GetLegacyGas. Arbitrum doesn't have a mempool. That means transaction priority follows a FIFO model.
+// Fetching a standard gas estimation will have the same effect for transaction inclusion.
+func (a *arbitrumEstimator) GetMaxLegacyGas(ctx context.Context, calldata []byte, gasLimit uint64, maxGasPriceWei *assets.Wei, opts ...fees.Opt) (gasPrice *assets.Wei, chainSpecificGasLimit uint64, err error) {
+	return a.GetLegacyGas(ctx, calldata, gasLimit, maxGasPriceWei, opts...)
+}
+
 // During network congestion Arbitrum's suggested gas price can be extremely volatile, making gas estimations less accurate. For any transaction, Arbitrum will only charge
 // the block's base fee. If the base fee increases rapidly there is a chance the suggested gas price will fall under that value, resulting in a fee too low error.
 // We use gasPriceWithBuffer to increase the estimated gas price by some percentage to avoid fee too low errors. Eventually, only the base fee will be paid, regardless of the price.
--- a/pkg/gas/block_history_estimator.go
+++ b/pkg/gas/block_history_estimator.go
@@ -260,16 +260,15 @@ func (b *BlockHistoryEstimator) GetMaxLegacyGas(_ context.Context, _ []byte, _ u
 
 	ok := b.IfStarted(func() {
 		gasPrice = b.getMaxPercentileGasPrice()
-		if gasPrice == nil {
-			err = errors.New("BlockHistoryEstimator has not finished the first gas estimation yet")
-			return
-		}
 	})
 	if !ok {
 		return gasPrice, 0, errors.New("BlockHistoryEstimator is not started")
 	}
 	if !b.initialFetch.Load() {
 		return gasPrice, 0, errors.New("BlockHistoryEstimator has not finished the first gas estimation yet, likely because a failure on start")
+	}
+	if gasPrice == nil {
+		return gasPrice, 0, errors.New("BlockHistoryEstimator has not finished the first gas estimation yet")
 	}
 	return gasPrice, 0, nil
 }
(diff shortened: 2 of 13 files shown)


pkg/gas/arbitrum_estimator.go
```diff
@@ -143,6 +143,12 @@ func (a *arbitrumEstimator) GetLegacyGas(ctx context.Context, calldata []byte, l
 	return
 }
 
+// GetMaxLegacyGas returns the result of GetLegacyGas. Arbitrum doesn't have a mempool. That means transaction priority follows a FIFO model.
```
pavel-raykov, line 146: sorry, is this a connected text? What is it trying to say?

dimriou: Sorry, I'm not following. This is a two line comment. Do you mean it's not clear it's a two line comment or the description is not providing enough context?

pavel-raykov: 1. Arbitrum doesn't have a mempool. 
2. That means transaction priority follows a FIFO model.
3. Fetching a standard gas estimation will have the same effect for transaction inclusion.

These are 3 sentences. Does 1 imply 2, and then 2 imply 3? Or is there some other logic connection between them?

dimriou: Yes 1 implies 2 and as a consequence 2 implies 3.

pavel-raykov (dismissed):

pkg/gas/models.go
```diff
@@ -152,6 +153,8 @@ type EvmEstimator interface {
 	// GetLegacyGas Calculates initial gas fee for non-EIP1559 transaction
 	// maxGasPriceWei parameter is the highest possible gas fee cap that the function will return
 	GetLegacyGas(ctx context.Context, calldata []byte, gasLimit uint64, maxGasPriceWei *assets.Wei, opts ...fees.Opt) (gasPrice *assets.Wei, chainSpecificGasLimit uint64, err error)
+	// GetMaxLegacyGas returns the maximum gas price allowed by the connectivity threshold to improve inclusion probability.
```
cll-gg: Probably good to add `non-EIP1559` here, e.g.:

```suggestion
	// GetMaxLegacyGas returns the maximum non-EIP1559 gas price allowed by the connectivity threshold to improve inclusion probability.
```

pkg/gas/block_history_estimator.go
```diff
@@ -78,13 +78,11 @@ const BumpingHaltedLabel = "Tx gas bumping halted since price exceeds current bl
 
 var _ EvmEstimator = &BlockHistoryEstimator{}
 
-type estimatorGasEstimatorConfig interface {
+type BlockHistoryEstimatorConfig interface {
```
cll-gg, line 81: This wasn't an exported type before, does it need to be exported?

pkg/gas/block_history_estimator.go
```diff
@@ -255,6 +253,27 @@ func (b *BlockHistoryEstimator) GetLegacyGas(_ context.Context, _ []byte, gasLim
...
+		return gasPrice, 0, errors.New("can't get max legacy gas, EIP1559 is enabled")
+	}
+
+	ok := b.IfStarted(func() {
+		gasPrice = b.getMaxPercentileGasPrice()
+		if gasPrice == nil {
+			err = errors.New("BlockHistoryEstimator has not finished the first gas estimation yet")
+			return
+		}
+	})
```
cll-gg, line 263: This works ok, but the `return` here will only return from the inner func, which is iffy: the `err` might be set but it's not checked. Maybe clearer to add an `if err != nil` on line 268 or 270

augustbleeds (dismissed):

pkg/gas/block_history_estimator.go
augustbleeds: Yeah I think there is a silent error here . If there is an error, "ok" will still be true and we'll go to line 274

pkg/gas/models.go
dimriou: That would imply there is EIP-1559 and non-EIP-1559 gas price. Maximum gas price seems reasonable to me.

pkg/gas/block_history_estimator.go
dimriou: Yes, it's used in `gas_test` package in `block_history_estimator_test.go`

pkg/gas/models.go
cll-gg: My point is that it's hard to deduce the meaning of `legacy`. If we add `non-EIP-1559` then the reader knows that's what `legacy` means. 
Similar to what's already in [line 153](https://github.com/smartcontractkit/chainlink-evm/pull/310/changes/BASE..11a87cb8733462dc81c7f5bbc55fe70910d07f27#diff-a91dda1a0cd687f57aab587546afa5d2b388bb38a0e1523e913aba44bf2f5bfaR153).

dimriou: I'm ok with rephrasing, but non-EIP1559 gas price is not a valid statement. How about:
`GetMaxLegacyGas returns the maximum non-EIP1559 gas fee allowed by the connectivity threshold to improve inclusion probability.`
Would that be clearer?

cll-gg: Yes!

cll-gg (approved):

pkg/gas/block_history_estimator.go
augustbleeds: (And on line 274 we'd return "nil" for gas, which seems incorrect)

augustbleeds (approved):

pavel-raykov (approved):

PR #315, opened 2026-01-02
Update wrapper gen 

RensR:
This is a PR that requires a lot of teams to review. Please review the section of the code you are codeowner of. 

--

Updated CRE, Payments and shared to use the new wrapper generation, as these projects already have a `shared` folder. 

The new wrapper gen removes the custom function that's currently added to all wrappers, `ParseLog`. 

It also fixes a long standing ABI bug that happens with custom struct

Before
```
"internalType\":\"structAccountStruct\"
```

After
```
"internalType\":\"struct AccountStruct\"
```

This PR removes some old gethwrapper generation logic for Link Token. There's already a new version of the link token which has gethwrappers. The old one will never be updated and doesn't need a new wrapper to be generated. Same for the Aggregator, there's a proper wrapper so the custom code is not needed.


gethwrappers/go_generate.go
```diff
@@ -5,17 +5,12 @@ package gethwrappers
 // Make sure solidity compiler artifacts are up-to-date. Only output stdout on failure.
 //go:generate ../contracts/scripts/native_solc_compile_all
 
-//go:generate go generate ./offchainaggregator
-//go:generate go run ./generation/generate_link/wrap_link.go
```
RensR, line 9: This was custom code to generate wrappers that have not been modified in years, and are never expected to change.

gethwrappers/keystone/go_generate.go
```diff
@@ -1,11 +0,0 @@
-// Package gethwrappers provides tools for wrapping solidity contracts with
-// golang packages, using abigen.
-package gethwrappers
```
RensR, line 3: Has moved to the cre folder

gethwrappers/workflow/go_generate.go
```diff
@@ -1,12 +0,0 @@
-// Package gethwrappers provides tools for wrapping solidity contracts with
-// golang packages, using abigen.
-package gethwrappers
-
-// Workflow
-
```
RensR, line 6: Has moved to the cre folder

contracts/cre/gobindings/wrap/wrap.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/smartcontractkit/chainlink-evm/gethwrappers/helpers/generate/wrap"
+	zksyncwrapper "github.com/smartcontractkit/chainlink-evm/gethwrappers/helpers/zksync"
+)
+
+func main() {
```
RensR, line 13: This file is the only thing a project needs to write, with gives maximum flexibility while keeping the code duplication to a minumum

contracts/cre/src/workflow/dev/test/MockKeystoneForwarder_BasicTest.t.sol
```diff
@@ -2,24 +2,36 @@
 pragma solidity 0.8.24;
 
 import {IRouter} from "../../../keystone/interfaces/IRouter.sol";
+import {IReceiver} from "../../../keystone/interfaces/IReceiver.sol";
```
RensR, line 5: Tests had been broken since test CI isn't required. It's getting fixed

dhaidashenko (approved):

RayXpub (approved):

ibrajer (approved):

pavel-raykov (approved):

PR #316, opened 2026-01-03
improve flexibility of gethwrapper helpers

RensR:
The previous helpers required abigen to be in a hard coded relative location. These changes allow us to call the gethwrapper logic directly with args for the abigen executable.

### DIFF
--- a/gethwrappers/helpers/generate/wrap.go
+++ b/gethwrappers/helpers/generate/wrap.go
@@ -9,7 +9,7 @@ import (
 )
 
 func main() {
-	projectRoot := os.Args[1]
+	solcProjectRoot := os.Args[1]
 	abiGenPath := os.Args[2]
 	contract := os.Args[3]
 	pkgName := os.Args[4]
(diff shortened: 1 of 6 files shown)


gethwrappers/helpers/generate/wrap.go
```diff
@@ -1,107 +1,27 @@
...
+	contract := os.Args[3]
+	pkgName := os.Args[4]
 
-	var outDirSuffix string
-	if len(os.Args) >= 5 {
-		outDirSuffix = os.Args[4]
-	} else {
-		outDirSuffix = "latest"
-	}
+	outDirSuffix := "latest"
```
RensR, line 17: Wrappers are always generated into `latest`. Teams that wish to adopt this new wrapper generation should also adopt the versioned gethwrappers.

gethwrappers/helpers/generate/wrap/wrap.go
```diff
@@ -0,0 +1,77 @@
...
+// contract is the name of the generated contract class.
+// pkgName is the name of the package the contract will be generated in. Try
+// to follow idiomatic Go package naming conventions where possible.
+//
+// outDirSuffixInput is the directory suffix to generate the wrapper in. If not provided, the
+// wrapper will be generated in the default location. The default location is
+// <project>/generated/<pkgName>/<pkgName>.go. The suffix will take place after
+// the <project>/generated, so the overridden location would be
+// <project>/generated/<outDirSuffixInput>/<pkgName>/<pkgName>.go.
+func GenWrapper(projectRoot, contract, pkgName, outDirSuffixInput, abiGenPath string) {
```
RayXpub: q: is this function not used in this repo and only intended to be a helper for other repos? it looks to me like this repo only uses https://github.com/smartcontractkit/chainlink-evm/blob/modify-gethwrapper-helpers/gethwrappers/generation/generate/genwrapper/genwrapper.go?
nit: could we add a comment/example of what the projectRoot arg is?

RayXpub (approved):

0xsuryansh (approved):

PR #318, opened 2026-01-06
Txm compatibility with keystore lib

connorwstein:
The existing EVM keystore interface (built on top of the the common core.Keystore interface https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/types/core/keystore.go#L23) assumes that the "accounts" in the common keystore are all EVM addresses: https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/keys/keystore.go#L113.  

That works fine if the core.Keystore implementation is backed by the core node which always uses the address as the identifier. However, if the core.Keystore implementation is backed by the newer keystore library, the account name isn't an address anymore but a user specified key name with an EVM prefix (e.g. evm/tx/myPolygonKey1). 

The fix proposed here is to just add a EVM keystore implementation which supports address based lookups. That way users of keys/v2 can still use the EVM keystore interface needed for the Txm. For example:

```
// Create a v2 key
ks, _ := keystore.LoadKeystore(ctx, keystore.NewFileStorage(tmpfile.Name()), "password")
txKey, _ := keysv2.CreateTxKey(ks, "key1")

// Use it with the TXM without any address wrangling. 
evmKeyStore := keys.NewStore(NewTxKeyCoreKeystore(ks))
// This will now work because TxKeyCoreKeystore exposes/works with addresses as Accounts.
addrs, _ := evmKeyStore.EnabledAddresses(context.Background())

myTxm := txm.NewTxm(..., evmKeyStore)
```

### DIFF
--- a/pkg/keys/v2/core_keystore.go
+++ b/pkg/keys/v2/core_keystore.go
@@ -25,10 +25,6 @@ func NewTxKeyCoreKeystore(ks keystore.Keystore) *TxKeyCoreKeystore {
 	}
 }
 
-func (s *TxKeyCoreKeystore) GetCache() map[string]string {
-	return s.cache
-}
-
 func (s *TxKeyCoreKeystore) Accounts(ctx context.Context) ([]string, error) {
 	keys, err := GetTxKeys(ctx, s.ks, []string{})
 	if err != nil {
--- a/pkg/keys/v2/core_keystore.go
+++ b/pkg/keys/v2/core_keystore.go
@@ -1,0 +1,78 @@
+package keys
+
+import (
+	"context"
+	"errors"
+
+	"github.com/smartcontractkit/chainlink-common/keystore"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/core"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/keys/v2/core_keystore.go
```diff
@@ -0,0 +1,78 @@
...
+func NewTxKeyCoreKeystore(ks keystore.Keystore) *TxKeyCoreKeystore {
+	return &TxKeyCoreKeystore{
+		ks:    ks,
+		cache: make(map[string]string),
+	}
+}
+
+func (s *TxKeyCoreKeystore) GetCache() map[string]string {
+	return s.cache
+}
```
jmank88: Could this be moved to the `_test.go` file?

connorwstein: yeah for sure - I was thinking it might be mildly useful exported but totally ok with just keeping private

jmank88 (approved):

MStreet3 (approved):

PR #321, opened 2026-01-08
Rename CRE contracts 'keystone' -> 'capabilities'

justinkaseman:
### Description
CRE specific contracts were recently moved to `contracts/cre`. This PR cleans up some of the legacy naming and adds versions folders in front of the contracts.

### Changes
- Renames 'keystone' to 'core'
- Adds version folders within 'core'
- Adds code owners to specific CRE folders

### Additional Steps
1. Regen gas snapshots
2. Regen geth wrappers


contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
```diff
@@ -0,0 +1,1411 @@
+// SPDX-License-Identifier: MIT
```
ibrajer: @justinkaseman is this copied by accident? we already have it under cre/src/core/v2 right?

contracts/cre/gas-snapshots/workflow-v1.gas-snapshot
```diff
@@ -0,0 +1,64 @@
+WorkflowRegistryManager_activateVersion:test_WhenTheVersionNumberIsNotActive_AndWhenThereAreNoActiveVersions() (gas: 529151)
```
ibrajer, line 1: @justinkaseman I think we should delete `workflow.gas-snapshot` if we now have a way to generate one for each version? But I guess that also applies to core as well (I only see one gas snapshot so far)?

contracts/src/v0.8/workflow/dev/v2/CapabilitiesRegistry.sol
justinkaseman: Oooh. Good catch. This wasn't supposed to be there.

contracts/cre/gas-snapshots/workflow-v1.gas-snapshot
justinkaseman: I made it so that it has a snapshot for each version for both `core` and `workflow`.

contracts/cre/foundry.toml
```diff
@@ -43,13 +43,22 @@ solc_version = '0.8.26'
...
 optimizer_runs = 1_000_000
 solc_version = '0.8.24'
-src = 'src/keystone'
-test = 'src/keystone/test'
+src = 'src/core/v1'
+test = 'src/core/v1/test'
 via_ir = false
 
+[profile.core-v2]
+optimizer_runs = 1_000
```
RensR, line 57: Just 1k?

contracts/cre/foundry.toml
```diff
@@ -43,13 +43,22 @@ solc_version = '0.8.26'
 src = 'src/workflow/v2'
 test = 'src/workflow/v2/test'
 
-[profile.keystone]
+[profile.core]
+src = 'src/core'
```
RensR, line 47: Would this ever be used or would you always specify v1/v2?

contracts/cre/foundry.toml
```diff
@@ -43,13 +43,22 @@ solc_version = '0.8.26'
...
+src = 'src/core/v1'
+test = 'src/core/v1/test'
 via_ir = false
 
+[profile.core-v2]
+optimizer_runs = 1_000
+solc_version = '0.8.26'
+src = 'src/core/v2'
+test = 'src/core/v2/test'
+
```
RensR, line 61: Not turning off via-ir here is a deviation from previous code, same for the Solidity version. Is this intentional?

justinkaseman:
Re-working to be a single CODEOWNER.

PR #332, opened 2026-01-16
Add TXMv2 integration tests

dimriou:
This PR adds integration tests for TXMv2. It can support both Devnet and any EVM-based Testnet.
The goal is to create different cases by changing the behaviour of the simulated client and forcing the TXM to handle each case.


pkg/txm/integration-tests/config_wrapper.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+const (
+	envPath    = "env.toml"
+	configPath = "configs.toml"
+)
+
+// This wrapper is required because of the way Gas Estimator components expect configs.
+// Instead of passing down a struct with values, we need to implement an interface with
+// the required methods.
```
cll-gg, line 28: Is the interface known? If so, good to add a `var _ <interface> = (AppConfig)(nil)` line

pkg/txm/integration-tests/integration_test.go
```diff
@@ -0,0 +1,349 @@
...
+		return txm, store, chainID, lggr, fromAddress, simulatedClient, configs
+	}
+	return nil, nil, nil, nil, common.Address{}, nil, AppConfig{}
+}
+
+func setupGasEstimator(t *testing.T, lggr logger.Logger, observedLogs *observer.ObservedLogs, client gas.FeeEstimatorClient, chainID *big.Int, configs AppConfig) gas.EvmFeeEstimator {
+	estimator, err := gas.NewEstimator(lggr, client, "", chainID, configs, nil)
+	require.NoError(t, err)
+	servicetest.Run(t, estimator)
+	require.NoError(t, err)
```
cll-gg: Looks like this is supposed to check the error returned from `servicetest.Run()` (if it's there)

pkg/txm/integration-tests/config_wrapper.go
dimriou: No, this is a superset of multiple interfaces declared in each one of the underlying components.

cll-gg: Ah ok

alejoberardino (approved):
Looks great to me, we can add fancier cases on top of this afterwards. Very nice 👌

jmank88 (approved):

PR #333, opened 2026-01-19
PLEX-2281: addresses with no contract shouldn't be able to register i…

fernandezlautaro:
…n log poller

Jira: https://smartcontract-it.atlassian.net/browse/PLEX-2281


dhaidashenko (dismissed):

pkg/logpoller/log_poller.go
```diff
@@ -284,6 +285,21 @@ func (lp *logPoller) RegisterFilter(ctx context.Context, filter Filter) error {
 		}
 	}
 
+	var invalidAddresses []string
+	for i, addr := range filter.Addresses {
+		code, err := lp.ec.CodeAt(ctx, addr, nil)
```
dhaidashenko, line 291: I wonder what will happen if RPC is down or unavailable. Will workflow DON try to reregister trigger?

dhaidashenko: If not, maybe we should add some retry logic here

fernandezlautaro: RN probably yes, as we are not discriminating between user and system error.

But this comment made realize once that's fixed on platform side, this coming from the internals calling here [lts.EVMService.RegisterLogTracking(ctx, filterQuery)](https://github.com/smartcontractkit/capabilities/blob/main/chain_capabilities/evm/trigger/trigger.go#L263) will clearly be swallowed up which defeats the purpose of this early validation. 

This being clearly a user error needs to be done differently.

Will update PR to throw a particular error in logPoller and in log trigger check if it needs to be a user error or not.

fernandezlautaro: Updated code here to export the user error, and added a regression test here in chainlink [Test_CRE_V2_EVM_LogTrigger_Invalid_Address_Regression](https://github.com/smartcontractkit/chainlink/pull/20836/changes#diff-f0bb51253a4ad0e1401c63b38d32c9d7279a2ed92253497640f2af6ec7419d78) to show that using local cre+loops the correct error bubbles up

dhaidashenko (approved):

ilija42 (approved):

PR #339, opened 2026-01-22
OEV-779 Send custom event on atlas error

cll-gg:
### Depends on
* https://github.com/smartcontractkit/chainlink-protos/pull/270
* https://github.com/smartcontractkit/chainlink-protos/pull/276
* https://github.com/smartcontractkit/chainlink-protos/pull/277


cll-gg:
> ## ⚠️ API Diff Results - Breaking changes detected
> ### 📦 Module: `github-com-smartcontractkit-chainlink-evm`
> #### 🔴 Breaking Changes (1)
> ##### `pkg/txm/clientwrappers/dualbroadcast` (1)
> * **`NewMetaMetrics`** — Type changed:
> 
> func(
>   - ~string~
>   + string, 
>   + github.com/smartcontractkit/chainlink-common/pkg/logger.Logger
> )
> (*MetaMetrics, error)
> 📄 [View full apidiff report](https://github.com/smartcontractkit/chainlink-evm/actions/runs/21482079487/#summary-61880888205)

No problem, NewMetaMetrics is only used in chainlink-evm, [see here](https://github.com/search?q=org%3Asmartcontractkit%20NewMetaMetrics&type=code)

alejoberardino (dismissed):

cll-gg:
CI failures are due to https://github.com/smartcontractkit/chainlink-protos/pull/264 not being merged, will fix that when it's merged.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -98,3 +109,64 @@ func (m *MetaMetrics) RecordSendOperationError(ctx context.Context) {
...
 	)
 }
+
+// emitAtlasError emits an OTel event to track FastLane Atlas errors
+func (m *MetaMetrics) emitAtlasError(ctx context.Context, errType string, customURL *url.URL, err error, tx *types.Transaction) {
+	m.emitAtlasErrorWithHttpStatusCode(ctx, errType, customURL, err, -1, tx)
+}
+
+// emitAtlasErrorWithHttpStatusCode emits an OTel event with an HTTP status code to track FastLane Atlas errors
+func (m *MetaMetrics) emitAtlasErrorWithHttpStatusCode(ctx context.Context, errType string, customURL *url.URL, cause error, httpStatusCode int, tx *types.Transaction) {
```
dimriou: If the main code doesn't ever utilize the `httpStatusCode` param, is there a point for tests to test for it? I feel we can avoid the passthrough method on line 115 and instead have one function.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -98,3 +109,64 @@ func (m *MetaMetrics) RecordSendOperationError(ctx context.Context) {
...
+		"beholder_domain", "svr",
+		"beholder_entity", "svr.v1.FastLaneAtlasError",
+		"beholder_data_schema", "/fastlane-atlas-error/versions/1",
+	}
+
+	mStr := protojson.MarshalOptions{
+		UseProtoNames:   true,
+		EmitUnpopulated: true,
+	}.Format(msg)
+	m.lggr.Infow("[Beholder.emit]", "message", mStr, "attributes", attrKVs)
```
dimriou, line 160: I'm on the fence as to whether this needs to be info or debug.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: It's used in the Atlas error struct, [see here](https://github.com/smartcontractkit/chainlink-evm/pull/339/changes#diff-66af629d1dde9bf676736bd41d82533307749f350551c1e620256313bc473b06R144)

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: We should only get a lot of errors in case of erroneous circumstances, then I think it'd be good to have these logs

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
dimriou: It's in the Atlas error struct but besides tests, is there another value passed other than -1?

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
dimriou: But since this is a node log, we get this information from other sources as well, so this just confirms that it sends the Beholder message on an info level. Anyway, as I mentioned, I don't have a strong preference so we can keep it like that.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: Ah I see what you mean, good call 👍 
I've removed the method and set the status code in the event to -1 for now, will remove the field from the schema in a separate PR.

dimriou (approved):

george-dorin (approved):

PR #343, opened 2026-01-30
Split MetaClient errors from standard RPC errors

dimriou:
This PR distinguishes the errors returned by the MetaClient from the standard RPC errors and marks as fatal the former. Additionally, it for the RPC errors it now rebroadcasts a past attempt instead of reauctioning.


pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -163,27 +170,39 @@ func (a *MetaClient) PendingNonceAt(ctx context.Context, address common.Address)
 	return a.c.PendingNonceAt(ctx, address)
 }
 
+// SendTransactions handles three diffenent cases:
```
cll-gg:
```suggestion
// SendTransactions handles three different cases:
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -163,27 +170,39 @@ func (a *MetaClient) PendingNonceAt(ctx context.Context, address common.Address)
...
+					return fmt.Errorf("failed to send operation for transactionID(%d): %w", tx.ID, ErrAuction)
+				}
+				return nil
+			}
+			a.lggr.Infof("No bids for transactionID(%d): ", tx.ID)
+			return ErrNoBids
+		}
+		if len(tx.Attempts) > 1 {
+			a.lggr.Infow("Intercepted attempt for tx(rebroadcasting first attempt)", "txID", tx.ID, "attempt", tx.Attempts[0])
+			return a.c.SendTransaction(ctx, tx.Attempts[0].SignedTransaction)
```
cll-gg: Maybe it's me, but I find it hard to reason about this many `if`s. Is there any way to make this less indented, maybe through an early return?

dimriou: Wasn't sure about simplifying the `is.IsPurgeable` statement as well, so I can revert this back to three main ifs which are explained at the top of the method.
1. First failed transmission
2. Other attempts
3. Empty transaction
Note: Part of the reason why this doesn't have much info, is because the Atlas integration was not publicly known before.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -116,9 +116,14 @@ const (
 )
 
 var ErrNoBids = errors.New("no bids")
+var ErrAuction = errors.New("auction error")
 
 var _ txm.Client = &MetaClient{}
 
+type MetaClientTxStore interface {
+	UpdateSignedAttempt(_ context.Context, txID uint64, attemptID uint64, signedTransaction *evmtypes.Transaction, fromAddress common.Address) error
```
cll-gg, line 124: This could use a comment/bit of documentation

cll-gg (dismissed):

cll-gg (approved):

HashWrangler (approved):

dhaidashenko (approved):

ilija42 (approved):

PR #344, opened 2026-01-30
pkg/logpoller: fix logPoller.loadFilters race

jmank88:
https://github.com/smartcontractkit/chainlink/actions/runs/21515300315/job/61992089790

### DIFF
--- a/pkg/logpoller/log_poller.go
+++ b/pkg/logpoller/log_poller.go
@@ -591,10 +591,17 @@ func (lp *logPoller) GetReplayFromBlock(ctx context.Context, requested int64) (i
 // loadFilters loads the filters from db, and activates count-based Log Pruning
 // if required by any of the filters
 func (lp *logPoller) loadFilters(ctx context.Context) error {
-	filters, err := lp.lockAndLoadFilters(ctx)
+	lp.filterMu.Lock()
+	defer lp.filterMu.Unlock()
+
+	filters, err := lp.orm.LoadFilters(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/logpoller/log_poller.go
```diff
@@ -591,10 +591,17 @@ func (lp *logPoller) GetReplayFromBlock(ctx context.Context, requested int64) (i
 // loadFilters loads the filters from db, and activates count-based Log Pruning
 // if required by any of the filters
 func (lp *logPoller) loadFilters(ctx context.Context) error {
-	filters, err := lp.lockAndLoadFilters(ctx)
+	lp.filterMu.Lock()
+	defer lp.filterMu.Unlock()
```
jmank88, line 595: Can't unlock until after we iterate over the map in the for loop.

pkg/logpoller/log_poller.go
```diff
@@ -591,10 +591,17 @@ func (lp *logPoller) GetReplayFromBlock(ctx context.Context, requested int64) (i
...
+
+	filters, err := lp.orm.LoadFilters(ctx)
```
pavel-raykov, line 597: sorry, I am not sure why this change is better - where was the leak before?

jmank88: The lock was only held by the helper, so the map was being modified during the for loop iteration.

pavel-raykov: sorry, you are right.

pavel-raykov (approved):

silaslenihan (approved):

dhaidashenko (approved):

PR #346, opened 2026-02-04
Oev 771 flashbots batch

dimriou:
This PR enables Flashbots bundles. Every time a transaction is broadcasted, it will fetch all the past in-flight transactions and create a bundle. The bundle will use the last attempt of each transaction. The bundle broadcasting is a fire and forget process that helps nonce unblocking, so it is not tracked.


pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
 		}
 		body := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method":"eth_sendRawTransaction","params":["%s"], "id":1}`, hexutil.Encode(data)))
 		_, err = d.signAndPostMessage(ctx, tx.FromAddress, body, params)
-		return err
+		if err != nil {
+			return err
+		}
+
+		// After successfully sending the transaction, send a bundle with all unconfirmed transactions
```
pszal, line 109: Why not send it regardless eth_sendRawRx failing?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+	unconfirmedTxs, err := d.txStore.FetchUnconfirmedTransactions(ctx, fromAddress)
+	if err != nil {
+		return fmt.Errorf("failed to fetch unconfirmed transactions: %w", err)
+	}
+
+	// TODO: Get the first attempt from each transaction for now.
+	attempts := make([]*types.Attempt, 0, len(unconfirmedTxs))
+	nonces := make([]uint64, 0, len(unconfirmedTxs))
+	for _, unconfirmedTx := range unconfirmedTxs {
+		if len(unconfirmedTx.Attempts) > 0 {
```
pszal: A bit unclear to me. Maybe add a comment clarifying why `attemtps` becomes a slice of pending txs with consecutive nonces? And what `len(unconfirmedTx.Attempts)` really means?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+			attempts = append(attempts, unconfirmedTx.Attempts[0])
+			if unconfirmedTx.Nonce != nil {
```
pszal: nil cannot happen since `FetchUnconfirmedTransactions` filters for that, right?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+			}
+		}
+	}
+
+	// Need at least 2 transactions to send a bundle
+	if len(attempts) < 2 {
+		return nil
+	}
+
+	// TODO: we don't have a good way to get this other than making an RPC call. Some async caching may help with the overhead.
```
pszal, line 215: That's a shame. AFAIR the single tx API has the default (0 = curr block), maybe it is worth checking with FB do they support it here too?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+	currentBlock, err := d.c.BlockByNumber(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get current block height: %w", err)
+	}
+	targetBlock := currentBlock.NumberU64() + 24
```
pszal: would `maxBlock` be a more accurate var name?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+		}
+
+		txData, err := attempt.SignedTransaction.MarshalBinary()
+		if err != nil {
+			return fmt.Errorf("failed to marshal transaction for attempt ID %d: %w", attempt.ID, err)
+		}
+
+		bodyItems = append(bodyItems, map[string]any{
+			"tx":         hexutil.Encode(txData),
+			"revertMode": "allow", // we always want to allow reverts so bundles are valid even if a single transaction within the bundle goes through
```
pszal, line 233: I don't see `revertMode`, only `canRevert`:
https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,213 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+		BundleHash string `json:"bundleHash"`
+	}
+	if err := json.Unmarshal(raw, &bundleResult); err != nil {
+		return fmt.Errorf("failed to decode response %v into bundle result: %w", raw, err)
+	}
+	d.lggr.Infow("Broadcasted transaction bundle", "nonces", nonces, "bundleHash", bundleResult.BundleHash)
+	return nil
+}
+
+func parseURLParams(params string) (Privacy, RefundConfig, error) {
```
pszal, line 283: Is the params string part of the config? No way around it? 
If it has to stay please consider some solid testing.

pszal (commented):
Nice! Some feedback below. Please consider adding more tests.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: I was on the fence with this one. The biggest reason was I didn't want to change the behaviour of the main flow. Here's a scenario I thought of:
The single transaction fails, the bundle is successful in transmitting, including the highest nonce transaction, but the main flow reads the response of the single transaction and makes false assumptions. Currently, there won't be any problems with TXM's logic but I wanted to be somewhat defensive at launch. Happy to discuss further though, if we think we get a significant advantage.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Transaction is the main structure that holds the information and an attempt is what actually gets sent to the network. Different attempts may have different fees or gas limits, which means they might have a different signature.`FetchUnconfirmedTransactions` fetches all the unconfirmed transactions in order. By design there are no nonce gaps on the fetched transactions. So we have to pick one attempt. In the majority of cases we only broadcast one attempt that's why I'm picking the first one here. `len(unconfirmedTx.Attempts)` is just a safety measure to avoid panics.

dimriou: I can add some of the info as a comment.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Indeed, I just added it for safety in case something gets changed by mistake.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: But that would still require an API call to their endpoint, right?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Indeed, but I followed the spec they gave me in our conversation.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: The thing is when we first implemented the solution you could add configs through URL params. According to flashbots, for bundles, this is not supported anymore and they require the configs to be added straight to the request body, hence the diff here. I can add some testing.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
pszal: I see!

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
pszal: A comment would be helpful!

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
pszal: No, I though that if we set it to 0 then mev-share would replace it with the currBlock, but seems they do it only for the maxBlock: https://github.com/flashbots/mev-share-node/blob/67003f8e8700bef34444dda71dc18c4585e5c2f6/mevshare/bundle_validation.go#L103

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
pszal: Does it make sense to document this spec somewhere (e.g., doc.go)?

dimriou: Added a link to the docs.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -42,15 +63,21 @@ func (d *FlashbotsClient) NonceAt(ctx context.Context, address common.Address, b
...
-	response, err := d.signAndPostMessage(ctx, address, body, "")
+	raw, err := d.signAndPostMessage(ctx, address, body, "")
 	if err != nil {
 		return 0, err
 	}
 
-	nonce, err := hexutil.DecodeUint64(response)
+	var resultStr string
+	if err := json.Unmarshal(raw, &resultStr); err != nil {
+		return 0, fmt.Errorf("failed to unmarshal response into string: %w", err)
```
pszal: I see that you log arg when it fails in other places, any reason to not log `raw` here?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,218 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+
+	// We fetch all the unconfirmed transactions in an ascending nonce order.
+	// For the bundle we need a signed transaction so we get the last attempt from each transaction.
+	// TODO: Implement a more sophisticated attempt selection logic if necessary.
+	attempts := make([]*types.Attempt, 0, len(unconfirmedTxs))
+	nonces := make([]uint64, 0, len(unconfirmedTxs))
+	for _, unconfirmedTx := range unconfirmedTxs {
+		if len(unconfirmedTx.Attempts) > 0 {
+			attempts = append(attempts, unconfirmedTx.Attempts[len(unconfirmedTx.Attempts)-1])
+			if unconfirmedTx.Nonce != nil {
```
pszal: if you check for that here, then it makes sense to check or guarantee in the if above that `attempts` and `nonces` are consistent (i.e., you append to both or neither).

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,218 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+
+	privacy.Builders = append(privacy.Builders, values["builder"]...)
+
+	privacy.Hints = append(privacy.Hints, values["hint"]...)
+
+	refundConfig := RefundConfig{}
+	refundRaw := values.Get("refund")
+	if refundRaw != "" {
+		parts := strings.Split(refundRaw, ":")
+		if len(parts) == 2 {
```
pszal: is it ok if its not 2? shouldn't we err?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,218 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+		return err
+	}
+
+	var bundleResult struct {
+		BundleHash string `json:"bundleHash"`
+	}
+	if err := json.Unmarshal(raw, &bundleResult); err != nil {
+		return fmt.Errorf("failed to decode response %v into bundle result: %w", raw, err)
+	}
+	d.lggr.Infow("Broadcasted transaction bundle", "nonces", nonces, "bundleHash", bundleResult.BundleHash)
```
pszal: is it intentional to log only nonces and not the actual bundle (or at least IDs of its txs)?

pszal (commented):
lgtm

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: No reason, I added `raw`  as well.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Good call

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Yes. Logs can be very noisy in Core, so the idea here is avoid replicating information that can be recovered by past logs. Similar to how we confirm transactions, we use nonce as an identifier to retrieve information if needed, as the individual transaction details should be available a few lines before this message. I'll add the IDs though to make it cleaner.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Yeah, agree

dhaidashenko (dismissed):

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,222 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
-		return err
+		if err != nil {
+			return err
+		}
+
+		// After successfully sending the transaction, send a bundle with all unconfirmed transactions
+		// Don't act on a bundle error - this is a fire and forget operation but we do want to log the error.
+		if d.bundles {
+			if err := d.SendBundle(ctx, tx.FromAddress, params); err != nil {
+				d.lggr.Error("error sending bundle: ", err)
```
amit-momin: nit: wouldn't it be easier for log parsing to do something like
```suggestion
				d.lggr.Errorw("error sending bundle", "err", err)
```

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,222 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+// SendBundle sends a bundle of all the in-flight transactions.
+func (d *FlashbotsClient) SendBundle(ctx context.Context, fromAddress common.Address, urlParams string) error {
+	unconfirmedTxs, err := d.txStore.FetchUnconfirmedTransactions(ctx, fromAddress)
+	if err != nil {
+		return fmt.Errorf("failed to fetch unconfirmed transactions: %w", err)
+	}
+
+	// We fetch all the unconfirmed transactions in an ascending nonce order.
+	// For the bundle we need a signed transaction so we get the last attempt from each transaction.
+	// TODO: Implement a more sophisticated attempt selection logic if necessary.
```
amit-momin, line 184: nit: Since this is ready to merge, should we remove the TODOs or is this for later?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,222 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+	attempts := make([]*types.Attempt, 0, len(unconfirmedTxs))
+	attemptIDs := make([]uint64, 0, len(unconfirmedTxs))
+	nonces := make([]uint64, 0, len(unconfirmedTxs))
+	ids := make([]uint64, 0, len(unconfirmedTxs))
+	for _, unconfirmedTx := range unconfirmedTxs {
+		if len(unconfirmedTx.Attempts) > 0 && unconfirmedTx.Nonce != nil {
+			latestAttempt := unconfirmedTx.Attempts[len(unconfirmedTx.Attempts)-1]
+			attempts = append(attempts, latestAttempt)
+			attemptIDs = append(attemptIDs, latestAttempt.ID)
+			nonces = append(nonces, *unconfirmedTx.Nonce)
```
amit-momin: Should we do some sanity check to make sure there aren't any gaps for whatever reason here?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -72,57 +99,222 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
...
+	if len(attempts) < 2 {
+		return nil
+	}
+
+	// TODO: we don't have a good way to get this other than making an RPC call. Some async caching may help with the overhead.
+	currentBlock, err := d.c.BlockByNumber(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get current block height: %w", err)
+	}
+	maxBlock := currentBlock.NumberU64() + 24
```
amit-momin: Where does this 24 come from?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: I have this as a placeholder after the change goes live so we can revisit.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: I've posted a link above with Flashbots' spec. 24 is their max allowed value, I can make it a const.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: By design InMemory store doesn't create nonce gaps but since it was mentioned by another reviewer, I'll add it.

amit-momin (approved):

dhaidashenko (approved):

ilija42 (approved):

PR #352, opened 2026-02-11
add configurable historical balance health check for finalized block …

Krish-vemula:
## Summary

Adds a multinode health check that detects RPC nodes unable to serve historical blockchain state (e.g. non-archive nodes). When enabled, nodes that fail consecutive checks are transitioned to a new `FinalizedStateNotAvailable` state and removed from the pool until they recover.

### Changes

**New Configuration Options (`[EVM.NodePool]`):**
- `FinalizedStateCheckFailureThreshold` (uint32, default: `0`) - Number of consecutive failures before marking node as `FinalizedStateNotAvailable`. Set to `0` to disable the check.
- `HistoricalBalanceCheckAddress` (address, default: `0x0...0`) - The address used for the `eth_getBalance` probe.

**New Configuration Option (`[EVM.NodePool.Errors]`):**
- `FinalizedStateUnavailable` (string) - Regex pattern to classify RPC errors as "state unavailable" (e.g. `missing trie node`) vs generic reachability errors.

### Implementation

- Polling is controlled entirely by `multinode` via `FinalizedStateCheckFailureThreshold > 0` - the EVM client always executes the check when called
- On each poll tick, `eth_getBalance` is called at the finalized block:
  - `FinalityTagEnabled = true`: uses the `"finalized"` block tag
  - `FinalityTagEnabled = false`: uses `latest - FinalityDepth`
- Errors matching `FinalizedStateUnavailable` regex return `ErrFinalizedStateUnavailable`; other errors are treated as generic RPC reachability failures
- After `FinalizedStateCheckFailureThreshold` consecutive `ErrFinalizedStateUnavailable` errors, the node transitions to `nodeStateFinalizedStateNotAvailable`
- Recovery: node re-enters the pool only after a successful `CheckFinalizedStateAvailability` poll

### Configuration Example

```toml
[[EVM]]
ChainID = '1'
FinalityTagEnabled = true

[EVM.NodePool]
PollInterval = '10s'
FinalizedStateCheckFailureThreshold = 3
HistoricalBalanceCheckAddress = '0x0000000000000000000000000000000000000000'

[EVM.NodePool.Errors]
FinalizedStateUnavailable = 'missing trie node'
```


jmank88 (dismissed):

pkg/client/helpers_test.go
```diff
@@ -118,6 +120,14 @@ func (tc TestNodePoolConfig) FinalizedBlockPollInterval() time.Duration {
 	return tc.NodeFinalizedBlockPollInterval
 }
 
+func (tc TestNodePoolConfig) HistoricalBalanceCheckEnabled() bool {
```
dhaidashenko: This flag should be part of the multinode config. As multinode will perform the polling

pkg/client/rpc_client.go
```diff
@@ -180,10 +184,35 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
...
+	if r.historicalBalanceCheckEnabled {
+		if err = r.checkHistoricalStateAtFinalized(ctx); err != nil {
+			return "", fmt.Errorf("historical balance health check failed: %w", err)
+		}
+	}
 	r.rpcLog.Debugf("client version: %s", version)
 	return version, nil
 }
 
+func (r *RPCClient) checkHistoricalStateAtFinalized(ctx context.Context) error {
```
dhaidashenko: This should be public method, so that we can poll it from multinode

pkg/client/helpers_test.go
```diff
@@ -142,6 +156,22 @@ func (tc TestNodePoolConfig) ExternalRequestMaxResponseSize() uint32 {
 	return tc.ExternalRequestMaxResponseSizeVal
 }
 
+func (tc TestNodePoolConfig) FinalizedStateCheckEnabled() bool {
```
dhaidashenko: Why do we need two feature flags?

pkg/client/rpc_client.go
```diff
@@ -184,6 +188,39 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
 	return version, nil
 }
 
+// CheckFinalizedStateAvailability verifies if the RPC can serve historical state at the finalized block.
+// This is used to detect non-archive nodes that cannot serve state queries for older blocks.
+// If probeAddress is provided, it uses that address; otherwise falls back to the configured HistoricalBalanceCheckAddress.
+// Returns nil immediately if historical balance check is not enabled and no probeAddress is provided.
+func (r *RPCClient) CheckFinalizedStateAvailability(ctx context.Context, probeAddress string) error {
```
dhaidashenko: IMO, there is no need to allow overriding the address. It adds complexity but we do not know if it'll be used.

pkg/client/rpc_client.go
```diff
@@ -184,6 +188,39 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
...
+		latestBlock, err := r.BlockNumber(ctx)
+		if err != nil {
+			return fmt.Errorf("fetching latest block number failed: %w", err)
+		}
+		latest := int64(latestBlock)
+		finalizedHeight := max(int64(0), latest-int64(r.finalityDepth))
+		blockNumber = big.NewInt(finalizedHeight)
+	}
+	_, err := r.BalanceAt(ctx, addr, blockNumber)
+	if err != nil {
```
dhaidashenko, line 217: We need to check the error against the regex here and return a dedicated error defined in the framework repo.
This way, we abstract away the EVM-specific way of checking the availability of finalized states.

pkg/config/chain_scoped_node_pool.go
```diff
@@ -38,6 +40,20 @@ func (n *NodePoolConfig) FinalizedBlockPollInterval() time.Duration {
 	return n.C.FinalizedBlockPollInterval.Duration()
 }
 
+func (n *NodePoolConfig) HistoricalBalanceCheckEnabled() bool {
+	if n.C.HistoricalBalanceCheckEnabled == nil {
```
dhaidashenko: Configs are built in a way that the value should never be nil. So here we should skip checks to keep everything consistent. 

Also, you need to add new config fields to [setFrom method](https://github.com/smartcontractkit/chainlink-evm/blob/1e75633e454ed6c5fb65ca257d43394eeea3ad64/pkg/config/toml/config.go#L1100).
It is required to handle overrides properly. 
By overrides I mean the ability to define a common value in the fallback file and override it in a chain-specific TOML file or during runtime via a custom TOML file.

pkg/client/helpers_test.go
```diff
@@ -88,18 +88,24 @@ func (c *TestClientErrors) TooManyResults() string          { return c.tooManyRe
...
+	HistoricalBalanceCheckAddressVal        string
+	NodeErrors                              config.ClientErrors
+	EnforceRepeatableReadVal                bool
+	NodeDeathDeclarationDelay               time.Duration
+	NodeNewHeadsPollInterval                time.Duration
+	ExternalRequestMaxResponseSizeVal       uint32
+	FinalizedStateCheckEnabledVal           bool
+	FinalizedStateCheckAddressVal           string
+	FinalizedStateCheckFailureThresholdVal  uint32
+	FinalizedStateUnavailableRegexVal       string
```
dhaidashenko: This regex should be defined as part of [ClientErrors](https://github.com/smartcontractkit/chainlink-evm/blob/1e75633e454ed6c5fb65ca257d43394eeea3ad64/pkg/config/config.go#L93)

pkg/client/rpc_client.go
```diff
@@ -184,6 +189,53 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
 	return version, nil
 }
 
+// CheckFinalizedStateAvailability verifies if the RPC can serve historical state at the finalized block.
+// This is used to detect non-archive nodes that cannot serve state queries for older blocks.
+// Returns nil immediately if historical balance check is not enabled.
+// Returns multinode.ErrFinalizedStateUnavailable if the error matches the FinalizedStateUnavailable pattern.
+func (r *RPCClient) CheckFinalizedStateAvailability(ctx context.Context) error {
+	if !r.historicalBalanceCheckEnabled {
```
dhaidashenko: IMO, the multi-node should be the one to decide if we should poll this method or not.
If this method is called, we should always make a call

pkg/client/rpc_client.go
```diff
@@ -184,6 +189,53 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
...
+func (r *RPCClient) isFinalizedStateUnavailableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	pattern := r.clientErrors.FinalizedStateUnavailable()
+	if pattern == "" {
+		return false
+	}
+	re, compileErr := regexp.Compile(pattern)
+	if compileErr != nil {
```
dhaidashenko, line 237: Log critical error. Without a correct regexp, check if disabled. We need to be as loud as possible to indicate the issue, as system does not behave as configurer expects it to behave

pkg/client/rpc_client.go
Krish-vemula: Done. Removed the early return guard from `CheckFinalizedStateAvailabilit`. The method now always executes the `eth_getBalance` call when invoked. Also removed `HistoricalBalanceCheckEnabled` entirely multinode controls whether to call the method via `FinalizedStateCheckFailureThreshold > 0`

pkg/client/rpc_client.go
Krish-vemula: Done. Added a `Criticalw` log when the `FinalizedStateUnavailable` regex fails to compile:

`r.rpcLog.Criticalw("FinalizedStateUnavailable regex pattern is invalid; finalized state availability check is effectively disabled", "pattern", pattern, "err", compileErr)`
- Used `Critical `specifically because without a valid regex, the check silently stops classifying errors as `ErrFinalizedStateUnavailable`, meaning non-archive nodes would never be marked `FinalizedStateNotAvailable` the system misbehaves without any indication.

dhaidashenko (dismissed):

pkg/client/rpc_client.go
```diff
@@ -193,6 +196,50 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
...
+}
+
+// isFinalizedStateUnavailableError checks if the error matches the FinalizedStateUnavailable regex pattern.
+func (r *RPCClient) isFinalizedStateUnavailableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	pattern := r.clientErrors.FinalizedStateUnavailable()
+	if pattern == "" {
+		return false
```
amit-momin, line 234: Shouldn't we print an error here too? Seems like we'd only reach here if the configs to enable the `FinalizedStateUnavailable` check are set. If we can't effectively do that without the pattern, we'd want to make some noise. Ideally, we would be able to do this during startup during config validation but node pool errors can be set [here](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/client/errors.go) too so maybe that's not possible.

Krish-vemula: Good point! I looked into it and we actually can validate this at startup during the TOML config validation step.

I've updated `NodePool.ValidateConfig()` in `pkg/config/toml/config.go`. Now, if `FinalizedStateCheckFailureThreshold > 0` (meaning the feature is enabled), it will throw a startup error if `Errors.FinalizedStateUnavailable` is missing, empty, or an invalid regex.

I also added the critical log here in `rpc_client.go` just as a fallback, in case the error pattern somehow gets wiped out or bypassed dynamically at runtime.

pkg/client/rpc_client.go
```diff
@@ -193,6 +196,51 @@ func (r *RPCClient) ClientVersion(ctx context.Context) (version string, err erro
...
+}
+
+// isFinalizedStateUnavailableError checks if the error matches the FinalizedStateUnavailable regex pattern.
+func (r *RPCClient) isFinalizedStateUnavailableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	pattern := r.clientErrors.FinalizedStateUnavailable()
+	if pattern == "" {
+		r.rpcLog.Criticalw("FinalizedStateUnavailable regex pattern is empty; finalized state availability check is effectively disabled")
```
amit-momin: nit: think this can just be `Critical`
```suggestion
		r.rpcLog.Critical("FinalizedStateUnavailable regex pattern is empty; finalized state availability check is effectively disabled")
```

amit-momin (dismissed):

Unheilbar (approved):

dhaidashenko (approved):

jmank88 (approved):

PR #358, opened 2026-02-15
Add Data Feeds 1 OCR2 Aggregator Gethwrapper

cawthorne:
The OCR2 Aggregator currently in this repo seems to differ to this one:
https://github.com/smartcontractkit/gauntlet-evm/blob/develop/packages/evm-gauntlet-ocr/artifacts/evm/AccessControlledOCR2Aggregator.json

Adding it as a gethwrapper.


cll-gg (approved):

RensR (approved):
Approving with the comment that it is confusing to have two files that imply they would be the same but are now. Needs follow up with DF people

PR #360, opened 2026-02-16
bump common for LOOPP PPROF support

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3493

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/1834


jmank88:
Closing since we already bumped past this point.

PR #383, opened 2026-03-05
INC-2328 Add supportsDualBroadcast to txmv2

cll-gg:
### Description

Add `supportsDualBroadcast()` to txmgr. 
This is so that when a job for an SVR feed starts running, we validate whether the txm actually supports dual broadcast. If it doesn't, the job doesn't start. 

### Why

We encountered the following (INC-2328): a NOP had wrongly disabled Txmv2 but was still running an SVR feed. Secondary transactions for this feed were sent using txmv1 to the public mempool. 

### Requires Dependencies
https://github.com/smartcontractkit/chainlink-framework/pull/87

### Supports
https://github.com/smartcontractkit/chainlink/pull/21548

### DIFF
--- a/pkg/chains/legacyevm/evm_txm.go
+++ b/pkg/chains/legacyevm/evm_txm.go
@@ -48,13 +48,17 @@ func newEvmTxm(
 
 	if opts.GenTxManager == nil {
 		var txmv2 txmgr.TxManager
-		if cfg.Transactions().TransactionManagerV2().Enabled() {
+		txV2Cfg := cfg.Transactions().TransactionManagerV2()
+		dualBroadcastEnabled := txV2Cfg.Enabled() &&
+			txV2Cfg.DualBroadcast() != nil && *txV2Cfg.DualBroadcast() &&
+			txV2Cfg.CustomURL() != nil
+		if txV2Cfg.Enabled() {
 			txmv2, err = txmgr.NewTxmV2(
 				ds,
 				cfg,
 				txmgr.NewEvmTxmFeeConfig(cfg.GasEstimator()),
 				cfg.Transactions(),
-				cfg.Transactions().TransactionManagerV2(),
+				txV2Cfg,
 				client,
 				lggr,
 				logPoller,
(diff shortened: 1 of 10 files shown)


pkg/chains/legacyevm/evm_txm.go
yashnevatia: nit: can just send in txV2Cfg ?

cll-gg: True! Fixed

alejoberardino (dismissed):

pkg/txmgr/builder.go
```diff
@@ -150,7 +152,8 @@ func NewTxmV2(
 	}
 	var eh txm.ErrorHandler
 	var c txm.Client
-	if txmV2Config.DualBroadcast() != nil && *txmV2Config.DualBroadcast() && txmV2Config.CustomURL() != nil {
+	dualBroadcast := txmV2Config.DualBroadcast() != nil && *txmV2Config.DualBroadcast() && txmV2Config.CustomURL() != nil
```
dimriou: The output of this is only used once. There is no reason to store the output in another variable.

pkg/txmgr/builder.go
```diff
@@ -160,7 +163,8 @@ func NewTxmV2(
 		c = clientwrappers.NewChainClient(client)
 	}
 	t := txm.NewTxm(lggr, chainID, c, attemptBuilder, inMemoryStoreManager, stuckTxDetector, config, keyStore, eh)
-	return txm.NewTxmOrchestrator(lggr, chainID, t, inMemoryStoreManager, fwdMgr, keyStore, attemptBuilder), nil
+	o := txm.NewTxmOrchestrator(lggr, chainID, t, inMemoryStoreManager, fwdMgr, keyStore, attemptBuilder)
```
dimriou: Similar as above. Here we can cleanly return the result and not create a new variable.

pkg/txmgr/builder.go
cll-gg: I found it slightly easier to indicate meaning, but I'm also ok reverting it

pkg/txmgr/builder.go
cll-gg: True, done!

dimriou (approved):

alejoberardino (approved):

jmank88 (approved):

PR #386, opened 2026-03-09
Add logging around auction request and timeouts

augustbleeds:


dimriou:
This is only for testing purposes, so it's fine.

PR #387, opened 2026-03-11
Read multiplexer

dimriou:
This PR allows v2 client wrappers to perform multiple read calls for the pending and the latest nonce.
Every time a call is required, the client will make a call to the primary RPC. If it is successful, it will return the result. If not, it will store the error and fan out the request to all the remaining alive RPCs until it receives a successful response. If it does it returns with the value, otherwise it will return the collection of the failed requests. This change is important because it makes sure the confirmation logic will resume, even if the MultiNode has selected a bad RPC.


pkg/config/toml/docs.toml
```diff
@@ -171,6 +171,8 @@ BlockTime = '10s' # Example
 CustomURL = 'https://example.api.io' # Example
 # DualBroadcast enables DualBroadcast functionality.
 DualBroadcast = false # Example
+# MultiCall controls whether txm chain client reads use multiplexed calls.
+MultiCall = false # Example
```
cll-gg: I'd opt for a more specific name, to clearly indicate it's only for read calls, for example:
```suggestion
ReadRequestsToMultipleNodes = false # Example
```
(naming is hard, maybe you can come up with something better 😄 )

pkg/txm/clientwrappers/chain_client.go
```diff
@@ -2,30 +2,106 @@ package clientwrappers
...
+	c client.Client,
+	method string,
+	args []interface{},
+	decode DecodeMultiplexedResultFunc[T],
+) (result T, callDuration time.Duration, err error) {
+	ctx, cancel := context.WithTimeout(parentCtx, MultiCallMaxTimeout)
+	defer cancel()
+
+	startedAt := time.Now()
+	raw, err := c.CallContextAllSequential(ctx, method, args...)
```
cll-gg: This means the total of the sequential read calls have max 1.5s in total to respond, isn't that too low? We can start with this, but do we have an idea of the avg/p99 read call latency?

pkg/txm/clientwrappers/chain_client.go
```diff
@@ -2,30 +2,106 @@ package clientwrappers
...
+
+	return decoded, callDuration, nil
+}
+
+func GetTransactionCountMultiCall(parentCtx context.Context, c client.Client, lggr logger.SugaredLogger, address common.Address, blockTag string) (uint64, error) {
+	nonce, callDuration, err := multiCallSequential(
+		parentCtx,
+		c,
+		"eth_getTransactionCount",
+		[]interface{}{address, blockTag},
```
cll-gg: Nitpick
```suggestion
		[]any{}{address, blockTag},
```

pkg/txm/clientwrappers/chain_client.go
```diff
@@ -2,30 +2,106 @@ package clientwrappers
...
+			}
+
+			return uint64(nonce), nil
+		},
+	)
+	if err != nil {
+		return 0, err
+	}
+
+	lggr.Debugw("TransactionCount", "address", address, "nonce", nonce, "callDuration", callDuration)
```
cll-gg: Perhaps this to make it even more clear?
```suggestion
	lggr.Debugw("eth_getTransactionCount", "address", address, "nonce", nonce, "callDuration", callDuration)
```

pkg/txm/clientwrappers/chain_client.go
```diff
@@ -2,30 +2,106 @@ package clientwrappers
...
+		return GetTransactionCountMultiCall(ctx, c.c, c.lggr, address, "pending")
+	}
 	return c.c.PendingNonceAt(ctx, address)
 }
 
 func (c *ChainClient) SendTransaction(ctx context.Context, _ *types.Transaction, attempt *types.Attempt) error {
 	return c.c.SendTransaction(ctx, attempt.SignedTransaction)
 }
+
+type DecodeMultiplexedResultFunc[T any] func(raw json.RawMessage) (T, error)
```
cll-gg: Does it need to be exported?
Also, do we need this indirection + generics if `multiCallSequential` is only called once?

pkg/txm/clientwrappers/chain_client.go
```diff
@@ -2,30 +2,106 @@ package clientwrappers
...
+
+	decoded, decodeErr := decode(raw)
+	if decodeErr != nil {
+		return result, callDuration, fmt.Errorf("error decoding %s result: %w", method, decodeErr)
+	}
+
+	return decoded, callDuration, nil
+}
+
+func GetTransactionCountMultiCall(parentCtx context.Context, c client.Client, lggr logger.SugaredLogger, address common.Address, blockTag string) (uint64, error) {
```
cll-gg: Unexport?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -37,7 +37,7 @@ type FlashbotsTxStore interface {
 type FlashbotsClientRPC interface {
 	BlockByNumber(ctx context.Context, number *big.Int) (*evmtypes.Block, error)
 	NonceAt(context.Context, common.Address, *big.Int) (uint64, error)
-	SendTransaction(context.Context, *evmtypes.Transaction) error
+	SendTransaction(context.Context, *types.Transaction, *types.Attempt) error
```
cll-gg, line 40: Why is the extra param needed?

pkg/txm/clientwrappers/chain_client.go
dimriou: This is a bit of unknown, but here's my justification:
Across the Core node, our absolute minimum interval for each component is 2 seconds. Anything lower than that and you start having 100% utilization. For this case in particular, we need to come up with a number that is high enough but not too high so it adds additional latency for the components on chains with fast block time. Given how fast `eth_transactionCount` RPC calls are, I think 1.5s is decent enough to cycle through 3 RPCs (or even more) and not add potential latency. We have to take in consideration that if MultiNode is "locked" in a bad RPC, it means the max timeout we use might be consumed for an extended period of time, so we have to be careful with how high that value is.
Finally my thought process was this: I'd rather fail occasionally if this context is not enough than add latency that the node can't handle, the node would have failed without this anyway. I do agree that we should monitor this after we ship this change and see if this value is reasonable enough and revisit it if needed.

pkg/txm/clientwrappers/chain_client.go
dimriou: IMHO depending on how the multi read goes, we might have to extend this to other calls, i.e. BlockByNumber so I'd rather have this functionality supported by the chain client so it can easily be extended by others. It doesn't add much complexity anyway.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Ah yes, this is a bit complicated. When Flashbots and Meta clients are used by the TXM, they require additional information in the `SendTransaction` method (transaction & attempt). However, chain client and geth client can be used both as standalone clients for TXM but also as injected clients for Flashbots and Meta clients. So to make chain and geth clients compatible with both the TXM's expected interface but also Flashbots' injected client we need to make this change. Essentially, this pattern allows the TXM to be used inside and outside the Core node. A bit confusing, but I hope it helps 😅

pkg/txm/clientwrappers/chain_client.go
cll-gg: Ok, that makes sense. Good to know about the 2s minimum interval for each core node component. 

It's probably worthwhile to add some otel metric(s) to the code so that we can monitor how this change behaves after shipping it.

pkg/txm/clientwrappers/chain_client.go
cll-gg: Ok if it's not exported then that removes a lot of surface, then I'm ok with keeping it 👍

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Thanks for the explanation, I don't fully understand it but if we need it then we need it. 

> allows the TXM to be used inside and outside the Core node
Does this mean the integration tests?

dimriou: Integration tests, standalone testing, CCIP's next iteration, A-DON, etc.

pkg/client/chain_client.go
```diff
@@ -43,6 +45,9 @@ type Client interface {
 
 	// Wrapped RPC methods
 	CallContext(ctx context.Context, result interface{}, method string, args ...interface{}) error
+	// CallContextAllSequential calls CallContext for every single node sequentially and returns the first successful result (excluding sendonlys)
+	// along with the number of node calls attempted.
+	CallContextAllSequential(ctx context.Context, method string, args ...interface{}) (json.RawMessage, int, error)
```
dhaidashenko: I'd prefer this to be a set of dedicated methods instead of one generic:
* `PendingNonceAtAllSequential`
* `NonceAtAllSequential`

Exposing such generic methods increases the risk of leaking RPC-specific logic to the higher-level components/products and complicates the audit of methods that are actually needed for chain integration

dimriou: But `PendingNonceAtAllSequential` and `NonceAtAllSequential` also expose the MultiNode's functionality the same way, it's just they have dedicated names, no?

dhaidashenko (dismissed):

cll-gg (approved):

dhaidashenko (approved):

ilija42 (approved):

PR #391, opened 2026-03-16
OEV-1013: add logging and better metrics for hitting api

augustbleeds:
Couple changes here:
- Add logging statement when context deadline exceeds
- Change histogram bins to be more granular from 0-5s mark. 
- Introduce reporting "0xx" status codes which are attempts with no server response. "7" is context-deadline exceeded and "0" is catch all. This way we can track the number of attempts via the status code metric as well (because the latency metric could overcount / undercount the timeouts)
- Move context deadline closer to the start time measurement
- Start the latency timer before the context timer, so that we overestimate rather than underestimate the time. So that, if you were to pull the latencies > [timeout duration] you'd be overcounting rather than undercounting.

The most precise way to find the amount of context-deadline exceeded errors is by looking at the status codes.

### DIFF
--- a/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
@@ -316,9 +316,6 @@ type Metacalldata struct {
 }
 
 func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transaction, attempt *types.Attempt, dualBroadcastParams string, fwdrDestAddress common.Address) (*MetacalldataResponse, error) {
-	ctx, cancel := context.WithTimeout(parentCtx, a.auctionRequestTimeout)
-	defer cancel()
-
 	m := []byte{97, 116, 108, 97, 115, 95, 111, 101, 118, 65, 117, 99, 116, 105, 111, 110}
 
 	cid := hexutil.Uint64(a.chainID.Uint64())
@@ -357,22 +354,40 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 		return nil, fmt.Errorf("failed to marshal signed params: %w", err)
 	}
 	body := fmt.Appendf(nil, `{"jsonrpc":"2.0","method":"%s","params":[%s], "id":1}`, string(m), marshalledParamsExtended)
+
+	// Start timing for endpoint latency measurement
+	// Latency should be > than the context timer to query context-timeout requests
+	// (opt to overcount rather than undercount reqs with timeout)
+	startTime := time.Now()
+	ctx, cancel := context.WithTimeout(parentCtx, a.auctionRequestTimeout)
+	defer cancel()
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, a.customURL.String(), bytes.NewBuffer(body))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create POST request: %w", err)
 	}
 	req.Header.Add("Content-Type", "application/json")
 
-	// Start timing for endpoint latency measurement
-	startTime := time.Now()
 	resp, err := http.DefaultClient.Do(req)
+
 	latency := time.Since(startTime)
 
 	// Record latency
 	a.metrics.RecordLatency(ctx, latency)
+
 	if err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			a.lggr.Info("Auction Request Context Deadline Exceeded")
+			// mark status code "7" as context deadline exceeded
+			// definitive source of context-exceeded requests
+			a.metrics.RecordStatusCode(ctx, 7)
+		} else {
+			// mark status code "0" as all other errors to track the # of attempts
+			a.metrics.RecordStatusCode(ctx, 0)
+		}
+
 		return nil, fmt.Errorf("failed to send POST request: %w", err)
 	}
+
 	defer resp.Body.Close()
 
 	// Record status code
--- a/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
@@ -385,7 +385,7 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 			a.metrics.RecordStatusCode(ctx, 0)
 		}
 
-		return nil, fmt.Errorf("failed to send POST request: %w", err)
+		return nil, fmt.Errorf("failed to send POST request with latency %s: %w", latency, err)
 	}
 
 	defer resp.Body.Close()
--- a/pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
@@ -36,7 +36,11 @@ func NewMetaMetrics(chainID string, lggr logger.Logger) (*MetaMetrics, error) {
 		return nil, err
 	}
 
-	latencyHistogram, err := beholder.GetMeter().Int64Histogram("meta_endpoint_latency")
+	latencyHistogram, err := beholder.GetMeter().Int64Histogram("meta_endpoint_latency",
+		metric.WithUnit("ms"),
+		metric.WithDescription("Latency of Meta auction endpoint requests"),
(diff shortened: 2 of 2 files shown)


pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -357,22 +354,40 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 		return nil, fmt.Errorf("failed to marshal signed params: %w", err)
 	}
 	body := fmt.Appendf(nil, `{"jsonrpc":"2.0","method":"%s","params":[%s], "id":1}`, string(m), marshalledParamsExtended)
+
+	// Start timing for endpoint latency measurement
+	// Latency should be > than the context timer to query context-timeout requests
+	// (opt to overcount rather than undercount reqs with timeout)
+	startTime := time.Now()
+	ctx, cancel := context.WithTimeout(parentCtx, a.auctionRequestTimeout)
+	defer cancel()
```
dimriou, line 363: FWIW, if you want to measure the exact latency you should start counting right before the `resp, err := http.DefaultClient.Do(req)` and not here.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -357,22 +354,40 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
...
 	resp, err := http.DefaultClient.Do(req)
+
 	latency := time.Since(startTime)
 
 	// Record latency
 	a.metrics.RecordLatency(ctx, latency)
+
 	if err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			a.lggr.Info("Auction Request Context Deadline Exceeded")
```
dimriou, line 379: I don't think this log message makes sense here. If the context times out, the error message will return the exact same error, so we already have this information anyway. I'd rather enrich the error message with any additional info and return it or remove this log.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -357,22 +354,40 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
...
+			// mark status code "7" as context deadline exceeded
+			// definitive source of context-exceeded requests
+			a.metrics.RecordStatusCode(ctx, 7)
+		} else {
+			// mark status code "0" as all other errors to track the # of attempts
+			a.metrics.RecordStatusCode(ctx, 0)
```
dimriou, line 385: What is the current status code returned if the context times out?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: Speaking from my personal on-call experience looking through the current logs is quite tedious. You have to look at search for "Broadcast attempted" (so not only are you looking at transaction related logs when you just care about the auction process but "attempt" is a confusing term and you have to know how auction <> txs are related which the DF on-call presumably doesn't). I can also enrich it in the error information though, but I do not like relying solely on that because it's messy to look for in the logs. Lots of clicking back and forth.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: There is no status code returned because the error returns before status is captured

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: yeah, that's how it was done before, however the issue with that is that you will be undercounting (likely why i was seeing less timeouts using the latency metric). the timeouts. I explained in the PR description but this changes so that you overcount the timeouts. Overcounting is only going to be milliseconds so we're not going to risk flagging lots of healthy requests as context-exceeded , and if that happens there are deeper issues because "healthy" requests should not be taking 5s anyway.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: The current error message returned already includes that context deadline is exceeded, so we can just add latency metrics

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: I'm still not sure why we need to overcount. Latency is a metric of its own. If we want to pull the data for the `context timeout` failures we have another metric for that. Anyway, doesn't make that big of a difference so it's fine.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: > "attempt" is a confusing term and you have to know how auction <> txs are related which the DF on-call presumably doesn't).

If you're looking at the TXM or Client wrapper logs, you should be aware of what is a transaction and what is an attempt of a transaction. I'm a bit surprised these terms sound confusing. You can't expect to debug logs of a component you don't know how it functions. I feel like we already record metrics for this behaviour, so even if you read this error message on its own, you would still have to go back to the TXM logs to understand what happened and which transaction it affected, so why not read the message right away and just add the proper filters that you need?
I don't have a strong preference for this, but adding logs that replay information is a slippery slope and I want us to be mindful of it.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Fair enough. What would have been the status code if we didn't return. Maybe we can reuse that?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: Just being pragmatic here ~ if you ask the DF team (non-SVR) they do not know this yet they're expected to respond to SVR incidents. It's much easier for on-call to just search this string than to know that a post request to auction is followed after txm and they should be looking for a POST error with a wrapped error of context deadline exceeded in the "txEr__" field of the Broadcast attempted log.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: Yeah, it's just a nice to have. The true count of context deadline exceeded is from status codes. But if you can also see an accourate count from latency that's even better

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
augustbleeds: there is no HTTP status code because it never reaches the server in the first place. This is a pretty common practice to assign 0xx to internal systems. We did it at my last company with good outcomes ~  if you don't want to assign an internal status code you need to create a seperate metric. If you can have data in 1 metric it's always better than having to correlate that split across multiple.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: My point is, DFs are not expected to look at the logs during an incident because they don't know how the components work in the first place. There are runbooks for this. Only 2-3 people know what this log message means, so I don't understand how this would be much of a help to them since they already know how to get that information. Having said that, since this doesn't add too much noise and you feel this a QOL addition, I'm happy to keep it.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Yeah no objections there, I was just curious right now we have this:

```suggestion
	startTime := time.Now()
	resp, err := http.DefaultClient.Do(req)
	latency := time.Since(startTime)

	// Record latency
	a.metrics.RecordLatency(ctx, latency)
	if err != nil {
		return nil, fmt.Errorf("failed to send POST request: %w", err)
	}
	defer resp.Body.Close()

	// Record status code
	a.metrics.RecordStatusCode(ctx, resp.StatusCode)
```

If we didn't return after the err, what would the StatusCode be?

augustbleeds: resp is nil so res.StatusCode would create a panic

dimriou (approved):

amit-momin (approved):

PR #396, opened 2026-03-23
Add TransactionLifecycleID

dimriou:
This PR adds TransactionLifecycleID to txmv2 and enables it both for v1 and v2 for the dual broadcast relayer.

- Depends on: https://github.com/smartcontractkit/chainlink-framework/pull/95


pkg/transmitter/dual_contract_transmitter.go
```diff
@@ -242,6 +254,10 @@ func (oc *dualContractTransmitter) lockSecondary(ctx context.Context) error {
 	return nil
 }
 
+func generateTracingIDForOCR2(configDigest ocrtypes.ConfigDigest, epoch uint32, round uint8) string {
```
cll-gg: Could make sense to add a quick test for this function, just to be 100% sure

pkg/txm/types/transaction.go
```diff
@@ -106,6 +120,18 @@ func (t *Transaction) GetMeta() (*TxMeta, error) {
...
+func (t *Transaction) GetTracingID(lggr logger.SugaredLogger) string {
+	meta, err := t.GetMeta()
+	if err != nil {
+		lggr.Errorw("failed to get meta of the transaction", "err", err)
+		return ""
+	}
+	if meta != nil && meta.TracingID != nil {
+		return *meta.TracingID
+	}
+	return ""
```
cll-gg, line 119: Maybe `unknown` is clearer? It will differentiate it from the error case

pkg/txm/types/transaction.go
```diff
@@ -106,6 +120,18 @@ func (t *Transaction) GetMeta() (*TxMeta, error) {
 	return &m, nil
 }
 
+func (t *Transaction) GetTracingID(lggr logger.SugaredLogger) string {
```
cll-gg: Unfortunate that we have to pass the logger but I also don't see any better option

pkg/txm/txm.go
```diff
@@ -188,7 +188,10 @@ func (t *Txm) Trigger(address common.Address) {
 		if !exists {
 			return
 		}
-		triggerCh <- struct{}{}
+		select {
+		case triggerCh <- struct{}{}:
+		default:
+		}
```
cll-gg, line 194: Can't fully oversee the consequences of this change, can you explain this a bit?

pkg/txm/types/transaction.go
dimriou: I don't think unknown is consistent. I'd rather have it empty, or probably not log the field at all if it's nil in a future PR.

pkg/txm/types/transaction.go
dimriou: Yes, I struggled to find a good way to do this, but v1's GetMeta already does that so I just went with it.

pkg/txm/types/transaction.go
cll-gg: Fair, I do think it would be nice to differentiate between the unknown/unset situation and the error situation though

pkg/txm/txm.go
dimriou: Yes, this is the default way in go to not block the caller and just fill the channel and exit. Before we would (mistakenly) block after the buffer was filled. The old way may impact throughput on some rare cases, hence the change.

pkg/transmitter/dual_contract_transmitter.go
```diff
@@ -242,6 +254,10 @@ func (oc *dualContractTransmitter) lockSecondary(ctx context.Context) error {
 	return nil
 }
 
+func generateTracingIDForOCR2(reportTimestamp ocrtypes.ReportTimestamp) string {
+	return fmt.Sprintf("0x%s:%d:%d", reportTimestamp.ConfigDigest.Hex(), reportTimestamp.Epoch, reportTimestamp.Round)
```
augustbleeds, line 260: the config digest captures the aggregator address right?

augustbleeds (dismissed):
what's the main debugging path you'll use for this -- figuring out what happened to transmissions where primary succeeded but secondary failed?

pkg/transmitter/dual_contract_transmitter.go
dimriou: Not directly, but it's encapsulated.

dimriou:
> what's the main debugging path you'll use for this -- figuring out what happened to transmissions where primary succeeded but secondary failed?

Two things:
- being able to quickly debug logs for a given transmission end to end (both primary & secondary).
- understanding the parts of the code that are considered "checkpoints" for transmission and what information we can expose so we can create metrics/fire alerts in a following PR.

pkg/transmitter/dual_contract_transmitter.go
```diff
@@ -120,7 +120,12 @@ func (oc *dualContractTransmitter) Transmit(ctx context.Context, reportCtx ocrty
 		oc.lggr.Warnw("failed to generate tx metadata for report", "err", err)
 	}
 
-	oc.lggr.Debugw("Transmitting report", "report", hex.EncodeToString(report), "rawReportCtx", rawReportCtx, "contractAddress", oc.contractAddress, "txMeta", txMeta)
+	if txMeta == nil {
+		txMeta = &txmgr.TxMeta{}
+	}
+	tracingID := generateTracingIDForOCR2(reportCtx.ReportTimestamp)
+	txMeta.TracingID = &tracingID
+	oc.lggr.Infow("Transmitting report", "configDigest", "0x"+reportCtx.ReportTimestamp.ConfigDigest.Hex(), "epoch", reportCtx.ReportTimestamp.Epoch, "round", reportCtx.ReportTimestamp.Round, "contractAddress", oc.contractAddress, "txMeta", txMeta, "tracingID", tracingID)
```
jmank88: I'm worried about confusion with otel trace IDs. Is there a more specific name than `tracingID` that we could use?

pkg/txm/types/transaction.go
dimriou: Sure, but we shouldn't relay an error in an unrelated log field just because the underlying method happened to fail. That was the intention of the error log above. Each time you'll make the call you'll see the empty field but you'll also see the `failed to get meta of the transaction` message. If you feel there is a better way to handle this, happy to discuss.

pkg/transmitter/dual_contract_transmitter.go
dimriou: I'll change it to transactionLifecycleID

dimriou: But that would also require a change in chainlink-framework

pkg/txm/types/transaction.go
cll-gg: I'm not completely following this. Let's say we grep the logs for broadcasted transactions by querying e.g. `<logs> |= "Broadcasted attempt"`. We will then see a number of broadcasts with their tracing ID attached. Then we can do a new search with `<logs> | tracingID="<tracingID>"`.

However, if tracing ID is empty, then we don't know what happened: is it actually not set or were we unable to retrieve Meta? We won't see the `failed to get meta of the transaction` log line because we've already scoped the query to `Broadcasted attempt`. 

Anyway, I'm fine leaving it like this since there's a very low chance GetMeta() returns an error, this would only happen on malformed json.

dimriou: Yes, if you filter specifically with `<logs> |= "Broadcasted attempt"` you won't see it, but if you filter for the TXM logs, which is usually what we do, you'll be able to see a bunch of errors explaining the issue. If you're expecting a tracingID value and it's not there it's somewhat easy to infer what happened. Overall, I see the point, it's just I don't want to pollute the implementation with a custom handling for such an edge case, while we can retrieve the information by another log

dimriou: Also this might be eventually split up to individual fields i.e. `configDigest`, `epoch`, `round`

cll-gg: That's true. Ok let's leave it like this

pkg/transmitter/dual_contract_transmitter.go
jmank88: Thanks

jmank88 (approved):

cll-gg (approved):

PR #402, opened 2026-03-26
OEV-1150: add latest rpc nonce metric

augustbleeds:
Adds a ~local txm nonce metric and~ rpc nonce metric (~pending vs~ latest)

### DIFF
--- a/pkg/txm/metrics.go
+++ b/pkg/txm/metrics.go
@@ -40,14 +40,10 @@ var (
 		Name: "txm_time_until_tx_confirmed",
 		Help: "The amount of time elapsed from a transaction being broadcast to being included in a block.",
 	}, []string{"chainID"})
-	promLocalNonce = promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "txm_local_nonce",
-		Help: "The next nonce to be assigned by the TXM for a given address.",
-	}, []string{"chainID", "address"})
 	promRPCNonce = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "txm_rpc_nonce",
-		Help: "The latest nonce reported by the RPC node for a given address.",
-	}, []string{"chainID", "address", "source"})
+		Help: "The latest confirmed nonce reported by the RPC node for a given address.",
+	}, []string{"chainID", "address"})
 )
 
 type txmMetrics struct {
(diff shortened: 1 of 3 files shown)


pkg/txm/txm.go
```diff
@@ -157,6 +157,9 @@ func (t *Txm) initializeNonce(ctx context.Context, address common.Address) {
 			continue
 		}
 		t.SetNonce(address, pendingNonce)
+		if t.Metrics != nil {
+			t.Metrics.SetRPCNonce(ctxWithTimeout, address, pendingNonce, "pending")
+		}
 		t.lggr.Debugf("Set initial nonce for address: %v to %d", address, pendingNonce)
```
dimriou, line 160: Can you upgrade this log message to info? This should log once and it's quite valuable.

pkg/txm/metrics.go
```diff
@@ -40,6 +40,14 @@ var (
 		Name: "txm_time_until_tx_confirmed",
 		Help: "The amount of time elapsed from a transaction being broadcast to being included in a block.",
 	}, []string{"chainID"})
+	promLocalNonce = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
dimriou: Do we actually need prom metrics?

pkg/txm/txm.go
dimriou, line 1: IMHO, I don't find much value in tracking the local nonce. It is initialized once and gets incremented sequentially. You can also infer that information by looking at the transactions of the node. Similar case for the  pending nonce, which is used once for initialization and then there are clear log errors if you're getting throttled. I would keep the latest nonce value to understand which transactions are being confirmed and which are not.

augustbleeds: I see what you're saying, in theory the latest nonce is going to be much more valuable. Yeah, I'll remove it then.

pkg/txm/metrics.go
augustbleeds: Just keeping the pattern, but I think it's good practice if NOPs want to see the metrics

augustbleeds:
Being rate-limited by docker hub today so will need to retrigger these tests later for approval @dimriou

cll-gg (approved):
Approved after resolving https://github.com/smartcontractkit/chainlink-evm/pull/402/changes#r2999965752

dimriou (approved):

PR #408, opened 2026-04-02
Upgrade error handler

dimriou:
This PR upgrades error handler for TXMv2. We no longer throw an error if the handler decides there was no transmission, which results in a better observability. Additionally, if the error handler fails to parse the error, txm will check if the pending nonce increased instead of returning an error, which follows the standard flow.


pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
```diff
@@ -17,15 +17,25 @@ func NewErrorHandler() *errorHandler {
...
 
-func (e *errorHandler) HandleError(ctx context.Context, tx *types.Transaction, txErr error, txStore txm.TxStore, setNonce func(common.Address, uint64), isFromBroadcastMethod bool) error {
+func (e *errorHandler) HandleError(
+	ctx context.Context,
+	lggr logger.Logger,
+	tx *types.Transaction,
+	txErr error,
+	txStore txm.TxStore,
+	setNonce func(common.Address, uint64),
+	isFromBroadcastMethod bool,
```
cll-gg: Let's remove the unused `isFromBroadcastMethod` while we're touching this func anyway

pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
```diff
@@ -17,15 +17,25 @@ func NewErrorHandler() *errorHandler {
...
+) (noTransmission bool, err error) {
```
cll-gg, line 27: Nitpick: I somewhat dislike negations in var names, since it will lead to double negations, which can lead to confusion. 

Maybe flip it around and use something like `transmissionHappened`?

pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
```diff
@@ -17,15 +17,25 @@ func NewErrorHandler() *errorHandler {
...
 	// Mark the tx as fatal only if this is the first broadcast. In any other case, other txs might be included on-chain.
 	if (errors.Is(txErr, ErrNoBids) || errors.Is(txErr, ErrAuction)) && tx.AttemptCount == 1 {
 		if err := txStore.MarkTxFatal(ctx, tx, tx.FromAddress); err != nil {
-			return err
+			return false, err
 		}
 		setNonce(tx.FromAddress, *tx.Nonce)
-		return fmt.Errorf("transaction with txID: %d marked as fatal", tx.ID)
+		lggr.Infof("transaction with txID: %d marked as fatal", tx.ID)
```
cll-gg: Can we log the `transactionLifecycleID` here?

pkg/txm/txm.go
```diff
@@ -337,11 +339,17 @@ func (t *Txm) sendTransactionWithError(ctx context.Context, tx *types.Transactio
...
-		if err = t.errorHandler.HandleError(ctx, tx, txErr, t.txStore, t.SetNonce, false); err != nil {
-			return
+	if txErr != nil {
+		// If ErrorHandler is set, try to decide if there was a successful transmission by parsing the error message. If there wasn't, we return early.
+		if t.errorHandler != nil {
+			noTransmission, hErr := t.errorHandler.HandleError(ctx, t.lggr, tx, txErr, t.txStore, t.SetNonce, false)
+			if hErr != nil {
+				return hErr
+			}
+			if noTransmission {
```
cll-gg, line 349: Would it be helpful to add a debug log here?

pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
dimriou: Seems reasonable to me, but the issue with this particular case is that there is guarantee a transmission happened and we can only be sure about the `noTransmission`. If we flip the name to `transmissionHappened` or `transmissionMayHappened` and we check for the inversed result, the reader might assume that the error handler can deterministically returned if a transaction  was broadcasted. It feels like `noTransmission` provides better conviction as to what happened.

pkg/txm/txm.go
dimriou: For now the only ErrorHandler we have implemented is the MetaErrorHandler which already logs a message if `noTransmission` is true, so I think we already get the information from that.

pkg/txm/clientwrappers/dualbroadcast/meta_error_handler.go
cll-gg: Ah I see, so otherwise it would be `transmissionMightHaveHappened`, which is not great 😅 
Let's leave it like this then. Might be worthwhile to add a comment to the code about that we can only be 100% sure a transmission has _not_ happened

dimriou: I included one on the interface level here: https://github.com/smartcontractkit/chainlink-evm/pull/408/changes#diff-b9fe8b4b005efc03661d7777aebac2846a0ff01a18ffff7d1cb715d6766736f3R60-R61

cll-gg: Nice!

jmank88 (approved):

mchain0 (approved):

PR #410, opened 2026-04-07
OEV-1179 Multiplex to Nova RPC as a secondary

cll-gg:
## What

Send FB requests to Nova as well, treating it as fire-and-forget. 

## Why

To improve SVR.


pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
```diff
@@ -0,0 +1,49 @@
...
+		secondary: secondary,
+	}
+}
+
+func (m *multiplexClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	go func() {
+		// Use a background context so the secondary isn't cancelled when the primary returns.
+		// Each client is responsible for its own timeout (e.g. novaClient has novaRPCTimeout).
+		if err := m.secondary.SendTransaction(context.Background(), tx, attempt); err != nil {
+			m.lggr.Errorw("Secondary backend send failed", "err", err, "transactionLifecycleID", tx.GetTransactionLifecycleID(m.lggr))
```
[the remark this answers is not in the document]
cll-gg: We set the tx lifecycle id ourselves

pkg/client/chain_client_test.go
dimriou, line 1: Are these changes needed? chain_client should be irrelevant to this.

pkg/config/toml/testdata/config-full.toml
```diff
@@ -46,6 +46,7 @@ DetectionApiUrl = 'http://example.net'
 Enabled = false
 BlockTime = '42s'
 CustomURL = 'http://txs.org'
+CustomURLSecondary = 'http://txs-secondary.org'
```
dimriou, line 49: I'm on the fence with this one. What happens if we want to add another OFA? Maybe it makes sense to create a config that has all the secondary urls and it's a slice of URLs.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -34,22 +34,25 @@ type FlashbotsTxStore interface {
 	FetchUnconfirmedTransactions(context.Context, common.Address) ([]*types.Transaction, error)
 }
 
-type FlashbotsClientRPC interface {
```
dimriou, line 37: It's common pattern to name the local interfaces with the prefix of the component they're declared for. Otherwise you run the risk of conflicting names. For example if you add another client with a different interface in this package you'd have to figure out another vague name to not conflict.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
```diff
@@ -91,32 +95,35 @@ func (d *FlashbotsClient) SendTransaction(ctx context.Context, tx *types.Transac
 		return err
 	}
 
-	if meta != nil && meta.DualBroadcast != nil && *meta.DualBroadcast && !tx.IsPurgeable {
```
dimriou, line 94: This seems like an unrelated change. Please revert this.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
```diff
@@ -0,0 +1,50 @@
...
+		primary:   primary,
+		secondary: secondary,
+	}
+}
+
+func (m *multiplexClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	go func() {
+		// Use a background context so the secondary isn't cancelled when the primary returns.
+		// Each client is responsible for its own timeout (e.g. novaClient has novaRPCTimeout).
+		if err := m.secondary.SendTransaction(context.Background(), tx, attempt); err != nil {
```
dimriou: Passing `context.Background()` here is unsafe even if the client handles it under the hood. We should instead pass the parent context so in case of a cancel it gets propagated.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
```diff
@@ -0,0 +1,50 @@
...
+func newMultiplexClient(lggr logger.Logger, primary txm.Client, secondary txm.Client) *multiplexClient {
+	return &multiplexClient{
+		lggr:      logger.Sugared(logger.Named(lggr, "Txm.MultiplexClient")),
+		primary:   primary,
+		secondary: secondary,
+	}
+}
+
+func (m *multiplexClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	go func() {
```
dimriou, line 41: We should properly handle the lifecycle of this goroutine, otherwise we run the risk of memory leaks.

pkg/txm/clientwrappers/dualbroadcast/MULTIPLEX_CLIENT.md
dimriou, line 1: Was this AI generated? Feels like it replicates unnecessary information that will be hard to maintain if we make changes and will quickly get out of date. There is already a TXM doc, that we should have deleted that looks like this.

pkg/txm/clientwrappers/dualbroadcast/nova_client.go
dimriou, line 1: The two clients share a large amount of overlapping logic, minus the signing. To make this less DRY we should combine them under one client and instead inject signing where needed.

pkg/txm/clientwrappers/dualbroadcast/ofa_metrics.go
dimriou, line 1: Are metrics needed for Titan? Seems like we never needed them for Flashbots.

pkg/txm/txm.go
```diff
@@ -267,6 +267,9 @@ func (t *Txm) loop(address common.Address, triggerCh chan struct{}) {
 	}
 }
 
+// BroadcastTransaction broadcasts a transaction for the given address.
+// It returns true if the transaction was broadcasted successfully, false otherwise.
+// It returns an error if the transaction could not be broadcasted.
```
dimriou: This is not an accurate comment and might cause ambiguity. Either we would have to add a very detailed and accurate description of what BroadcasTransaction does and what are the different cases or remove it.

pkg/client/chain_client_test.go
cll-gg: Yeah I was surprised as well. The test failed [here](https://github.com/smartcontractkit/chainlink-evm/actions/runs/24134950333/job/70423046944?pr=410) but it seems the fix hasn't fully helped. I'll have another look

pkg/config/toml/testdata/config-full.toml
cll-gg: Yes I was on the same fence 🙂 
I'm not sure what's best, so I thought keeping it simplest is ok for now and if we introduce another url we can change it to a slice.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Ah, didn't know that pattern. 
Then I propose naming it `ofaClientRPC` or `ofaRPC` and reusing it in both flashbots + nova client. Agree?

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Yeah it's to keep flashbots_client and nova_client in sync, now both clients have an early return and less indentation in this function.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
cll-gg: Good point on the cancel! 
The problem is if the primary client runs into a timeout, then the secondary gets canceled as well, while we actually want to make it fire-and-forget. I'll have a look what we can best do here.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
cll-gg: True, I'll add that

pkg/txm/clientwrappers/dualbroadcast/MULTIPLEX_CLIENT.md
cll-gg: Yes, AI generated indeed. I assume that any agent that updates the code will also update this readme. But happy to put this context elsewhere, do you have a good idea?

(Is this the TXM doc you mean --> https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/txm/docs/TRANSACTION_MANAGER_V2.md?)

pkg/txm/clientwrappers/dualbroadcast/nova_client.go
cll-gg: Good idea, will do that!

pkg/txm/clientwrappers/dualbroadcast/ofa_metrics.go
cll-gg: I think that's because there's been only 1 client so far. Now that we add another one it's handy to be able to differentiate them.

pkg/txm/txm.go
cll-gg: I'd be in favor of adding a more accurate description, without any doc/comment I find it hard to reason about what the unnamed `bool` return var depicts.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: If we rename `FlashbotsClient` to `OFAClient` or something similar, then yes.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
dimriou: Given this runs in production and the change is irrelevant, since we'll be dropping the nova client, we should revert this.
Also, it's much better to enhance the readability of the code by adding some structure vs making the linter happy. Indentation isn't inherently bad.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
dimriou: I think in this case the context comes straight from the txm, which doesn't have a timeout unless Close() is called so it should be ok, but let's double check.

pkg/txm/clientwrappers/dualbroadcast/MULTIPLEX_CLIENT.md
dimriou: Personally I would put the details in the description of the `multiplex_cient`. But I would definitely remove the configs, especially since they are in a Core node format (yes the txm file is the one you linked).

pkg/txm/txm.go
dimriou: Fair enough, but this change is irrelevant to TXM so let's open another ticket instead so I can add an accurate description for the different cases of this, and the rest of the methods.

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Yeah this is changing anyway due to the merge of flashbots + nova client. 
(It wasn't triggered by the linter though, just by my consistency/readability concerns)

pkg/txm/clientwrappers/dualbroadcast/MULTIPLEX_CLIENT.md
cll-gg: Agree on the configs, they don't make sense in this context. 
I'll put some description in multiplex_client instead, good idea, closer to the code 👍

pkg/txm/txm.go
cll-gg: Ok sounds good

pkg/txm/clientwrappers/dualbroadcast/ofa_metrics.go
dimriou: Not that I'm against of adding metrics but I don't want us to fall into the trap of adding them and then never use them. Core had this problem. It's a simple change but it consumes certain resources and it's yet another thing in the codebase. Personally, I would add them in a follow up PR in case there is a need, but we can also discuss with the team if there is a use case i.e. monitoring Nova endpoint.

pkg/txm/clientwrappers/dualbroadcast/nova_client.go
cll-gg: Done

cll-gg: I've kept the fallback for sending to the public mempool logic in flashbots_client for now, seeing as we want to use nova_client as secondary. If nova_client would also do this we would run the risk of sending the tx to the public mempool twice. Not a big issue but still not great. Let me know what you think!

pkg/txm/clientwrappers/dualbroadcast/flashbots_client.go
cll-gg: Renamed to `ofaRpcClient`

pkg/client/chain_client_test.go
cll-gg: Reverted for now

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
cll-gg: Added

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
cll-gg: Fixed now, using the main context

pkg/txm/clientwrappers/dualbroadcast/ofa_metrics.go
cll-gg: Ok I'll deploy it to Staging and validate if we need them there

pkg/txm/clientwrappers/dualbroadcast/MULTIPLEX_CLIENT.md
cll-gg: Done

pkg/txm/txm.go
cll-gg: Removed

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
```diff
@@ -0,0 +1,62 @@
...
+		lggr:                 logger.Sugared(logger.Named(lggr, "Txm.MultiplexClient")),
+		primary:              primary,
+		secondary:            secondary,
+		secondarySendTimeout: rpcTimeout,
+	}
+}
+
+func (m *multiplexClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	go func() {
+		secondaryCtx, cancel := context.WithTimeout(ctx, m.secondarySendTimeout)
```
lhmerino, line 42: Should this derive from context.Background() instead of ctx? If the TXM caller passes a per-call context and it gets cancelled right after primary.SendTransaction returns, the secondary could get cancelled as well?

dimriou:
I think we're introducing too many clients with this approach. Originally, the idea was to add only one multiplexing client and then repurpose/rename flashbots client and use it for both nova & flashbots, perhaps via an enum. Then, specifically in case of the flashbots client, we would have a helper signature method to sign the necessary fields. Clientwrappers are already wrappers on top of existing client wrappers so I'd like us to be mindful of the level of encapsulation we use.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
dimriou: This is better, but we still need a waitgroup to properly handle this go routine, otherwise we would cause a potential memory leak. We should discuss if it's acceptable to wait for this call to return and not make it completely fire and forget. If we can't wait and we need to make it completely fire and forget, then we'd have to take another approach by closing the client, but I'd try to avoid it unless necessary.

pkg/txm/clientwrappers/dualbroadcast/multiplex_client.go
dimriou: Passing `context.Background()` here may cause potential memory leaks. In principle, the TXM caller always uses the app's context, which won't be canceled unless the app gets closed, in which case the call should be cancelled as well. It should be each client's responsibility to add their own timeout, by calling WithTimeout, but they should also be aware if the app is still running or not.

pkg/txm/clientwrappers/dualbroadcast/nova_client.go
dimriou: We need some product input as to whether we plan on using nova client as a standalone client.  It feels like we can avoid future changes if we can decide on that now.

cll-gg:
Closing in favor of https://github.com/smartcontractkit/chainlink-evm/pull/445

PR #417, opened 2026-04-08
"FeeBoost" : SVR Gas Strategy

augustbleeds:
Instead of a separate gas config, we opt for a simpler approach to just use the max fee each if the feeboost ie enabled on txmv2


pavel-raykov (dismissed):

pkg/txm/attempt_builder.go
```diff
@@ -69,9 +71,9 @@ func (a *attemptBuilder) NewBumpAttempt(ctx context.Context, lggr logger.Logger,
 }
 
 func (a *attemptBuilder) NewAgnosticBumpAttempt(ctx context.Context, lggr logger.Logger, tx *types.Transaction, dynamic bool) (attempt *types.Attempt, err error) {
-	// if the transaction is purgeable, NewAttempt will return the max fee instantly, so there is no need to bump
+	// if the transaction is purgeable or feeBoost is enabled, NewAttempt will return the max fee instantly, so there is no need to bump
 	attempt, err = a.NewAttempt(ctx, lggr, tx, dynamic)
-	if tx.IsPurgeable || err != nil {
+	if tx.IsPurgeable || a.feeBoost || err != nil {
```
dimriou: Since we have another case, I would split the error check from the rest, just for better clarity.

pkg/txm/attempt_builder.go
```diff
@@ -69,10 +71,15 @@ func (a *attemptBuilder) NewBumpAttempt(ctx context.Context, lggr logger.Logger,
...
 	attempt, err = a.NewAttempt(ctx, lggr, tx, dynamic)
-	if tx.IsPurgeable || err != nil {
-		return
+
+	if err != nil {
+		return nil, err
+	}
+
+	if tx.IsPurgeable || a.feeBoost {
+		return nil, nil
```
dimriou: This is incorrect. You want to return the attempt and the error instead, or just call return.

pkg/txm/attempt_builder.go
```diff
@@ -23,22 +23,24 @@ type attemptBuilder struct {
...
 	}
 }
 
 func (a *attemptBuilder) NewAttempt(ctx context.Context, lggr logger.Logger, tx *types.Transaction, dynamic bool) (*types.Attempt, error) {
 	var fee gas.EvmFee
 	var estimatedGasLimit uint64
 	var err error
-	if tx.IsPurgeable {
+	if tx.IsPurgeable || a.feeBoost {
 		fee, estimatedGasLimit, err = a.EvmFeeEstimator.GetMaxFee(ctx, tx.Data, a.emptyTxLimitDefault, a.priceMaxKey(tx.FromAddress), &tx.FromAddress, &tx.ToAddress)
```
dimriou: There is a bug here: GetMaxFee uses the `emptyTxLimitDefault` which is only correct in case of an empty transaction. We need to use `SpecifiedGasLimit` instead.

pkg/txm/attempt_builder.go
dimriou, line 1: Please add a test case in `attempt_builder_test.go`

dimriou (approved):

pavel-raykov (approved):

pkg/config/toml/docs.toml
```diff
@@ -177,6 +177,8 @@ ReadRequestsToMultipleNodes = false # Example
 Bundles = false # Example
 # FastlaneAuctionRequestTimeout configures the HTTP request timeout for Fastlane Atlas auction requests. Defaults to 5s if not set.
 FastlaneAuctionRequestTimeout = '5s' # Example
+# FeeBoost enables using GetMaxFee instead of GetFee for all TxM v2 transaction attempts. This makes SVR transactions as aggressive as possible with gas pricing. Only active when DualBroadcast is also enabled.
+FeeBoost = false # Example
```
amit-momin, line 181: nit: wouldn't this be the `default` instead of an example? Looks like even if it's nil, we treat it as false below

augustbleeds: I kept the example false here in the practice of showing the default / off behavior. Yes, it'll work with false or omission

augustbleeds: Oh I see what you mean (the comment)

amit-momin (approved):

PR #419, opened 2026-04-10
adding jovay's nonce too low regex

fernandezlautaro:


pkg/client/rpc_client.go
```diff
@@ -182,7 +183,18 @@ func NewRPCClient(
...
+		rpcURL = wsuri.String()
+	} else if httpuri != nil {
+		rpcURL = httpuri.String()
+	}
+	isSendOnly := tier == multinode.Secondary
+	var rpcBaseMetrics frameworkmetrics.RPCClientMetrics
+	if r.beholderMetrics != nil {
+		rpcBaseMetrics = r.beholderMetrics.rpcClientMetrics
+	}
+	r.RPCClientBase = multinode.NewRPCClientBase[*evmtypes.Head](cfg, QueryTimeout, lggr, r.latestBlock, r.latestFinalizedBlock, rpcURL, isSendOnly, rpcBaseMetrics)
```
fernandezlautaro, line 197: worst scenario, rpcBaseMetrics will be nil, which is ok as the code in framework is nil-safe, see https://github.com/smartcontractkit/chainlink-framework/blob/b2dbdd927efe88fdabd96142f4b9fa4a0497cf3b/multinode/rpc_client_base.go#L217

silaslenihan (approved):

Unheilbar (approved):

ilija42 (approved):

PR #420, opened 2026-04-13
Add configurable historical balance health check for finalized block

Krish-vemula:
Re-implement the finalized state availability check (originally PR #352) on top of the updated chainlink-framework that uses optional interfaces instead of modifying core NodeConfig/RPCClient interfaces. This avoids forcing unrelated chains (e.g. Solana) to implement boilerplate methods.

## Summary

Implements the EVM finalized-state availability check used by multinode.

The check verifies that an RPC can serve historical state at the latest finalized block by calling `BalanceAt` for a configurable probe address. The check is disabled by default and only runs when `FinalizedStateCheckFailureThreshold > 0`.

## Changes

- Added EVM config:
  - `HistoricalBalanceCheckAddress`
  - `FinalizedStateCheckFailureThreshold`
  - `Errors.FinalizedStateUnavailable`
- Default probe address is:
  - `0x0000000000000000000000000000000000000000`
- Implements `CheckFinalizedStateAvailability` on the EVM RPC client.
- Finalized block selection:
  - uses `finalized` tag when `FinalityTagEnabled = true`
  - uses `latest - FinalityDepth` when `FinalityTagEnabled = false`
- Calls `BalanceAt` at the selected finalized block.
- Maps configured historical-state errors to `multinode.ErrFinalizedStateUnavailable`.
- Adds config docs/defaults/test fixtures.


pkg/config/toml/config.go
```diff
@@ -1166,20 +1176,32 @@ func (p *NodePool) setFrom(f *NodePool) {
...
 }
 
 func (p *NodePool) ValidateConfig(finalityTagEnabled *bool) (err error) {
 	if finalityTagEnabled != nil && *finalityTagEnabled {
 		if p.FinalizedBlockPollInterval == nil {
 			err = multierr.Append(err, commonconfig.ErrMissing{Name: "FinalizedBlockPollInterval", Msg: "required when FinalityTagEnabled is true"})
-			return
-		}
-		if p.FinalizedBlockPollInterval.Duration() <= 0 {
+		} else if p.FinalizedBlockPollInterval.Duration() <= 0 {
```
dhaidashenko, line 1199: I understand why you've removed return. But not sure why you've added `else if` instead of `if`

pkg/config/toml/config.go
```diff
@@ -1166,20 +1176,32 @@ func (p *NodePool) setFrom(f *NodePool) {
...
 			err = multierr.Append(err, commonconfig.ErrInvalid{Name: "FinalizedBlockPollInterval", Value: p.FinalizedBlockPollInterval,
 				Msg: "must be greater than 0"})
 		}
 	}
+	if p.FinalizedStateCheckFailureThreshold != nil && *p.FinalizedStateCheckFailureThreshold > 0 {
+		if p.Errors.FinalizedStateUnavailable == nil || *p.Errors.FinalizedStateUnavailable == "" {
```
dhaidashenko: Follow the example of TooManyResults. Right now, you are not setting the default for FinalizedStateUnavailable

pkg/config/toml/config.go
Krish-vemula: I used `else if` so we never call `Duration()` when the interval is nil. A second standalone `if` would need an explicit `p.FinalizedBlockPollInterval != nil` guard; `else if` encodes that same dependency without duplicating the check.

pkg/config/toml/config.go
Krish-vemula: Added fix for this, can you verify if its good

pkg/client/rpc_client.go
```diff
@@ -83,6 +84,14 @@ var (
 
 const rpcSubscriptionMethodNewHeads = "newHeads"
 
+// defaultFinalizedStateUnavailableRe matches common RPC errors when historical state is unavailable
+// at the effective finalized block (e.g. pruned / non-archive nodes). Used when
+// [NodePool.Errors].FinalizedStateUnavailable is unset, analogous to built-in heuristics in IsTooManyResults.
+// Keep aligned with examples in pkg/config/toml/docs.toml and CONFIG.md.
+var defaultFinalizedStateUnavailableRe = regexp.MustCompile(
```
dhaidashenko: This is redundant. We have a fallback file, thus default value should always be available

pkg/config/toml/testdata/config-full.toml
```diff
@@ -147,6 +149,7 @@ Fatal = '(: |^)fatal'
 ServiceUnavailable = '(: |^)service unavailable'
 TooManyResults = '(: |^)too many results'
 MissingBlocks = '(: |^)invalid block range'
+FinalizedStateUnavailable = '(: |^)missing trie node'
```
dhaidashenko: fallback value does not match the regex defined in code

dhaidashenko (dismissed):

dhaidashenko (dismissed):

CONFIG.md
```diff
@@ -955,6 +955,8 @@ SyncThreshold = 5 # Default
 LeaseDuration = '0s' # Default
 NodeIsSyncingEnabled = false # Default
 FinalizedBlockPollInterval = '5s' # Default
+HistoricalBalanceCheckAddress = '0x0000000000000000000000000000000000000000' # Default
```
jmank88: Do we need a zero address default? Or can we put an example instead?

jmank88 (changes requested):

CONFIG.md
Krish-vemula: Yes, we need a runtime default of the zero address when the field is unset; that matches the task (probe historical state via BalanceAt at 0x000…000, overridable in config). We only care that state is available, not the balance.

jmank88: Sorry, why does that preclude using an example value in the docs? Nil checking is trivial, so we should prioritize the user experience here

Krish-vemula: Sounds good, switched it to # Example

pkg/config/chain_scoped_node_pool.go
```diff
@@ -42,6 +42,13 @@ func (n *NodePoolConfig) FinalizedBlockPollInterval() time.Duration {
 	return n.C.FinalizedBlockPollInterval.Duration()
 }
 
+func (n *NodePoolConfig) HistoricalBalanceCheckAddress() string {
+	if n.C.HistoricalBalanceCheckAddress == nil {
+		return ""
+	}
+	return n.C.HistoricalBalanceCheckAddress.String()
```
jmank88: Why don't we return the field here instead of formatting as string? The callers are just parsing it again.
```suggestion
func (n *NodePoolConfig) HistoricalBalanceCheckAddress() *types.EIP55Address {
	if n.C.HistoricalBalanceCheckAddress == nil {
		return nil // or explicit zero?
	}
	return n.C.HistoricalBalanceCheckAddress
```

Krish-vemula: Good catch - updated HistoricalBalanceCheckAddress() to return `*types.EIP55Address` and convert to common.Address once in rpc_client (nil → zero address).

jmank88 (approved):

amit-momin (approved):

PR #429, opened 2026-04-17
L-02 Fix false clearing of finality violation error in case of termproary …

dhaidashenko:
## Problem
Possible Clearing of finalityViolated Masks Unresolved Finality Violations

The PollAndSaveLogs function clears the finalityViolated latch whenever pollAndSaveLogs returns nil, but that helper returns nil on several non-success paths: failed latestBlocks or latestSafeBlock reads, failed InsertLogsWithBlocks writes, and even when there are no new blocks to process. As a result, after a genuine finality violation has been latched, a later soft failure can incorrectly clear that latch, emit a misleading "completed successfully" log, and cause Healthy to report no error even though no successful recovery poll has actually occurred.

The interaction between the backup and primary pollers can amplify this behaviour. Both run in the same select loop, and when a finality violation is detected by BackupPollAndSaveLogs and latched via finalityViolated, the primary poller (which runs far more frequently) may clear it on its very next tick. Thus, there is a risk of persistent oscillation where the true error state is visible only in brief windows and Healthy reports healthy for the majority of the time, even though the underlying violation has not been resolved.

## Fix
Clear finality violation error only once we successfully confirm that DB state matches RPC


mchain0:
it's a pr between two feature branches, not sure entirely sure if you need my (or anyone's) review. added @pavel-raykov since he probably has more context over this codebase

ilija42 (dismissed):

ilija42 (approved):

Unheilbar (approved):

PR #443, opened 2026-04-22
PLEX-2810

silaslenihan:
More details in Jira, this PR addresses the subtasks:
- [Adjust logging of batch call context to exclude results](https://smartcontract-it.atlassian.net/browse/PLEX-2811)
- [Increment finalized tx counter only when we actually finalize transaction](https://smartcontract-it.atlassian.net/browse/PLEX-2813)
         - Additionally I noticed that we only ever increment this counter by 1 rather than by the number of tx hashes, I also addressed that
- [Create a metric for transactions that are stuck in non terminal state](https://smartcontract-it.atlassian.net/browse/PLEX-2814)
- [Dynamically adjust batch size limit in case of error](https://smartcontract-it.atlassian.net/browse/PLEX-2817)

[Create a dashboard and alert for tx_manager_transactions_non_terminal](https://smartcontract-it.atlassian.net/browse/PLEX-2850) will follow once this code is released.

### DIFF
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -418,11 +418,12 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
 
 	ctx, cancel, client := r.makeLiveQueryCtxAndSafeGetClient(rootCtx, r.largePayloadRPCTimeout)
 	defer cancel()
-	lggr := r.newRqLggr().With("nBatchElems", len(b))
 
 	logElems := batchElemsForLog(b)
 
-	lggr.Tracew("RPC call: evmclient.Client#BatchCallContext", "batchElems", logElems)
+	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", logElems)
+
+	lggr.Tracew("RPC call: evmclient.Client#BatchCallContext")
 	start := time.Now()
 	err := r.wrapRPCClientError(client.rpc.BatchCallContext(ctx, b))
 	duration := time.Since(start)
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -369,6 +369,28 @@ func (r *RPCClient) CallContext(ctx context.Context, result interface{}, method 
 	return err
 }
 
+// batchElemLog is a logging-safe view of rpc.BatchElem that omits Result. After
+// BatchCallContext returns, Result fields may hold large payloads (for example
+// many eth_getTransactionReceipt responses); including them in logger context
+// massively increases log volume even when the batch fails.
+type batchElemLog struct {
+	Method string
+	Args   []interface{}
+	Error  error
+}
+
+func batchElemsForLog(b []rpc.BatchElem) []batchElemLog {
+	out := make([]batchElemLog, len(b))
+	for i := range b {
+		out[i] = batchElemLog{
+			Method: b[i].Method,
+			Args:   b[i].Args,
+			Error:  b[i].Error,
+		}
+	}
+	return out
+}
+
 func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem) error {
 	// Astar's finality tags provide weaker finality guarantees than we require.
 	// Fetch latest finalized block using Astar's custom requests and populate it after batch request completes
@@ -396,14 +418,16 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
 
 	ctx, cancel, client := r.makeLiveQueryCtxAndSafeGetClient(rootCtx, r.largePayloadRPCTimeout)
 	defer cancel()
-	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", b)
+	lggr := r.newRqLggr().With("nBatchElems", len(b))
 
-	lggr.Trace("RPC call: evmclient.Client#BatchCallContext")
+	logElems := batchElemsForLog(b)
+
+	lggr.Tracew("RPC call: evmclient.Client#BatchCallContext", "batchElems", logElems)
 	start := time.Now()
 	err := r.wrapRPCClientError(client.rpc.BatchCallContext(ctx, b))
 	duration := time.Since(start)
 
-	r.logResult(ctx, lggr, err, duration, r.getRPCDomain(), "BatchCallContext")
+	r.logResult(ctx, lggr.With("batchElems", logElems), err, duration, r.getRPCDomain(), "BatchCallContext")
 	if err != nil {
 		return err
 	}
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -423,12 +423,12 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
 
 	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", logElems)
 
-	lggr.Tracew("RPC call: evmclient.Client#BatchCallContext")
+	lggr.Trace("RPC call: evmclient.Client#BatchCallContext")
 	start := time.Now()
 	err := r.wrapRPCClientError(client.rpc.BatchCallContext(ctx, b))
 	duration := time.Since(start)
 
-	r.logResult(ctx, lggr.With("batchElems", logElems), err, duration, r.getRPCDomain(), "BatchCallContext")
+	r.logResult(ctx, lggr, err, duration, r.getRPCDomain(), "BatchCallContext")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/client/rpc_client.go
```diff
@@ -396,14 +418,16 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
 
 	ctx, cancel, client := r.makeLiveQueryCtxAndSafeGetClient(rootCtx, r.largePayloadRPCTimeout)
 	defer cancel()
-	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", b)
+	lggr := r.newRqLggr().With("nBatchElems", len(b))
```
dhaidashenko: nit: why not add logElems to loggr here

pkg/txmgr/evm_tx_store.go
```diff
@@ -1635,6 +1636,20 @@ func (o *evmTxStore) CountTransactionsByState(ctx context.Context, state txmgrty
 	return count, nil
 }
 
+// CountNonTerminalTransactions returns the number of transactions for the chain that are not in a
+// terminal state (finalized or fatal_error). Used for observability of backlog and stuck workflows.
+func (o *evmTxStore) CountNonTerminalTransactions(ctx context.Context, chainID *big.Int) (count uint32, err error) {
```
dhaidashenko: Just counting non-thermal transactions does not allow us to create a reliable alert.
Seeing a value above 0 does not mean that we have stuck transactions; it's possible that the load is even, and we constantly have transactions in flight.
An increase in the count can also occur in the happy path, when we just observe a tmp spike in the number of writes.

It might be better to track the max time (now() - created_at) for non-terminal transactions.

pkg/txmgr/finalizer.go
```diff
@@ -369,6 +388,56 @@ func (f *evmFinalizer) batchCheckReceiptHashesOnchain(ctx context.Context, block
...
 
+// receiptOuterBatchCap returns the max number of receipt RPCs to bundle for the given number of
+// attempts remaining, respecting rpcBatchSize config and any lowered effective size.
+func (f *evmFinalizer) receiptOuterBatchCap(remaining int) int {
+	if remaining < 1 {
+		return 1
+	}
+	v := int(f.receiptEffectiveBatchSize.Load())
+	if v > 0 {
+		if f.rpcBatchSize > 0 {
```
dhaidashenko: This complicates logic a lot. It might be better to reject an invalid batchSize on higher level (evmFinalizer creation/config)

pkg/txmgr/finalizer.go
```diff
@@ -369,6 +388,56 @@ func (f *evmFinalizer) batchCheckReceiptHashesOnchain(ctx context.Context, block
...
+}
+
+func (f *evmFinalizer) setReceiptEffectiveBatchSize(n int) {
+	if n < 1 {
+		n = 1
+	}
+	if f.rpcBatchSize > 0 && n > f.rpcBatchSize {
+		n = f.rpcBatchSize
+	}
+	prev := f.receiptEffectiveBatchSize.Load()
```
dhaidashenko: Do we need to persist the batch size across iterations? 
In theory, it could lead to performance degradation that's difficult to identify.
Imagine a case where one RPC of the pool is misconfigured to allow only 1 request per batch.

pkg/txmgr/finalizer.go
```diff
@@ -381,37 +450,74 @@ func (f *evmFinalizer) FetchAndStoreReceipts(ctx context.Context, head, latestFi
...
-		if fetchErr != nil {
-			errorList = append(errorList, fetchErr)
-			continue
+		offset := 0
+		subBatch := outerCap
+		if subBatch < 1 {
+			subBatch = 1
 		}
+		for offset < len(batch) {
+			chunkEnd := offset + subBatch
```
dhaidashenko: Why do we need sub batch? batch is already of a reduced size, no?

pkg/txmgr/evm_tx_store.go
silaslenihan: Ah ok, I thought we wanted to track runaway count but that makes more sense. Fixed.

pkg/txmgr/finalizer.go
silaslenihan: removed.

pkg/txmgr/finalizer.go
silaslenihan: simplified

pkg/client/rpc_client.go
```diff
@@ -396,14 +418,17 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
...
 	ctx, cancel, client := r.makeLiveQueryCtxAndSafeGetClient(rootCtx, r.largePayloadRPCTimeout)
 	defer cancel()
-	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", b)
 
-	lggr.Trace("RPC call: evmclient.Client#BatchCallContext")
+	logElems := batchElemsForLog(b)
+
+	lggr := r.newRqLggr().With("nBatchElems", len(b), "batchElems", logElems)
+
+	lggr.Tracew("RPC call: evmclient.Client#BatchCallContext")
```
amit-momin: nit: Think this should stay
```suggestion
	lggr.Trace("RPC call: evmclient.Client#BatchCallContext")
```

pkg/client/rpc_client.go
```diff
@@ -396,14 +418,17 @@ func (r *RPCClient) BatchCallContext(rootCtx context.Context, b []rpc.BatchElem)
...
 	start := time.Now()
 	err := r.wrapRPCClientError(client.rpc.BatchCallContext(ctx, b))
 	duration := time.Since(start)
 
-	r.logResult(ctx, lggr, err, duration, r.getRPCDomain(), "BatchCallContext")
+	r.logResult(ctx, lggr.With("batchElems", logElems), err, duration, r.getRPCDomain(), "BatchCallContext")
```
amit-momin: Since we're including log elems above, we'd want to remove them here right?
```suggestion
	r.logResult(ctx, lggr, err, duration, r.getRPCDomain(), "BatchCallContext")
```

pkg/txmgr/finalizer.go
```diff
@@ -381,33 +411,48 @@ func (f *evmFinalizer) FetchAndStoreReceipts(ctx context.Context, head, latestFi
...
+			receipts, fetchErr := f.batchFetchReceipts(ctx, chunk)
+			if fetchErr != nil {
+				if evmclient.IsBatchRPCResponseEnvelopeUnmarshalError(fetchErr) {
+					chunkLen := len(chunk)
+					if chunkLen <= 1 {
+						f.lggr.Criticalw("Receipt batch fetch failed with batch response envelope error at batch size 1; RPC may be misconfigured or overloaded",
+							"err", fetchErr, "txHash", chunk[0].Hash)
+						errorList = append(errorList, fetchErr)
+						i++
+						break
```
amit-momin: If we encounter this error, do we still want to continue with the outer loop? Guess if the RPC is just overloaded we're optimistically trying for other attempts but think more likely it's a misconfig and we'd be creating a lot of noise in the logs with crits for each attempt

pkg/txmgr/finalizer.go
```diff
@@ -381,33 +411,48 @@ func (f *evmFinalizer) FetchAndStoreReceipts(ctx context.Context, head, latestFi
...
+					}
+					chunkSize = max(1, chunkSize/2)
+					if sleepErr := sleepCtx(ctx, receiptBatchRPCEnvelopeRetryDelay); sleepErr != nil {
+						return sleepErr
+					}
+					f.lggr.Warnw("RPC returned a batch response envelope instead of a result array; reducing receipt fetch chunk size for this batch and retrying after delay",
```
amit-momin: Could you also add a notice to this log that the `RPCDefaultBatchSize` config needs to be adjusted for the chain? Might make it easier to triage the issue.

pkg/txmgr/finalizer.go
silaslenihan: Might be a better question for @dhaidashenko - my understanding is we do want to optimistically retry in case it was caused by a limit being reached. With the new metric we should have visibility into this and be able to address any config issues etc.

pkg/txmgr/finalizer.go
silaslenihan: Done

pkg/txmgr/finalizer.go
dhaidashenko: I'm ok with aborting fetching if the request fails with 1 batch size.

In theory, a new attempt could be routed to a different RPC, so we will get additional progress. But the noise argument also makes sense.

amit-momin: Another worry maybe is that if there is some limit we reached that doesn't just resolve between these batch call, we could be stuck retrying batch calls for a while. For an `RPCDefaultBatchSize=200`, we'd have 8 retries until we hit 1 then move onto the next receipt (i++). With the sleep of 250ms between each retry, we'd take 2s per receipt which could add up depending on the number of receipts waiting to be finalized. Since the Finalizer's run loop processes 1 head at a time, heads would probably start getting backed up. I think it'd be better to just fail out at this point and say we tried optimistically enough by reducing the batch to 1. If it fails at that point, just let processing retry on the next head.

dhaidashenko (approved):

amit-momin (approved):

ilija42 (approved):

PR #445, opened 2026-04-23
OEV-1179 Multiplex to Nova RPC as a secondary

cll-gg:
## What

Send FB requests to Nova as well, treating it as fire-and-forget. 

## Why

To improve SVR.


pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -1,45 +1,45 @@
...
-		mc, err := NewMetaClient(lggr, chainClient, keyStore, url, chainID, txStore, auctionRequestTimeout, lifecycleMetrics)
-		if err != nil {
-			return nil, nil, err
-		}
-		return mc, NewErrorHandler(), nil
+	return newMultiOfaClient(lggr, chainClient, keyStore, ofaURLs, chainID, txStore, bundles, auctionRequestTimeout)
+}
+
+// redactURL returns u as a string safe for logs: same redaction as url.URL.Redacted for userinfo, and api_key query values are replaced with "xxxxx". It does not mutate the original URL.
+func redactURL(u *url.URL) string {
```
dimriou: This seems auto generated by mistake. We should remove this.

pkg/txm/clientwrappers/dualbroadcast/selector_test.go
dimriou, line 1: If ofa client selection is transferred to multiofa_client then `selector.go` is barely a shin that calls two methods. We should remove this test file and instead test individually inside the multiofa_client.go the construction as these tests add to the maintenance cost with negligible returns.

pkg/config/config.go
```diff
@@ -132,6 +132,9 @@ type TransactionManagerV2 interface {
 	Enabled() bool
 	BlockTime() *time.Duration
 	CustomURL() *url.URL
+	// CustomOFAURLs returns the ordered OFA endpoints: index 0 is primary; additional indices are multiplexed secondaries.
+	// If CustomURLs is configured, it wins; otherwise a single legacy CustomURL is returned as a one-element slice.
+	CustomOFAURLs() []*url.URL
```
dimriou: We already have comments for this in the right place (configs). This file is just a helper, so we should remove the comments from this file to avoid duplication.

pkg/config/toml/docs.toml
```diff
@@ -167,8 +167,10 @@ MinAttempts = 3 # Example
 Enabled = false # Default
 # BlockTime controls the frequency of the backfill loop of TransactionManagerV2.
 BlockTime = '10s' # Example
-# CustomURL configures the base url of a custom endpoint used by the ChainDualBroadcast chain type.
+# CustomURL configures the base url of a custom endpoint used by the ChainDualBroadcast chain type (legacy single endpoint).
```
dimriou: `ChainDualBroadcast` ? How did that get there in the first place, let's update it.
```suggestion
# CustomURL configures the base url of a custom endpoint used by the dual broadcast functionality. (legacy single endpoint).
```

pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -1,45 +1,45 @@
...
-	url *url.URL,
-	chainID *big.Int,
-	txStore txm.TxStore,
-	readRequestsToMultipleNodes bool,
-	bundles *bool,
-	auctionRequestTimeout *time.Duration,
-	lifecycleMetrics txm.Metrics,
-) (txm.Client, txm.ErrorHandler, error) {
+// SelectClient builds the txm.Client for dual broadcast. ofaURLs must be non-empty; index 0 is the
+// primary (determines broadcast outcome and nonce queries). The implementation is always a multiOfaClient:
```
dimriou: nit: I think `determines broadcast outcome` is enough. I'd avoid the comment for nonce queries as it can easily be changed in the future and this comment might be forgotten, plus it doesn't give much context to people without codebase knowledge.

pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -1,45 +1,45 @@
...
+// it classifies each URL (Flashbots, Nova, Meta, …), builds one backend per entry, and with one URL has
```
dimriou: I'd also remove the comment about backend. Backend is abstracted at this level and reader is only aware of the client themselves and not how the operate under the hood.

pkg/txm/clientwrappers/dualbroadcast/selector.go
dimriou, line 1: I was going to say that we should simplify the comments, but I realized selector is no longer serving its original purpose. Can we consider deleting it and letting multiofa_client do the entire thing? I don't see a reason why we should keep it, it's barely a shin. Especially if we remove `redactURL`

dimriou: Either we remove the selector and rely on multiofa_client completely or meta client needs to be extracted from it and put it back to selector.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,178 @@
...
+	for i, u := range ofaURLs {
+		urlStrs[i] = redactURL(u)
+	}
+
+	secondaryLabels := make([]string, len(secondaries))
+	for i, secondary := range secondaries {
+		secondaryLabels[i] = secondary.Label()
+	}
+
+	lggr.Infow("TransactionManagerV2 OFA client created",
```
dimriou: nit: TransactionManagerV2 is abstracted from the clients at this level. Either remove or rename to something like `MultiOFA client created`

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-evm/pkg/keys"
+	"github.com/smartcontractkit/chainlink-evm/pkg/txm"
+	"github.com/smartcontractkit/chainlink-evm/pkg/txm/clientwrappers"
+	"github.com/smartcontractkit/chainlink-evm/pkg/txm/types"
+)
+
+// ofaBackend is used to broadcast to an OFA and read nonces.
+type ofaBackend interface {
```
dimriou: nit: To follow a consistent patter we should rename this to `multiOfaBackend` or `multiOfaBackendClient` and rename the `ofaTXClient` the same as the file suggests: `ofa_backend`. This way it's crystal clear what each interface/struct does.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,178 @@
...
+	PendingNonceAt(ctx context.Context, address common.Address) (uint64, error)
+	NonceAt(ctx context.Context, address common.Address, blockNumber *big.Int) (uint64, error)
+	SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error
+	Label() string
```
dimriou: We should remove `Label()` from the client as the other methods are already compatible with the native Geth methods and this might cause incompatibility issues in the future.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,178 @@
...
+
+// SendTransaction sends the transaction to the primary and secondaries, unless it is a non-dual-broadcast transaction.
+// In that case, it falls back to sending the transaction to the chain RPC directly.
+func (m *multiOfaClient) SendTransaction(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) error {
+	meta, err := tx.GetMeta()
+	if err != nil {
+		return err
+	}
+
+	// TODO(gg): check that this is the correct behavior for meta_client as well --> seems to not fully be the case, case #2 in meta_client is not handled here!
```
dimriou: If that's the case then we should move the meta client outside of this multiofa client back to selector?

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -0,0 +1,398 @@
...
+	bodyItems := make([]map[string]any, 0, len(attempts))
+	for _, attempt := range attempts {
+		txData, err := attempt.SignedTransaction.MarshalBinary()
+		if err != nil {
+			return fmt.Errorf("failed to marshal transaction for attempt ID %d: %w", attempt.ID, err)
+		}
+
+		bodyItems = append(bodyItems, map[string]any{
+			"tx":         hexutil.Encode(txData),
+			"revertMode": "allow",
```
dimriou: I had a comment here that was removed:
`// we always want to allow reverts so bundles are valid even if a single transaction within the bundle goes through`

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -0,0 +1,398 @@
...
+		nonce := attempt.SignedTransaction.Nonce()
+		nonces = append(nonces, nonce)
+		expectedNonce := prevNonce + 1
+		if nonce != expectedNonce {
+			return fmt.Errorf("bundle attempts must be contiguous and strictly increasing: expected nonce %d, got nonce %d", expectedNonce, nonce)
+		}
+		prevNonce = nonce
+	}
+
+	currentBlock, err := d.c.BlockByNumber(ctx, nil)
```
dimriou, line 295: There was a comment here about fetching the block each time. Reinstate the comment but remove the todo message so readers are aware.

pkg/txm/clientwrappers/dualbroadcast/selector.go
cll-gg: Why? Logging urls without api creds is handy for troubleshooting

pkg/txm/clientwrappers/dualbroadcast/selector_test.go
cll-gg: Now that we moved meta_client instantiation to selector.go, this test file contains only tests for selecting either meta_client or multiofa_client

pkg/txm/clientwrappers/dualbroadcast/selector.go
cll-gg: Agree, updated!

pkg/txm/clientwrappers/dualbroadcast/selector.go
cll-gg: Updated, please check

pkg/txm/clientwrappers/dualbroadcast/selector.go
cll-gg: I've done the latter

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: Done

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: Indeed, done

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Oh that's unexpected, good catch! 

I've put it back

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Done

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: I like it, makes the naming more consistent 👍 
Done!

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: Makes sense, done

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou, line 1: Let's avoid touching unrelated components. The PR is already big enough and it adds unnecessary complexity to the reviewers.

pkg/txm/clientwrappers/dualbroadcast/selector.go
```diff
@@ -1,45 +1,50 @@
...
 
-	urlString := url.String()
-	switch {
-	case strings.Contains(urlString, "flashbots"):
-		return NewFlashbotsClient(lggr, chainClient, keyStore, url, txStore, bundles), nil, nil
-	default:
-		mc, err := NewMetaClient(lggr, chainClient, keyStore, url, chainID, txStore, auctionRequestTimeout, lifecycleMetrics)
+	var mc txm.Client
+	// for fastlane atlas we support single URL only
+	if len(ofaURLs) == 1 && strings.Contains(ofaURLs[0].String(), "auctioneer") {
```
dimriou: I would prefer fastlane since it's in the domain name rather than auctioneer, although it's not blocking.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: I can remove the validation but I think the log message is helpful, since we're introducing more OFA clients with this PR

pkg/txm/clientwrappers/dualbroadcast/selector.go
cll-gg: Fine by me! Updated

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,178 @@
...
+		return err
+	}
+
+	// If not dual-broadcast, fall back to sending the transaction to the chain RPC directly.
+	if meta == nil || meta.DualBroadcast == nil || !*meta.DualBroadcast || tx.IsPurgeable {
+		return m.chainClient.SendTransaction(ctx, tx, attempt)
+	}
+
+	// send to secondaries in parallel, fire-and-forget: in case of error, log and continue
+	var wg sync.WaitGroup
```
dimriou: This will bound the primary transmission by the slowest secondary which is not what we want. If we want to do proper synchronizing we'd have to add a close method and update the interfaces. Since this is not a simple change, instead let's remove the waitgroup and make sure every context request is bound by secondarySendTimeout

cll-gg: Ok sounds good

dimriou: And also update the tests accordingly.

cll-gg: Obviously

cll-gg: Fixed & updated!

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -20,62 +20,107 @@ import (
...
 
-type FlashbotsClient struct {
+type ofaPostResponse struct {
+	Result json.RawMessage `json:"result,omitempty"`
+	Error  ofaPostError
+}
+
+type ofaPostError struct {
+	Message string `json:"message,omitempty"`
+}
```
dimriou: Not that important, but for future reference these don't have to use the ofa prefix because they're not part of the ofaBackend struct. They're helper structures of a method. It could actually bring more confusion if you add a prefix.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	// If not dual-broadcast, fall back to sending the transaction to the chain RPC directly.
+	if meta == nil || meta.DualBroadcast == nil || !*meta.DualBroadcast || tx.IsPurgeable {
+		return m.chainClient.SendTransaction(ctx, tx, attempt)
+	}
+
+	// Secondaries are best-effort and do not block the primary. Each call is bounded by
+	// secondarySendTimeout (or rpcTimeout if the field is unset/invalid).
+	secTimeout := m.secondarySendTimeout
```
dimriou: Don't store struct fields in temp variables, otherwise we would have passed it as method param.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -0,0 +1,179 @@
...
+	if secTimeout <= 0 {
+		secTimeout = rpcTimeout
+	}
```
dimriou: Here's why it's a bad practice to use temp variables:
In this case `rpcTimeout` is the const that initializes `m.secondarySendTimeout`. This check is invalid/redundant  based on the structure we've created.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -20,62 +20,107 @@ import (
...
+	case ofaNova:
+		return "nova"
+	default:
+		return "unknown"
+	}
 }
 
-type FlashbotsClientRPC interface {
+// chainRPCClient is used for public mempool fallback and reads shared by OFA implementations.
+type chainRPCClient interface {
```
dimriou: Similar to what we had before, this should be:

```suggestion
// ofaBackendClient is used for public mempool fallback and reads shared by OFA implementations.
type ofaBackendClient interface {
```

jmank88 (dismissed):

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: I don't fully understand: If we would name them postResponse and postError, would they not be too generic? Or that would be fine because these are in the `ofa_backend` file so they are implicitly considered to be helper methods for the ofa backend?

pkg/client/chain_client_test.go
```diff
@@ -1003,7 +1003,7 @@ func TestChainClient_NonceAtWithFallback(t *testing.T) {
 		require.Equal(t, uint64(10), nonce)
 		require.Equal(t, int32(1), mainCalls.Load())
 		require.Equal(t, int32(1), fastFallbackCalls.Load())
-		require.Less(t, duration, 100*time.Millisecond)
```
dimriou, line 1006: This line is my mistake. Please remove this line completely.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: Agree, this is redundant. We only have a separate `secondarySendTimeout` field on the struct so that we can override it in tests and don't have to wait the full `rpcTimeout` (10s). 
I've removed these lines but if you have a better way of overriding a default timeout in tests, lmk!

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: I think the files got mixed up. I was talking about the renames in `meta_error_handler.go`

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Happy to change it, but I don't fully grok this convention: having this interface being called `dualbroadcast.ofaBackendClient` sounds to me like it's a client for an OFA backend, while actually it is an "RPC client for the chain". Could you explain this convention more?

pkg/client/chain_client_test.go
cll-gg: Done

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: Correct. These structs could have been anonymous as well. They're not being exported and they're used for helper methods rather than struct-specific purposes. So in this case it's better if we use a more "localized" name vs what we do for the interfaces expected to be stored by the struct, which other components from other packages have to be compatible with.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Ah ok I can revert those, just did that because it makes the selector clearer. 
Reverted!

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Ok thanks, that helps! 
Renamed

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: It's not exported so it's not expected to be called through it's package, but if you feel it's still ambiguous you can rename it to `ofaBackendRPCCLient` although it might be an overkill.

> sounds to me like it's a client for an OFA backend, while actually it is an "RPC client for the chain".
It's the `ofaBackend`'s (RPC) client. The prefix dictates who the "owner" is. This is not a generic interface. This interface dictates the ofaBackend's contract with its client.

To give you some perspective, let's say in the `dualbroadcast` package you had 10 implementations: nova, flashbots, meta, etc. and every single of one of them needed to make some RPC calls, but not all of them require the exact calls so you can't declare a single interface for all of them. For example flashbots need a client with the call `BlockByNumber` but nova needs a client with `ChainID`. If you go with a generic name like chainRPCClient for flashbots when declaring its interface, then when you want to declare a slightly different interface for nova you would have to find another generic. What would that name be? You'd eventually flood the package with generic interface names. Instead what you could do is name the flashbots' expected client interface as flashbotsRPCClient, then nova's novaRPCClient, and so on.

cll-gg: Ok thanks for that context, helps a lot! I'll do the rename

jmank88 (approved):

dimriou (approved):

dhaidashenko (approved):

PR #462, opened 2026-05-06
Add gas tiering for multiOFA client

dimriou:
SVR needs an easy way to keep track which OFA included the transaction on-chain. To accomplish this we reserve the last two digits of the gas limit and we assign 10 values space for each one of them, for example:
- OFA1: 10 - 19
- OFA2: 20 - 29
... and so on.
For now only the first position is used, i.e. 10, 20, etc. However in the future we might want to include per-feed tiering across each OFA, so we reserve 10 spaces upfront.


pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +181,56 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+		GasTipCap: attempt.Fee.GasTipCap.ToInt(),
+		Data:      tx.Data,
+	}
+
+	signedTx, err := d.keystore.SignTx(ctx, tx.FromAddress, evmtypes.NewTx(&dynamicTx))
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign tiered attempt for txID: %v: %w", tx.ID, err)
+	}
+
+	d.lggr.Infow("Intercepting attempt for tx", "txID", tx.ID, "hash", signedTx.Hash(), "toAddress", tx.ToAddress, "gasLimit", tieredGasLimit,
```
cll-gg: Does this need to be info level? 
I'd rather have an info level log after the attempt has been sent to the OFA to print all the context

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -66,26 +65,31 @@ type postError struct {
 	Message string `json:"message,omitempty"`
 }
 
-// FlashbotsTxStore is required when bundle sending is enabled for Flashbots.
-type FlashbotsTxStore interface {
+// ofaBackendTxStore is required when bundle sending is enabled for Flashbots.
+type ofaBackendTxStore interface {
```
cll-gg, line 73: Nice, better naming indeed

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -38,10 +36,10 @@ type multiOfaClient struct {
 var _ txm.Client = (*multiOfaClient)(nil)
 
 // newMultiOfaClient builds backends from URLs: index 0 is primary (outcome and nonces); the rest are secondaries.
-func newMultiOfaClient(
+func NewMultiOfaClient(
```
cll-gg, line 39: Why export this?

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
```diff
@@ -38,10 +36,10 @@ type multiOfaClient struct {
 var _ txm.Client = (*multiOfaClient)(nil)
 
 // newMultiOfaClient builds backends from URLs: index 0 is primary (outcome and nonces); the rest are secondaries.
```
cll-gg:
```suggestion
// NewMultiOfaClient builds backends from URLs: index 0 is primary (outcome and nonces); the rest are secondaries.
```

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +181,56 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
 	return err
 }
 
+// createAttemptWithTiering creates a tiered attempt that has a different gas limit for each OFA backend, by reserving the last two digits.
+// Each OFA has 10 values spaced 10 units apart for the gas limit so there is room for feed-specific tiers per OFA in the future.
+// Unfortunately, the per-OFA attempts are not tracked by the txm. We can track them by looking at the individual logs per attempt stored.
+func (d *ofaBackend) createAttemptWithTiering(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) (*types.Attempt, error) {
```
cll-gg: I'd prefer to not use the keyword `tiering` here since this is more about tracking, and the term might get confused with feed tiering

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +181,56 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+	if tx.Nonce == nil {
+		return nil, fmt.Errorf("failed to create tiered attempt for txID: %v: nonce empty", tx.ID)
+	}
+	if !attempt.Fee.ValidDynamic() {
+		return nil, errors.New("attempt is not a dynamic fee attempt") // We could potentially support legacy transactions, but dynamic transactions are the default so this is ok.
```
cll-gg, line 195: Include the tx id here as well?

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +181,56 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+		"TipCap", attempt.Fee.GasTipCap, "FeeCap", attempt.Fee.GasFeeCap, "transactionLifecycleID", tx.GetTransactionLifecycleID(d.lggr), "ofa", d.ofa.name())
+
+	tieredAttempt := *attempt
+	tieredAttempt.Hash = signedTx.Hash()
+	tieredAttempt.GasLimit = tieredGasLimit
+	tieredAttempt.SignedTransaction = signedTx
+	return &tieredAttempt, nil
+}
+
+func (k ofa) tieredGasLimit(gasLimit uint64) (uint64, error) {
```
cll-gg: Would be good to have a table test for this function

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
dimriou: We need to be able to import this client for local testing, similar to the rest of the clients.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
dimriou: https://github.com/smartcontractkit/chainlink-evm/pull/462/changes#r3195011713

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: It was the name necessarily the problem, but we needed better abstraction so I could import the client locally for testing. That's why proper naming/abstraction is so crucial.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: We tend to keep txID for the more important logs so it's easily trackable, similar to what we did for transactionLifecycleID. This check is unlikely to fail.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: The existing log has inaccurate information of the transaction since it has the wrong details (gasLimit). This log is the only way to get the hash of the different OFAs accordingly so we can track them on-chain.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: There's already one [here](https://github.com/smartcontractkit/chainlink-evm/pull/462/changes#diff-1cef2547645c193641c948db4d1e4b5b6403057c7494c92647f32af928893d2cR470-R492) per OFA, it just checks the entire attempt construction.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: Any suggestions?

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: Also this error will bubble up and print the txID from a higher-level log.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +181,56 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+	suffix := (uint64(k) + 1) * 10
+	if suffix >= 100 {
+		return 0, fmt.Errorf("cannot calculate OFA gas limit tier for %s: suffix %d is not two digits", k.name(), suffix)
+	}
+	return gasLimit - gasLimit%100 + suffix, nil
```
lhmerino: I would suggest bumping the gas limit to the next hundred to ensure the new gas limit is always greater than or equal to the original gas limit for safety reasons.
Could be as simple as the following statement after applying the suffix change `tieredLimit := gasLimit - (gasLimit % 100) + suffix`:
```
if tieredLimit < gasLimit {
		tieredLimit += 100
}
```

lhmerino (dismissed):
Looks great, thank you!

I have just one safety comment regarding the gas limit.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
lhmerino: Following on, might also be nice to make `100` be a constant since its used in multiple locations as well.

alejoberardino (dismissed):

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: For me this links to the changes in the PR, I don't understand the explanation 😅

cll-gg: Nvm, it's answered below

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
cll-gg: Ah ok, clear. Can you point me to the way you do local testing for this? Is there a doc/some guidelines?

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Maybe `createAttemptWithAccountability` or `createAttemptForOFAAccountability`? 
We can add to the docs that we've reserved space to support feed tiering in the future.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: Ok nice

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
cll-gg: What if we would log this directly after a tx gets sent to an OFA? Then the hash and gasLimit would be correct and we wouldn't have to distinguish mentally between intercepted and non-intercepted attempts.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dhaidashenko: How about createAttemptWithTracedFee

dhaidashenko (dismissed):

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +184,61 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+		ChainID:   tx.ChainID,
+		Nonce:     *tx.Nonce,
+		To:        &tx.ToAddress,
+		Value:     tx.Value,
+		Gas:       tieredGasLimit,
+		GasFeeCap: attempt.Fee.GasFeeCap.ToInt(),
+		GasTipCap: attempt.Fee.GasTipCap.ToInt(),
+		Data:      tx.Data,
+	}
+
```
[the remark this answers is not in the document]
dimriou: This is no longer the case, keystore is passed in both cases.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +184,61 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+// createAttemptWithTiering creates a tiered attempt that has a different gas limit for each OFA backend, by reserving the last two digits.
+// Each OFA has 10 values spaced 10 units apart for the gas limit so there is room for feed-specific tiers per OFA in the future.
+// Unfortunately, the per-OFA attempts are not tracked by the txm. We can track them by looking at the individual logs per attempt stored.
+func (d *ofaBackend) createAttemptWithTiering(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) (*types.Attempt, error) {
+	if tx.Nonce == nil {
+		return nil, fmt.Errorf("failed to create tiered attempt for txID: %v: nonce empty", tx.ID)
+	}
+	if !attempt.Fee.ValidDynamic() {
+		return nil, errors.New("attempt is not a dynamic fee attempt") // We could potentially support legacy transactions, but dynamic transactions are the default so this is ok.
+	}
```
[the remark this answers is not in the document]
dimriou: There is no expectation to support legacy transactions in the future. It's the same concept as `meta_client`. If there is, we'll add it.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +184,61 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
...
+	tieredAttempt.SignedTransaction = signedTx
+	return &tieredAttempt, nil
+}
+
+func (k ofa) tieredGasLimit(gasLimit uint64) (uint64, error) {
+	suffix := (uint64(k) + 1) * 10
+	if suffix >= tieringSpace {
+		return 0, fmt.Errorf("cannot calculate OFA gas limit tier for %s: suffix %d is not in the range 0-%d", k.name(), suffix, tieringSpace-1)
+	}
+	tieredLimit := gasLimit - gasLimit%tieringSpace + suffix
```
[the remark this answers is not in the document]
dimriou: Having two OFAs of the same type makes no sense.

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
dimriou: I changed it to `createAttemptWithOFAGasTier` for better clarity.

pkg/txm/clientwrappers/dualbroadcast/multiofa_client.go
dimriou: We have the integration tests that can work with devnet and testnet [here](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/txm/integration-tests/README.md), but personally I still use my old [txmv2test](https://github.com/dimriou/txmv2test) repo as I have too many custom hacks that I'm too lazy to transfer to the integration tests.

dhaidashenko (approved):

pkg/txm/clientwrappers/dualbroadcast/ofa_backend.go
```diff
@@ -171,6 +180,61 @@ func (d *ofaBackend) sendDualBroadcastTx(ctx context.Context, tx *types.Transact
 	return err
 }
 
+// createAttemptWithOFAGasTier creates an OFA-specific gas tiered attempt that has a different gas limit for each OFA backend, by reserving the last two digits.
+// Each OFA has 10 values spaced 10 units apart for the gas limit so there is room for feed-specific tiers per OFA in the future.
+// Unfortunately, the per-OFA attempts are not tracked by the txm. We can track them by looking at the individual logs per attempt stored.
+func (d *ofaBackend) createAttemptWithOFAGasTier(ctx context.Context, tx *types.Transaction, attempt *types.Attempt) (*types.Attempt, error) {
+	if tx.Nonce == nil {
```
amit-momin: nit: should we validate tx and attempt are non-nil at the beginning for sanity?

pkg/txm/clientwrappers/dualbroadcast/ofa_backend_test.go
```diff
@@ -428,12 +454,36 @@ func TestNovaClient_SendTransaction_DualBroadcast(t *testing.T) {
...
+		want uint64
+	}{
+		{name: "flashbots", ofa: ofaFlashbots, want: 21110 + ofaGasTierSpace},
+		{name: "nova", ofa: ofaNova, want: 21120 + ofaGasTierSpace},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			backend := &ofaBackend{lggr: logger.Sugared(logger.Test(t)), ofa: tt.ofa, keystore: testKeyStore()}
+			tieredAttempt, err := backend.createAttemptWithOFAGasTier(context.Background(), tx, attempt)
```
amit-momin, line 472: nit: 
```suggestion
			tieredAttempt, err := backend.createAttemptWithOFAGasTier(t.Context(), tx, attempt)
```

pkg/txm/metrics.go
```diff
@@ -195,6 +195,10 @@ func (m *txmMetrics) SetRPCNonce(ctx context.Context, address common.Address, no
 }
 
 func (m *txmMetrics) EmitTxMessage(ctx context.Context, txHash common.Hash, fromAddress common.Address, tx *types.Transaction) error {
+	return EmitTxMessage(ctx, m.chainID.String(), txHash, fromAddress, tx)
+}
+
+func EmitTxMessage(ctx context.Context, chainID string, txHash common.Hash, fromAddress common.Address, tx *types.Transaction) error {
```
cll-gg, line 201: By extracting this into a package function we can't validate the correct call in unit tests. Is there an option to keep going through `txmMetrics` instead?

dimriou: I'm not following, can you expand on this?

cll-gg: Sure! At the moment in txm unit tests we can validate that the EmitTxMessage function is called in a certain way by mocking the metrics component and calling `metrics.AssertExpectations()`. When it becomes a package level function this is not possible anymore.

dimriou: Do we do that now?
I purposefully pass [NOOP metrics](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/txm/metrics.go#L134) in TXM unit tests  to decouple the two, as TXM unit tests should only test for TXM's functionality and abstract the rest. Unless I'm missing some other case, I feel like `metrics_test.go` already tests what we want.

cll-gg: I'm not sure if we do it now. This is what a unit test should do: test the component's functionality + verify that handoffs to other components happen correctly, i.e. the correct parameters are provided.

dimriou: Specifically for the unit tests it should be the opposite. We should test exactly what the component does. Anything that gets accepted as an interface via dependency injection is considered a contract for that service and ideally should be mocked, otherwise it doesn't provide the correct isolation. In this case, TXM's metrics is an interface so it makes sense to pass the NOOP metric structure.

dhaidashenko (approved):

pkg/txm/metrics.go
cll-gg: Discussed offline, clear on a way forward

amit-momin (approved):

PR #465, opened 2026-05-08
DF-23700 harmonize rpc client metrics

vlfig:
Pass the rpc domain to RPCClientBase, not the url, for consistency.

Bump `chainlink-framework/metrics`. Brings better labels and sanitization (https://github.com/smartcontractkit/chainlink-framework/pull/106)

### DIFF
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -180,18 +180,12 @@ func NewRPCClient(
 		lggr.Error("RPC client is configured with only WebSocket URL. If this CL Node serves external requests, it must also have an HTTP URL configured. Otherwise, there is a serious DDoS risk.")
 	}
 
-	var rpcURL string
-	if wsuri != nil {
-		rpcURL = wsuri.String()
-	} else if httpuri != nil {
-		rpcURL = httpuri.String()
-	}
 	isSendOnly := tier == multinode.Secondary
 	var rpcBaseMetrics frameworkmetrics.RPCClientMetrics
 	if r.beholderMetrics != nil {
 		rpcBaseMetrics = r.beholderMetrics.rpcClientMetrics
 	}
-	r.RPCClientBase = multinode.NewRPCClientBase[*evmtypes.Head](cfg, QueryTimeout, lggr, r.latestBlock, r.latestFinalizedBlock, rpcURL, isSendOnly, rpcBaseMetrics)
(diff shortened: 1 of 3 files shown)


pkg/client/rpc_client.go
```diff
@@ -180,18 +180,12 @@ func NewRPCClient(
...
-	} else if httpuri != nil {
-		rpcURL = httpuri.String()
-	}
 	isSendOnly := tier == multinode.Secondary
 	var rpcBaseMetrics frameworkmetrics.RPCClientMetrics
 	if r.beholderMetrics != nil {
 		rpcBaseMetrics = r.beholderMetrics.rpcClientMetrics
 	}
-	r.RPCClientBase = multinode.NewRPCClientBase[*evmtypes.Head](cfg, QueryTimeout, lggr, r.latestBlock, r.latestFinalizedBlock, rpcURL, isSendOnly, rpcBaseMetrics)
+	r.RPCClientBase = multinode.NewRPCClientBase[*evmtypes.Head](cfg, QueryTimeout, lggr, r.latestBlock, r.latestFinalizedBlock, r.getRPCDomain(), isSendOnly, rpcBaseMetrics)
```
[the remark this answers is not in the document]
vlfig: The theoretical panic risk is already there and it's not happening in reality because we're always called with at least one of `wsuri` or `httpuri` non-nil.

jmank88 (approved):

pkg/client/rpc_client.go
```diff
@@ -180,18 +180,12 @@ func NewRPCClient(
 		lggr.Error("RPC client is configured with only WebSocket URL. If this CL Node serves external requests, it must also have an HTTP URL configured. Otherwise, there is a serious DDoS risk.")
 	}
 
-	var rpcURL string
-	if wsuri != nil {
-		rpcURL = wsuri.String()
-	} else if httpuri != nil {
-		rpcURL = httpuri.String()
-	}
```
amit-momin, line 188: Seems like the `getRPCDomain` method reverses the priority of which type of RPC we want to record. Just making sure that's not an issue here

vlfig: Not an issue. Usually the wss and the http host is the same, and this order is the one that's used in the other metrics we're harmonizing with.

amit-momin (approved):

PR #468, opened 2026-05-13
Update Ronin configs after L2 Migration

dimriou:
Ronin underwent a hard fork which changes the network properties. The updated behaviour is similar to Optimism post-Bedrock. This PR updates the configs accordingly.

### DIFF
--- a/pkg/config/toml/defaults/Ronin_Mainnet.toml
+++ b/pkg/config/toml/defaults/Ronin_Mainnet.toml
@@ -1,11 +1,30 @@
 ChainID = '2020'
 FinalityTagEnabled = true
-# Ronin produces blocks every 3 seconds
-LogPollInterval = "3s"
-NoNewHeadsThreshold = "3m"
+LogPollInterval = "4s"
+MinIncomingConfirmations = 1
+# finality time ~25-30m
+NoNewFinalizedHeadsThreshold = '30m'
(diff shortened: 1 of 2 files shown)


dhaidashenko (approved):

pkg/config/toml/defaults/Ronin_Mainnet.toml
```diff
@@ -1,11 +1,30 @@
...
+NoNewFinalizedHeadsThreshold = '30m'
 LinkContractAddress = '0x3902228D6A3d2Dc44731fD9d45FeE6a61c722D0b'
 
 [GasEstimator]
-# Ronin uses default gas price of 20 gwei https://docs.skymavis.com/mavis/mpc/guides/estimate-gas#overview
-Mode = 'FeeHistory'
-PriceMax = "1000 gwei"
+EIP1559DynamicFees = true
+Mode = "FeeHistory"
+PriceMax = "120 gwei"
```
vlfig, line 12: New config mentioned in the war room had `1000 gwei`. Just making sure this is intentional.

vlfig (approved):
One comment just to be sure, but otherwise lgtm.

pkg/config/toml/defaults/Ronin_Mainnet.toml
dimriou: I missed this. You're right, we might have to increase that.

PR #469, opened 2026-05-13
[CRE-4317] Implement EVM ocr3types.OnchainKeyring2

pavel-raykov:
Requires https://github.com/smartcontractkit/chainlink-common/pull/2137


pkg/keys/v2/ocr3_onchain_keyring_test.go
```diff
@@ -0,0 +1,636 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-common/keystore"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+)
+
+var _ keystore.Storage = &TestMemoryStorage{}
+
+/*
```
kaleofduty, line 21: add comment with link to the offchain-reporting repo where these come from?

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,163 @@
...
+	// v is either 0 or 1; stores the recovery id of the underlying public key.
+	v := sig[64]
+	if v != 0 && v != 1 {
+		panic("v not in {0, 1}, cannot happen with cryptographic certainty")
+	}
+
+	result := make([]byte, 64)
+	copy(result, sig[:64])
+
+	// In the smart contracts, a value of v=0 is assumed.
```
kaleofduty: nit: insert link to smart contract?

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,163 @@
...
+// Uncompresses the given ECDSA signature in its external format (r, s') into the
+// format expected internally (r, s', v=0).
+// Note that uncompression does not guarantee that the resulting signature verifies
+// using the "github.com/ethereum/go-ethereum/crypto".VerifySignature function. Instead, it guarantees that
+// the correct public key is recovered using "github.com/ethereum/go-ethereum/crypto".SigToPub.
+func uncompressSignature(sig []byte) []byte {
+	return append(sig, 0)
+}
+
+func hashReport[RI any](configDigest ocrtypes.ConfigDigest, seqNr uint64, report ocr3types.ReportWithInfo[RI]) []byte {
```
kaleofduty: document domain separation from ocr2 hashing scheme?

pkg/keys/v2/ocr3_onchain_keyring_test.go
pavel-raykov: Done

pkg/keys/v2/ocr3_onchain_keyring.go
pavel-raykov: Done

pkg/keys/v2/ocr3_onchain_keyring.go
pavel-raykov: Done

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/keystore"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+)
+
+// CreateOCR3OnchainKeyring creates a new OCR3 onchain keyring.
+// Note that key names are prefixed with PrefixEVM and PrefixOCR2Onchain (to preserve compatibility with OCR2)
+// For example, a key named "test-key" will be stored at the path "evm/ocr2_onchain/test-key".
+func CreateOCR3OnchainKeyring[RI any](ctx context.Context, ks keystore.Keystore, keyringName string) (ocr3types.OnchainKeyring2[RI], error) {
```
kaleofduty: I am wondering whether we should (a) specialize this to RI being struct{} to make it more misuse resistant or (b) document the following caveat:
Products may embed safety-critical information affecting signature computation into RI. (E.g. marking reports as "specimen" that must not to be accepted onchain.) If we allow any RI, one could mistakenly use this keyring, but the safety-critical information would be silently discarded.

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,165 @@
...
+// format expected internally (r, s', v=0).
+// Note that uncompression does not guarantee that the resulting signature verifies
+// using the "github.com/ethereum/go-ethereum/crypto".VerifySignature function. Instead, it guarantees that
+// the correct public key is recovered using "github.com/ethereum/go-ethereum/crypto".SigToPub.
+func uncompressSignature(sig []byte) []byte {
+	return append(sig, 0)
+}
+
+// hashReport performs OCR3 report hashing. The main difference from OCR2 hashing (https://github.com/smartcontractkit/chainlink-evm/blob/b4068bf735e6e1af28602eece9e29a0bd31eed37/pkg/keys/v2/ocr2_onchain_keyring.go#L98)
+// is that we hash the sequence number `seqNr` instead of epoch||round (while also changing the hashing order).
```
kaleofduty:
```suggestion
// hashReport computes an EVM-friendly cryptographic hash function over configDigest, seqNr, and reportWithInfo.Report. 
// The main difference from OCR2 hashing (https://github.com/smartcontractkit/chainlink-evm/blob/b4068bf735e6e1af28602eece9e29a0bd31eed37/pkg/keys/v2/ocr2_onchain_keyring.go#L98)
// is that we hash the sequence number `seqNr` instead of `epoch||round||extraHash`.
// Domain separation is guaranteed because the result of this function is a Keccak hash computed over 96 bytes
// while the aforementioned OCR2 function is a Keccak hash computed over 128 bytes.
```

pavel-raykov: Done.

pkg/keys/v2/ocr3_onchain_keyring.go
pavel-raykov: Currently, for any types A and B we can create OnchainKeyring2[A] with CreateOCR3OnchainKeyring[A],  and then read it as OnchainKeyring2[B] with ListOCR3OnchainKeyrings[B]. Do we want to do something to prevent that? If not, I would rather restrict these methods to struct{}.

kaleofduty: I don't see a way to prevent the scenario you describe without forcing RI to a concrete type. But that would make this keyring unusable for plugins that make us of RI.

Therefore, I would lean towards documenting this behavior, but leaving as is.

pavel-raykov: Added a comment that RI acts as an interface boundary.

george-dorin (dismissed):

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/smartcontractkit/chainlink-common/keystore/ocr3util"
+)
+
+// CreateOCR3OnchainKeyring creates a new OCR3 onchain keyring.
+// Note that key names are prefixed with PrefixEVM and PrefixOCR2Onchain (to preserve compatibility with OCR2)
+// For example, a key named "test-key" will be stored at the path "evm/ocr2_onchain/test-key".
+// The generic RI parameter is not used anywhere in the implementation. It acts as a defensive interface boundary to
+// avoid signing a wrong type of reports. Note that it is not 100% bulletproof since one can also create a key with
+// CreateOCR3OnchainKeyring[A] and then read it with ListOCR3OnchainKeyrings[B].
+func CreateOCR3OnchainKeyring[RI any](ctx context.Context, ks keystore.Keystore, keyringName string) (ocr3types.OnchainKeyring2[RI], error) {
```
kaleofduty: @pavel-raykov , do we still want RI here? My understanding of our previous discussion was that we wanted consumers of the library to follow this pattern:
```
keyring := keys.CreateOCR3OnchainKeyring(ctx, ks, "test-key")
ocr3keyring := ocr3util.AsOCR3OnchainKeyring2IgnoringRI[whatver](keyring)
// pass ocr3keyring to libocr via OCR3OracleArgs2
```

pavel-raykov: Yes, RI is a part of the interface coming from OCR library, so the clients should have a convenient function to instantiate it (by itself RI-less interface should be useless to them)

pavel-raykov: Updated the function name with `IgnoringRI` suffix and updated the comment

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/smartcontractkit/chainlink-common/keystore"
+	"github.com/smartcontractkit/chainlink-common/keystore/ocr3util"
+)
+
+// CreateOCR3OnchainKeyringIgnoringRI creates a new OCR3 onchain keyring.
+// Note that key names are prefixed with PrefixEVM and PrefixOCR2Onchain (to preserve compatibility with OCR2).
+// For example, a key named "test-key" will be stored at the path "evm/ocr2_onchain/test-key".
+// Be careful! The generic RI parameter is not used anywhere in the implementation. It acts as a defensive interface boundary to avoid signing a wrong type of reports.
+// If you have security-critical information in RI, you should make your own ocr3types.OnchainKeyring2 that does use the information in RI for signing.
+func CreateOCR3OnchainKeyringIgnoringRI[RI any](ctx context.Context, ks keystore.Keystore, keyringName string) (ocr3types.OnchainKeyring2[RI], error) {
```
kaleofduty, line 27: 💙

pkg/keys/v2/ocr3_onchain_keyring.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return result
+}
+
+// Uncompresses the given ECDSA signature in its external format (r, s') into the
+// format expected internally (r, s', v=0).
+// Note that uncompression does not guarantee that the resulting signature verifies
+// using the "github.com/ethereum/go-ethereum/crypto".VerifySignature function. Instead, it guarantees that
+// the correct public key is recovered using "github.com/ethereum/go-ethereum/crypto".SigToPub.
+func uncompressSignature(sig []byte) []byte {
```
PhilippSchindler, line 158: To be on the safe side here, I would recommend allocating a new []byte instead. This avoids a potential footgun where `append(sig, 0)` to a backing array with capacity larger than 64 than actually clears a value in the backing array (rather than doing a fresh allocation).

```
out := make([]byte, 65)
copy(out, sig)
return out
```

pavel-raykov: Done.

PhilippSchindler (approved):
I've reviewed the cryptographic part of the implementation.

Overall verdict: Looks good — the implementation is correct and closely matches what we previously implemented in offchain-reporting ([PR #528](https://github.com/smartcontractkit/offchain-reporting/pull/528)). The (r, s) v-stripping scheme, the s to n-s normalization for v=0, the report hashing (incl. comment regarding domain separation), and Verify all check out.

One minor suggestion I had to remove a latent footgun (uncompressSignature) was already addressed. 

LGTM, great. 👍

george-dorin (approved):

PR #485, opened 2026-06-09
PLEX-3106: removing sensitive RPC info from logs

fernandezlautaro:


silaslenihan (dismissed):

dhaidashenko (dismissed):

pkg/client/rpc_client.go
```diff
@@ -331,15 +331,22 @@ func (r *RPCClient) String() string {
 	s := fmt.Sprintf("(%s)%s", r.tier.String(), r.name)
 	ws := r.ws.Load()
 	if ws != nil {
-		s = s + ":" + ws.uri.Redacted()
+		s = s + ":" + shortenURL(ws.uri)
```
amit-momin: Think it might be better to add two new fields to `RPCClient` like `shortenedWsUri` and `shortenedHttpUri`. That way we can just print it directly here and don't have to parse it on every call to String()

amit-momin (approved):

Unheilbar (approved):

PR #488, opened 2026-06-23
Sanitize rpc errors

DylanTinianov:
Remove URLs from rpc errors to avoid leaking API keys.

### DIFF
--- a/pkg/client/rpc_client.go
+++ b/pkg/client/rpc_client.go
@@ -1412,6 +1412,16 @@ func (r *RPCClient) prepareCallArgs(msg ethereum.CallMsg) interface{} {
 	return toBackwardCompatibleCallArgWithChainTypeSupport(msg, r.chainType)
 }
 
+// Matches URLs and captures only the scheme + host, allowing path segments (e.g. API keys) to be stripped via replacement with "$1/".
+var rpcURLPathRegexp = regexp.MustCompile(`(https?://[^/\s"']+)/[^"'\s]*`)
+
+func sanitizeRPCError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return errors.New(rpcURLPathRegexp.ReplaceAllString(err.Error(), "$1/"))
+}
+
 func (r *RPCClient) wrapRPCClientError(err error) error {
 	if err == nil {
 		r.rpcLog.Trace("Call succeeded")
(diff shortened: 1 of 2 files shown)


pkg/client/rpc_client.go
```diff
@@ -1412,6 +1412,16 @@ func (r *RPCClient) prepareCallArgs(msg ethereum.CallMsg) interface{} {
...
 
+// Matches URLs and captures only the scheme + host, allowing path segments (e.g. API keys) to be stripped via replacement with "$1/".
+var rpcURLPathRegexp = regexp.MustCompile(`(https?://[^/\s"']+)/[^"'\s]*`)
+
+func sanitizeRPCError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return errors.New(rpcURLPathRegexp.ReplaceAllString(err.Error(), "$1/"))
+}
```
amit-momin: Think it might be safer to just entirely remove any URLs from the error. We already log the specific (sanitized) node that had an error in the logger

amit-momin: Looks good now! Could you also update this to handle ws endpoints also?

pkg/client/rpc_client_internal_test.go
```diff
@@ -28,6 +29,37 @@ import (
...
+}
+
+func TestRPCClient_WrapRPCClientError_TimeoutIsWrappedAndSanitized(t *testing.T) {
+	t.Parallel()
+
+	rpcClient := NewTestRPCClient(t, RPCClientOpts{})
+
+	err := &url.Error{
+		Op:  "Post",
+		URL: "https://test-cl-03.test.xyz/dOnOtLeAkAnyOfThis/doNotLeak/test/mainnet/",
```
amit-momin: Once we add support for ws, we should update the test to also sanity check that it works for endpoints lik
- `http://secret.com/doNotLeak/`
- `ws://secret.com/doNotLeak/`
- `wss://secret.com/doNotLeak/`
- `https://user:pass@secret.com/`

DylanTinianov: Nice, added coverage for these as well

amit-momin (dismissed):

Unheilbar (dismissed):

amit-momin (approved):

Unheilbar (approved):

PR #489, opened 2026-06-25
Add AutomationReceiver to compilation scripts and generated wrappers

bitrider23:
This changes where made in order to deploy `AutomationReceiver` contract on CLD. 
## Summary

- Add `AutomationReceiver.sol` and its base contract `ReceiverTemplate.sol` to `contracts/src/v0.8/automation/` — a generic bridge that executes Automation-style upkeeps delivered by a CRE workflow
- Add `IReceiver.sol` interface defining the `onReport` callback
- Register `AutomationReceiver` in the compilation script (`native_solc_compile_all_automation`) using the `automation-compile-23` profile (solc 0.8.19)
- Add `go:generate` directive in `gethwrappers/go_generate_automation.go` and commit the generated Go binding at `gethwrappers/generated/automation_receiver/automation_receiver.go`

## Contract design

`AutomationReceiver` has two independent authorization layers:

1. **Inbound** (inherited from `ReceiverTemplate`) — validates the CRE Forwarder address plus optional `workflowId` / `workflowName` / `workflowOwner` checks
2. **Outbound** — a closed-by-default allowlist of `(target, selector)` pairs that the owner must explicitly enable before the contract can execute any call

### go-ethereum version alignment (`gethwrappers/go.mod`)
`gethwrappers/go.mod` had `go-ethereum v1.17.0` while the root `go.mod` already had `v1.17.3`. The CI builds `abigen` from scratch on a clean runner and the version mismatch caused it to skip the build entirely, resulting in:
Error: go-ethereum version (v1.17.0) does not match root package go-ethereum version (v1.17.3)
no such file or directory: tools/bin/abigen
Fixed by aligning `gethwrappers/go.mod` to `v1.17.3`.

### CVE fix — `go.opentelemetry.io/otel`
Upgrading `go-ethereum` pulled in `go.opentelemetry.io/otel v1.40.0`, which was flagged as high severity (remote DoS via multi-value baggage header parsing). Bumped to `v1.44.0` to resolve it.
### GETH_VERSION in wrapper dependency files
All `generated-wrapper-dependency-versions-do-not-edit.txt` files across the repo still tracked `GETH_VERSION: 1.17.0`. After rebuilding `abigen` with `v1.17.3`, the CI regenerated all wrappers and detected the mismatch via `git diff`, causing the `Check if GO solidity wrappers are updated` check to fail. Updated all files to reflect `GETH_VERSION: 1.17.3`. This was a pre-existing inconsistency in the repo exposed by this PR running on a clean CI runner without cached artifacts.


contracts/src/v0.8/automation/IERC165.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.4.0) (utils/introspection/IERC165.sol)
```
RensR: Why don't we simply use the OZ 5.1 refs we use everywhere else? No need to copy this file from OZ

RensR (dismissed):

ibrajer (dismissed):

gethwrappers/go_generate.go
```diff
@@ -4,8 +4,10 @@ package gethwrappers
 
 // Make sure solidity compiler artifacts are up-to-date. Only output stdout on failure.
 //go:generate ../contracts/scripts/native_solc_compile_all
+//go:generate ../contracts/scripts/native_solc_compile_all_automation-cre
```
RensR: This isn't required as it's specified in native_solc_compile_all

gethwrappers/go_generate_automation_cre.go
```diff
@@ -0,0 +1,3 @@
+package gethwrappers
+
+//go:generate go run ./generation/generate_automation_cre/wrap.go AutomationReceiver AutomationReceiver automation_receiver
```
RensR: Can you look at gethwrappers/l2ep/go_generate.go on how to do it the modern way? Automation is a bad example.

RensR (approved):

ibrajer (approved):

PR #497, opened 2026-07-09
Make ListOCR3OnchainKeyrings return a tuple OCR3OnchainKeyringWithPublicKey

kaleofduty:
Unlike its predecessors, the OCR3OnchainKeyring2 interface does not provide a way of discovering a keyring's public key. We expose the public key as an additional entry on the tuple to make the ListOCR3OnchainKeyrings API more ergonomic to use.

I have discussed this change with @pavel-raykov prior to creating the PR.

### DIFF
--- a/pkg/keys/v2/ocr3_onchain_keyring.go
+++ b/pkg/keys/v2/ocr3_onchain_keyring.go
@@ -20,7 +20,7 @@ import (
 )
 
 // OCR3OnchainKeyringWithPublicKey pairs an OCR3 onchain keyring with the public
-// key (an EVM address) that it signs with.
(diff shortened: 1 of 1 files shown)


pkg/keys/v2/ocr3_onchain_keyring.go
pavel-raykov: Let's also return OCR3OnchainKeyringWithPublicKey here.

pavel-raykov (dismissed):

pavel-raykov (dismissed):

pavel-raykov (approved):

george-dorin (approved):

PR #500, opened 2026-07-10
prevent array index out of range in feeHistoryEstimator

yashnevatia:


pkg/gas/fee_history_estimator.go
```diff
@@ -237,6 +237,10 @@ func (f *FeeHistoryEstimator) RefreshDynamicPrice() error {
 		return err
 	}
 
+	if len(feeHistory.BaseFee) == 0 {
+		return errors.New("fee history returned empty base fee")
```
amit-momin: Not sure if we want to return an error here. Think it might be better to just maintain the existing baseFee and continue this method. Otherwise, we'd skip updates to the priority fee also which we probably still need

pkg/gas/fee_history_estimator.go
```diff
@@ -240,7 +240,10 @@ func (f *FeeHistoryEstimator) RefreshDynamicPrice() error {
 	// eth_feeHistory doesn't return the latest baseFee of the range but rather the latest + 1, because it can be derived from the existing
 	// values. Source: https://github.com/ethereum/go-ethereum/blob/b0f66e34ca2a4ea7ae23475224451c8c9a569826/eth/gasprice/feehistory.go#L235
 	// nextBlock is the latest returned + 1 to be aligned with the base fee value.
-	nextBaseFee := assets.NewWei(feeHistory.BaseFee[len(feeHistory.BaseFee)-1])
+	nextBaseFee, err := f.getNextBaseFee()
```
amit-momin: nit: maybe need a comment here that we're ignoring nil base fee errors here so we don't block updating it to an actual value. Otherwise at a quick glance it might seem like something that was missed and could cause issues if someone adds an err check later

amit-momin (dismissed):

pkg/gas/fee_history_estimator.go
```diff
@@ -240,7 +240,12 @@ func (f *FeeHistoryEstimator) RefreshDynamicPrice() error {
 	// eth_feeHistory doesn't return the latest baseFee of the range but rather the latest + 1, because it can be derived from the existing
 	// values. Source: https://github.com/ethereum/go-ethereum/blob/b0f66e34ca2a4ea7ae23475224451c8c9a569826/eth/gasprice/feehistory.go#L235
 	// nextBlock is the latest returned + 1 to be aligned with the base fee value.
-	nextBaseFee := assets.NewWei(feeHistory.BaseFee[len(feeHistory.BaseFee)-1])
+	// If the fee history base fee list is empty, we use the cached next base fee to continue updating priorityFeeThresholdWei
+	// We are intentionally ignoring the nil base fee error here to avoid blocking a valid update of the next base fee
+	nextBaseFee, err := f.getNextBaseFee()
```
amit-momin: Sorry just realized that a nil nextBaseFee will cause problems with the `maxFeePerGas` further down. We may need to add some checks to maintain the existing `maxFeePerGas` too further down

pkg/gas/fee_history_estimator.go
```diff
@@ -240,7 +240,15 @@ func (f *FeeHistoryEstimator) RefreshDynamicPrice() error {
...
+	if err != nil {
+		return err
+	}
```
amit-momin: Don't think we can do this here otherwise we'll block updating the `nextBaseFee` to an actual value from nil. We may need to do something like 
1. fetch the existing value
2. try to set it using `feeHistory.BaseFee`
3. check if it's nil after then return an error

ilija42 (approved):

Unheilbar (approved):

PR #501, opened 2026-07-13
PLEX-3129 LP skip blocks

dhaidashenko:
Expose the API to force LP to fast-forward to a new finalized checkpoint.


pkg/logpoller/log_poller_internal_test.go
```diff
@@ -2147,3 +2148,148 @@ func BenchmarkFilter100_10(b *testing.B) {
...
+				b.FinalizedBlockNumber == anchorBlockNumber &&
+				b.SafeBlockNumber == anchorBlockNumber
+		}),
+		minBlockToPrune,
+	).Return(nil).Once()
+}
+
+func Test_skipToBlock(t *testing.T) {
+	t.Parallel()
+	ctx := testutils.Context(t)
```
amit-momin: nit: think this probably got copied over from other tests but think we can just use `t.Context()` now

pkg/logpoller/orm_test.go
```diff
@@ -2537,3 +2547,131 @@ func TestDSORM_DeleteLogsAndBlocksAfter_usesTransactionalDataSource(t *testing.T
...
+		require.NoError(t, err)
+		require.Equal(t, int64(10), b.BlockNumber)
+	})
+}
+
+func TestDSORM_StoreNewFinalizedCheckpoint_usesTransactionalDataSource(t *testing.T) {
+	t.Parallel()
+	testutils.SkipShortDB(t)
+
+	ctx := testutils.Context(t)
```
amit-momin: same here

amit-momin (dismissed):

pkg/logpoller/log_poller.go
```diff
@@ -1923,6 +1965,57 @@ func (lp *logPoller) DeleteLogsAndBlocksAfter(ctx context.Context, start int64)
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch anchor block %d from RPC: %w", anchorBlockNumber, err)
+	}
+
+	err = validateHead((*evmtypes.Head)(anchorHead))
+	if err != nil {
+		return fmt.Errorf("failed to validate anchor block %d: %w", anchorBlockNumber, err)
+	}
+
+	anchorBlock := finalizedHeadToBlock((*evmtypes.Head)(anchorHead))
```
ilija42: nit, this is getting casted twice, I'd just convert it first and change validateHead to validate the block struct that  `finalizedHeadToBlock` returns

pkg/logpoller/log_poller.go
```diff
@@ -1923,6 +1965,57 @@ func (lp *logPoller) DeleteLogsAndBlocksAfter(ctx context.Context, start int64)
...
+		return fmt.Errorf("aborted, SkipToBlock request cancelled: %w", ctx.Err())
+	}
+
+	latest, err := lp.orm.SelectLatestBlock(ctx)
+	if err != nil && !pkgerrors.Is(err, sql.ErrNoRows) {
+		return fmt.Errorf("failed to select latest block: %w", err)
+	}
+
+	minBlockToPrune := anchorBlockNumber
+	if latest != nil {
```
ilija42, line 2011: shouldn't an error be returned if latest finalized block can't be confirmed?

dhaidashenko: If the latest block is missing, it just means the DB is empty.
It can occur only if the main run has not ticked yet, which is unlikely. But even if it occurs, we'll save the finalized block as a checkpoint. Thus, the LP state will be valid.

pkg/logpoller/log_poller.go
dhaidashenko: Changing the type that validate Head accepts requires changes in all other places that use it, which creates a ripple effect. 
Fixed double cast by introduction of new method

amit-momin (dismissed):

ilija42 (dismissed):

ilija42 (approved):

amit-momin (approved):

PR #502, opened 2026-07-14
OEV-1414 Add feedAddress to meta metrics

cll-gg:
## What

Add feedAddress to meta client related metrics

## Why

To improve observability

### DIFF
--- a/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
@@ -331,10 +331,6 @@ type Metacalldata struct {
 }
 
 func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transaction, attempt *types.Attempt, txMeta *types.TxMeta, fwdrDestAddress common.Address) (*MetacalldataResponse, error) {
-	if txMeta == nil || txMeta.FwdrDestAddress == nil || txMeta.DualBroadcastParams == nil {
-		return nil, errors.New("missing tx meta fields for dual broadcast request")
-	}
-
 	m := []byte{97, 116, 108, 97, 115, 95, 111, 101, 118, 65, 117, 99, 116, 105, 111, 110}
 
 	cid := hexutil.Uint64(a.chainID.Uint64())
--- a/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
+++ b/pkg/txm/clientwrappers/dualbroadcast/meta_client.go
@@ -205,7 +205,7 @@ func (a *MetaClient) SendTransaction(ctx context.Context, tx *types.Transaction,
 		// Auction & Validate
 		meta, err := a.SendRequest(ctx, tx, attempt, txMeta, tx.ToAddress)
 		if err != nil {
-			a.metrics.RecordSendRequestError(ctx, txMeta.FwdrDestAddress.Hex())
+			a.metrics.RecordSendRequestError(ctx, *txMeta.FwdrDestAddress)
 			a.metrics.emitAtlasError(ctx, "send_request", a.customURL, err, tx)
 			a.lifecycleMetrics.IncrementLifecycleFailure(ctx, txm.StageAuction)
 			return fmt.Errorf("error sending request for transactionID(%d): %w", tx.ID, errors.Join(err, ErrAuction))
@@ -213,7 +213,7 @@ func (a *MetaClient) SendTransaction(ctx context.Context, tx *types.Transaction,
 		// Send Metacall
 		if meta != nil {
 			if err := a.SendOperation(ctx, tx, attempt, *meta); err != nil {
-				a.metrics.RecordSendOperationError(ctx, txMeta.FwdrDestAddress.Hex())
+				a.metrics.RecordSendOperationError(ctx, *txMeta.FwdrDestAddress)
 				a.metrics.emitAtlasError(ctx, "send_operation", a.customURL, err, tx)
 				return fmt.Errorf("failed to send operation for transactionID(%d): %w", tx.ID, err)
 			}
@@ -392,17 +392,17 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 	latency := time.Since(requestStartTime)
 
 	// Record latency
-	a.metrics.RecordLatency(ctx, latency, txMeta.FwdrDestAddress.Hex())
+	a.metrics.RecordLatency(ctx, latency, *txMeta.FwdrDestAddress)
 
 	if err != nil {
 		if ctx.Err() == context.DeadlineExceeded {
 			a.lggr.Info("Auction Request Context Deadline Exceeded")
 			// mark status code "7" as context deadline exceeded
 			// definitive source of context-exceeded requests
-			a.metrics.RecordStatusCode(ctx, 7, txMeta.FwdrDestAddress.Hex())
+			a.metrics.RecordStatusCode(ctx, 7, *txMeta.FwdrDestAddress)
 		} else {
 			// mark status code "0" as all other errors to track the # of attempts
-			a.metrics.RecordStatusCode(ctx, 0, txMeta.FwdrDestAddress.Hex())
+			a.metrics.RecordStatusCode(ctx, 0, *txMeta.FwdrDestAddress)
 		}
 
 		return nil, fmt.Errorf("failed to send POST request with latency %s: %w", latency, err)
@@ -411,7 +411,7 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 	defer resp.Body.Close()
 
 	// Record status code
-	a.metrics.RecordStatusCode(ctx, resp.StatusCode, txMeta.FwdrDestAddress.Hex())
+	a.metrics.RecordStatusCode(ctx, resp.StatusCode, *txMeta.FwdrDestAddress)
 
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("request %v failed with status: %d", req, resp.StatusCode)
@@ -431,7 +431,7 @@ func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transactio
 	if response.Error.ErrorMessage != "" {
 		a.metrics.emitAtlasUserOp(ctx, response.Error.Data.UserOpHash, tx, requestStartTime.UnixMicro(), responseReceivedAt)
 		if strings.Contains(response.Error.ErrorMessage, NoSolverOps) || strings.Contains(response.Error.ErrorMessage, NoSolverOpsAfterSimulation) {
-			a.metrics.RecordBidsReceived(ctx, 0, txMeta.FwdrDestAddress.Hex())
+			a.metrics.RecordBidsReceived(ctx, 0, *txMeta.FwdrDestAddress)
 			return nil, nil
 		}
 		return nil, errors.New(response.Error.ErrorMessage)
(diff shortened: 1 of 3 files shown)


pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -68,49 +69,54 @@ func NewMetaMetrics(chainID string, lggr logger.Logger) (*MetaMetrics, error) {
 }
 
 // RecordStatusCode records the HTTP status code from Meta endpoint
-func (m *MetaMetrics) RecordStatusCode(ctx context.Context, statusCode int) {
+func (m *MetaMetrics) RecordStatusCode(ctx context.Context, statusCode int, feedAddress string) {
 	m.statusCodeCounter.Add(ctx, 1,
 		metric.WithAttributes(
 			attribute.String("chainID", m.chainID),
 			attribute.String("statusCode", strconv.Itoa(statusCode)),
+			attribute.String("feedAddress", strings.ToLower(feedAddress)),
```
dimriou: You can pass `feedAddress` as a `common.Address` param and avoid spamming txMeta.FwdrDestAddress.Hex() in the caller and also avoid `strings.ToLower(feedAddress)` since all callers will be consistent with `.Hex()` call
```suggestion
func (m *MetaMetrics) RecordStatusCode(ctx context.Context, statusCode int, feedAddress common.Address) {
m.statusCodeCounter.Add(ctx, 1,
metric.WithAttributes(
attribute.String("chainID", m.chainID),
attribute.String("statusCode", strconv.Itoa(statusCode)),
			attribute.String("feedAddress", feedAddress.Hex()),
```

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -68,49 +69,54 @@ func NewMetaMetrics(chainID string, lggr logger.Logger) (*MetaMetrics, error) {
...
 		),
 	)
 }
 
 // RecordLatency records the latency of Meta endpoint requests
-func (m *MetaMetrics) RecordLatency(ctx context.Context, duration time.Duration) {
+func (m *MetaMetrics) RecordLatency(ctx context.Context, duration time.Duration, feedAddress string) {
```
dimriou: Same comment as above.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -68,49 +69,54 @@ func NewMetaMetrics(chainID string, lggr logger.Logger) (*MetaMetrics, error) {
...
 		metric.WithAttributes(
 			attribute.String("chainID", m.chainID),
+			attribute.String("feedAddress", strings.ToLower(feedAddress)),
 		),
 	)
 }
 
 // RecordBidsReceived records the distribution of bids per transaction
-func (m *MetaMetrics) RecordBidsReceived(ctx context.Context, bidCount int) {
+func (m *MetaMetrics) RecordBidsReceived(ctx context.Context, bidCount int, feedAddress string) {
```
dimriou: Same as the other two.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: Ah yes, passing address is a lot nicer 👍 
Will update.

> also avoid strings.ToLower(feedAddress) since all callers will be consistent with .Hex() call

WDYM with this? 
The toLower call is to prevent mixed-case addresses from being labels in PromQL, because PromQL doesn't have a `lowercase` function, making it impossible to match a mixed-case address with a lowercased address.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: Fixed

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: Fixed

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -194,27 +193,27 @@ func (a *MetaClient) PendingNonceAt(ctx context.Context, address common.Address)
...
 
 	// #1
-	if meta != nil &&
-		meta.DualBroadcast != nil && *meta.DualBroadcast && meta.DualBroadcastParams != nil && meta.FwdrDestAddress != nil &&
+	if txMeta != nil &&
+		txMeta.DualBroadcast != nil && *txMeta.DualBroadcast && txMeta.DualBroadcastParams != nil && txMeta.FwdrDestAddress != nil &&
 		tx.AttemptCount == 1 && !tx.IsPurgeable {
 		// Auction & Validate
-		meta, err := a.SendRequest(ctx, tx, attempt, *meta.DualBroadcastParams, tx.ToAddress)
+		meta, err := a.SendRequest(ctx, tx, attempt, txMeta, tx.ToAddress)
```
dimriou: `SendRequest` only has one caller. It would be better if we explode the params and separate the `DualBroadcastParams` from the `feedAddress`. We can pass both independently keep the `DualBroadcastParams` as is and just pass `feeAddress` to the underlying callers. This avoid spamming `*txMeta.FwdrDestAddress` .
```suggestion
		meta, err := a.SendRequest(ctx, tx, attempt, *meta.DualBroadcastParams, tx.ToAddress, feedAddress)
```

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -331,7 +330,11 @@ type Metacalldata struct {
 	GasRefundBeneficiary common.Address
 }
 
-func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transaction, attempt *types.Attempt, dualBroadcastParams string, fwdrDestAddress common.Address) (*MetacalldataResponse, error) {
+func (a *MetaClient) SendRequest(parentCtx context.Context, tx *types.Transaction, attempt *types.Attempt, txMeta *types.TxMeta, fwdrDestAddress common.Address) (*MetacalldataResponse, error) {
+	if txMeta == nil || txMeta.FwdrDestAddress == nil || txMeta.DualBroadcastParams == nil {
+		return nil, errors.New("missing tx meta fields for dual broadcast request")
+	}
```
dimriou: I would consider this redundant as we make the same check right before we call `SendRequest`.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
```diff
@@ -194,27 +193,27 @@ func (a *MetaClient) PendingNonceAt(ctx context.Context, address common.Address)
...
+	txMeta, err := tx.GetMeta()
 	if err != nil {
 		return err
 	}
 
 	// #1
-	if meta != nil &&
-		meta.DualBroadcast != nil && *meta.DualBroadcast && meta.DualBroadcastParams != nil && meta.FwdrDestAddress != nil &&
+	if txMeta != nil &&
+		txMeta.DualBroadcast != nil && *txMeta.DualBroadcast && txMeta.DualBroadcastParams != nil && txMeta.FwdrDestAddress != nil &&
```
dimriou: No need to rename this local variable, `meta` seems clear enough.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
```diff
@@ -68,49 +69,54 @@ func NewMetaMetrics(chainID string, lggr logger.Logger) (*MetaMetrics, error) {
...
 	)
 }
 
 // RecordLatency records the latency of Meta endpoint requests
-func (m *MetaMetrics) RecordLatency(ctx context.Context, duration time.Duration) {
+func (m *MetaMetrics) RecordLatency(ctx context.Context, duration time.Duration, feedAddress common.Address) {
 	m.latencyHistogram.Record(ctx, duration.Milliseconds(),
 		metric.WithAttributes(
 			attribute.String("chainID", m.chainID),
+			attribute.String("feedAddress", strings.ToLower(feedAddress.Hex())),
```
dimriou: I would consider not using `strings.ToLower` as this forces other callers to ensure they call it in other places to stay consistent. If they miss it, then some attributes will be upper case and other lower case. Just by calling `.Hex()` , which is required by the attribute param, we ensure everyone does the same.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: And there's only one caller, fair

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: The toLower call is to prevent mixed-case addresses from being labels in PromQL, because PromQL doesn't have a `lowercase` function, making it impossible to match a mixed-case address with a lowercased address.

We encountered this already in matching SDS metrics (lowercased addresses) with RDDPE metrics (checksummed addresses), see https://github.com/smartcontractkit/rdd-prometheus-exporter/pull/139.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Agree on its own, but consider [line 207](https://github.com/smartcontractkit/chainlink-evm/blob/develop/pkg/txm/clientwrappers/dualbroadcast/meta_client.go#L207), there we reassign/shadow `meta` to a different variable, so it changes from `TxMeta` struct to a `MetacalldataResponse` struct. 
This makes it impossible to call `meta.FwdrDestAddress` after line 207.

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Fixed

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Makes sense, updated

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
dimriou: Ok makes sense. We could also rename the returned value of `SendRequest` since meta is now used inside here, whatever feels best.

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
dimriou: But `Hex()` doesn't return a random mixed case, it returns the EIP55-compliant hex string. Can't we use that standard for querying?

pkg/txm/clientwrappers/dualbroadcast/meta_client.go
cll-gg: Good point, I've opted for renaming the returned value from `SendRequest` instead

pkg/txm/clientwrappers/dualbroadcast/meta_metrics.go
cll-gg: True, that's likely better and more consistent. I'll have to update SDS for that

cll-gg: Removed the lowercase here

dimriou (approved):
Nice, looks pretty.

ilija42 (approved):

PR #503, opened 2026-07-15
Bump chainlink-protos/svr to v1.3.0 and emit per-OFA gas limit

mostlyconsistent:
## What                                                                                                                                                                                              
                                                                                                                                                                                                       
- Bumps `github.com/smartcontractkit/chainlink-protos/svr` from `v1.2.0` → `v1.3.0`, which adds the optional `gas_limit` field (9) to the `TxMessage` proto.                                         
- Plumbs the per-OFA "accountability" (OFA-tiered) gas limit into the emitted SVR `TxMessage.GasLimit`.                                                                                              
                                                                                                                                                                                                       
## Why                                                                                                                                                                                               
                                                                                                                                                                                                       
The OFA gas tier reserves the last two digits of the gas limit per OFA backend so an on-chain transaction can be attributed to a specific OFA (e.g. Titan / MEV-Share) for auction and performance comparison. Until now that tiered limit was computed at broadcast time but never made it into the beholder `TxMessage`, so downstream consumers couldn't attribute transactions.  

Builds on top of #462

### DIFF
--- a/pkg/txm/metrics.go
+++ b/pkg/txm/metrics.go
@@ -58,7 +58,7 @@ type Metrics interface {
 	ReachedMaxAttempts(context.Context, bool)
 	RecordTimeUntilTxConfirmed(context.Context, float64)
 	SetRPCNonce(context.Context, common.Address, uint64)
-	EmitTxMessage(context.Context, common.Hash, common.Address, *types.Transaction) error
+	EmitTxMessage(ctx context.Context, txHash common.Hash, fromAddress common.Address, tx *types.Transaction, gasLimit uint64) error
 }
 
 // txmMetrics is the default metrics recorder for the TXMv2 transaction lifecycle.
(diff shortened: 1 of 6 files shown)


jmank88 (dismissed):

pkg/txm/metrics.go
```diff
@@ -194,11 +194,14 @@ func (m *txmMetrics) SetRPCNonce(ctx context.Context, address common.Address, no
...
-func (m *txmMetrics) EmitTxMessage(ctx context.Context, txHash common.Hash, fromAddress common.Address, tx *types.Transaction) error {
-	return EmitTxMessage(ctx, m.chainID.String(), txHash, fromAddress, tx)
+func (m *txmMetrics) EmitTxMessage(ctx context.Context, txHash common.Hash, fromAddress common.Address, tx *types.Transaction, gasLimit uint64) error {
+	return EmitTxMessage(ctx, m.chainID.String(), txHash, fromAddress, tx, gasLimit)
 }
 
-func EmitTxMessage(ctx context.Context, chainID string, txHash common.Hash, fromAddress common.Address, tx *types.Transaction) error {
+// EmitTxMessage emits an SVR TxMessage to beholder. gasLimit is the gas limit of the broadcast
+// attempt; for OFA (dual-broadcast) transactions this is the per-OFA "accountability" gas tier,
+// which attributes the on-chain transaction to a specific OFA. A zero gasLimit is left unset.
```
dimriou: This message seems auto-generated. `EmitTxMessage` should be agnostic as to how OFA tiering works. This is just a method for sending transaction object data. How we price the gas limit of different OFA attempts should be encapsulated inside the `ofa_backend.go` file instead. Let's remove this.

pkg/txm/metrics.go
```diff
@@ -58,7 +58,7 @@ type Metrics interface {
 	ReachedMaxAttempts(context.Context, bool)
 	RecordTimeUntilTxConfirmed(context.Context, float64)
 	SetRPCNonce(context.Context, common.Address, uint64)
-	EmitTxMessage(context.Context, common.Hash, common.Address, *types.Transaction) error
+	EmitTxMessage(ctx context.Context, txHash common.Hash, fromAddress common.Address, tx *types.Transaction, gasLimit uint64) error
```
dimriou: We should either have named params for every method or none. I'd skip this for the scope of this PR.

pkg/txm/metrics.go
```diff
@@ -230,6 +230,10 @@ func EmitTxMessage(ctx context.Context, chainID string, txHash common.Hash, from
 		DualBroadcastParams: dualBroadcastParams,
 	}
 
+	if gasLimit != 0 {
+		message.GasLimit = &gasLimit
+	}
+
```
dimriou: nit: can you move this under the dualBroadcastParams and do it the same way we do it there? I think it look more consistent that way.

dimriou (approved):

jmank88 (approved):

PR #523, opened 2026-08-14
various cleanup

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-5694

### DIFF
--- a/pkg/txmgr/evm_tx_store.go
+++ b/pkg/txmgr/evm_tx_store.go
@@ -1311,15 +1311,16 @@ func (o *evmTxStore) GetAbandonedTransactionsByBatch(ctx context.Context, chainI
 	return txes, err
 }
 
-func (o *evmTxStore) GetTxByID(ctx context.Context, id int64) (txe *Tx, err error) {
+func (o *evmTxStore) GetTxByID(ctx context.Context, id int64) (*Tx, error) {
 	var cancel context.CancelFunc
 	ctx, cancel = o.stopCh.Ctx(ctx)
 	defer cancel()
 
-	err = o.Transact(ctx, true, func(orm *evmTxStore) error {
+	var txe *Tx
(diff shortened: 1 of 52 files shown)


ibrajer (dismissed):

pkg/txmgr/evm_tx_store.go
```diff
@@ -1335,7 +1335,7 @@ func (o *evmTxStore) GetTxByID(ctx context.Context, id int64) (txe *Tx, err erro
 		return nil
 	})
 
-	return txe, nil
+	return
```
RensR: Would be slightly more intuitive if we `return txe, err`?

pkg/config/toml/defaults/fallback.toml
```diff
@@ -33,6 +33,7 @@ ResendAfterThreshold = '1m'
 
 [Transactions.AutoPurge]
 Enabled = false
+Threshold = 5
```
dhaidashenko: Why do we need to set default for it. Node won't start if the value is missing and not having a fallback requires chain integrator to define proper value.

pkg/txmgr/evm_tx_store.go
jmank88: Maybe, but then the named return seems odd. I'll tweak it

pkg/config/toml/defaults/fallback.toml
jmank88: Well we have 7 chains that do not set the threshold. Should they be customized individually instead?
Screenshot 2026-08-14 at 10 37 06 AM

pavel-raykov (dismissed):

pavel-raykov (dismissed):

pavel-raykov (approved):

dhaidashenko (approved):

RensR (approved):

dimriou (approved):
Approving for TXMv2 changes.

PR #533, opened 2026-08-21
use forge std from npm

RensR:
Removes the vendor forge-std (20k+ lines) and replaces it with an NPM ref. The PR also updates it, which is significantly easier/safer using the ref.


.github/workflows/solidity-foundry.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           || needs.changes.outputs.non_src_changes == 'true' }}
         uses: ./.github/actions/setup-nodejs
         with:
-          prod: "true"
+          prod: "false"
```
RensR, line 146: Needed because forge-std is a dev dependency

contracts/GNUmakefile
```diff
@@ -1,6 +1,6 @@
 # ALL_FOUNDRY_PRODUCTS contains a list of all products that have a foundry
 # profile defined and use the Foundry snapshots.
-ALL_FOUNDRY_PRODUCTS = automation-cre keystone l2ep llo-feeds operatorforwarder payments shared workflow data-feeds
```
RensR, line 3: These have been moved out a while ago, fixing tech debt

pavel-raykov (approved):

RayXpub (approved):

ibrajer (approved):

PR #540, opened 2026-08-26
Add TXMv2 fixes

dimriou:
This PR:
- Fixes nonce initialization
- Optimizes pending nonce check during initial transmission.
- Fixes an edge case with MetaClient.
- Cleanup

Best if it gets merged after the StuckTxDetection fixes:
- https://github.com/smartcontractkit/chainlink-evm/pull/539


pkg/txm/txm.go
```diff
@@ -242,17 +253,17 @@ func (t *Txm) loop(address common.Address, triggerCh chan struct{}) {
 			t.lggr.Debug("Transaction broadcasting time elapsed: ", time.Since(start))
 		}
 		if bo {
-			broadcastCh = time.After(broadcastWithBackoff.Duration())
+			broadcastTimer.Reset(broadcastWithBackoff.Duration())
 		} else {
 			broadcastWithBackoff.Reset()
-			broadcastCh = time.After(timeutil.JitterPct(0.1).Apply(broadcastInterval))
+			broadcastTimer.Reset(timeutil.JitterPct(0.1).Apply(broadcastInterval))
 		}
```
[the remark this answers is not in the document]
dimriou: Resetting the timer without draining its channel is safe: since Go 1.23 timer channels are unbuffered and Reset guarantees no stale expirations are received afterwards.

pkg/txm/txm.go
```diff
@@ -242,17 +253,17 @@ func (t *Txm) loop(address common.Address, triggerCh chan struct{}) {
 			t.lggr.Debug("Transaction broadcasting time elapsed: ", time.Since(start))
 		}
 		if bo {
-			broadcastCh = time.After(broadcastWithBackoff.Duration())
+			broadcastTimer.Reset(broadcastWithBackoff.Duration())
 		} else {
 			broadcastWithBackoff.Reset()
-			broadcastCh = time.After(timeutil.JitterPct(0.1).Apply(broadcastInterval))
+			broadcastTimer.Reset(timeutil.JitterPct(0.1).Apply(broadcastInterval))
```
cll-gg: While you're at it, maybe add a comment about the reason for doing this? Kinda hard to understand why

pkg/txm/txm.go
```diff
@@ -434,7 +448,9 @@ func (t *Txm) BackfillTransactions(ctx context.Context, address common.Address)
 		// - The transaction has never been broadcasted successfully before
 		// - The last broadcast was more than RetryBlockThreshold blocks ago
 		// - The transaction is purgeable
-		if tx.LastBroadcastAt == nil || time.Since(*tx.LastBroadcastAt) > (t.config.BlockTime*time.Duration(t.config.RetryBlockThreshold)) || tx.IsPurgeable {
+		if tx.LastBroadcastAt == nil ||
+			time.Since(*tx.LastBroadcastAt) > (time.Duration(t.config.RetryBlockThreshold)*t.config.BlockTime) ||
+			tx.IsPurgeable {
 			t.lggr.Info("Rebroadcasting attempt for txID: ", tx.ID)
```
cll-gg: Would be great if we could add our `transactionLifecycleID` tracing here, but I guess that's not easily done since that's specific to the `dualContractTransmitter`, right?

pkg/txm/txm.go
```diff
@@ -278,7 +289,7 @@ func (t *Txm) BroadcastTransaction(ctx context.Context, address common.Address)
 		// to insufficient balance. We're making this trade-off to avoid storing stuck transactions and making unnecessary
 		// RPC calls. The upper limit is always MaxInFlightTransactions regardless of the pending nonce.
 		if unconfirmedCount >= MaxInFlightSubset {
-			if unconfirmedCount > MaxInFlightTransactions {
+			if unconfirmedCount >= MaxInFlightTransactions {
```
cll-gg, line 280: I would've expected there to be a test to be updated for this change, don't we have unit tests for this functionality?

pkg/txm/txm.go
dimriou: Yes, but I can still add it anyway.

pkg/txm/txm.go
dimriou: The test was checking for `MaxInFlightTransactions + 1` although the limit was at `MaxInFlightTransactions`, hence why it still passes. Now it checks exactly at the limit.

pkg/txm/txm.go
dimriou: This simply re-uses a single timer instead of creating a new one each time.

cll-gg: That makes sense, but from reading it it's hard to see the reason for the timer and why we reset it with a duration in one place while resetting it with the broadcast interval in the other place

pkg/txm/txm.go
cll-gg: Nice!

pkg/txm/txm.go
cll-gg: Ok cool

pkg/txm/txm.go
dimriou: Since we got at it, I moved around a few things to make the loop method cleaner, including the backoff mechanism and why it's there.

amit-momin (approved):

pkg/txm/txm.go
```diff
@@ -246,18 +240,13 @@ func (t *Txm) loop(address common.Address, triggerCh chan struct{}) {
 
 	for {
 		start := time.Now()
-		bo, err := t.BroadcastTransaction(ctx, address)
+		shouldBackoff, err := t.BroadcastTransaction(ctx, address) // use a backoff if transmission is being throttled.
```
cll-gg, line 243: Now it makes sense, thanks!

cll-gg (approved):

PR #548, opened 2026-09-04
OEV-1648 Emit whitelisted EVM chain config metric

mostlyconsistent:
## What                                                                                                                                                                                                                                
                                                                                                                                                                                                                                         
Emits SVR-relevant EVM chain config as an OpenTelemetry info gauge from each chain's startup path:                                                                                                                                                                                                                  
```
evm_chain_config_info{chain_id, transaction_v2_enabled, dual_broadcast} = 1 
```                                                                                                                                                 

Value is always 1; all state is in the labels. The label whitelist is exhaustive and contains no URL.                                                                                                                                                                                                        
                                                                                                                                                                                                                                         
 ## Why                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                         
Since ~2026-07-20 the INCIDENT-2541 mitigation (drop filters in chip-ingress and beholder-gateway-logs) drops the whole NodeConfig beholder event, because that event carried the entire node TOML — including RPC/OFA URLs with embedded credentials. Collateral damage:                                                                                                                                                                                                        
- `svr_beholder.node_configs` (BigQuery) stopped ingesting.                                                                                                                                                                            
- `node_config_boot_time`, which OFE derived from that event, went absent.                                                                                                                                                             
- **DF-SVR-018** ("NOP node config DualBroadcast/TxV2 disabled") depends on that metric and has been evaluating blind since.


mostlyconsistent:
@cll-gg let's look at this after the weekend. No code should land on a Friday :)

pkg/chains/legacyevm/chain_config_metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	gauge, err := meter.Int64Gauge(
+		evmChainConfigInfoMetricName,
+		metric.WithDescription("Whitelisted EVM chain configuration; value is always 1, state is in the labels"),
+		metric.WithUnit("{info}"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create %s gauge: %w", evmChainConfigInfoMetricName, err)
+	}
+
+	gauge.Record(ctx, 1, metric.WithAttributes(chainConfigAttributes(chainID, txV2Enabled, dualBroadcast)...))
```
jmank88: Constructing and emitting together like this seems problematic. Can we construct the gauge once ahead of time?

mostlyconsistent: can you go into more details on why it is problematic?

pkg/chains/legacyevm/chain_config_metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+// emitChainConfigInfo reports this chain's whitelisted config state. A nil meter
+// falls back to the global beholder meter. Failures are non-fatal: a metrics
+// hiccup must never block chain startup.
+func (c *chain) emitChainConfigInfo(ctx context.Context, meter metric.Meter) {
+	if meter == nil {
+		meter = beholder.GetMeter()
```
jmank88: We typically want to push these global beholder calls up/out to a higher level so that we have clean injection everywhere internal. It also keeps high level packages from being imported if we reference just the otel types.

pkg/chains/legacyevm/chain_config_metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	txV2 := c.cfg.EVM().Transactions().TransactionManagerV2()
+	if err := recordChainConfigInfo(ctx, meter, c.id.String(), txV2.Enabled(), derefBool(txV2.DualBroadcast())); err != nil {
```
jmank88: IMHO the indirection makes this harder to read compared to inlining:
```suggestion
	var dualBroadcast bool
	if db := txV2.DualBroadcast(); db != nil {
		dualBroadcast = *db
	}
	if err := recordChainConfigInfo(ctx, meter, c.id.String(), txV2.Enabled(), dualBroadcast); err != nil {
```
But maybe a different name would be fine?
```suggestion
	if err := recordChainConfigInfo(ctx, meter, c.id.String(), txV2.Enabled(), isTrue(txV2.DualBroadcast())); err != nil {
```

pkg/chains/legacyevm/chain.go
```diff
@@ -148,6 +151,11 @@ type ChainOpts struct {
 
 	DS sqlutil.DataSource
 
+	// Meter is used to report chain metrics. Defaults to the global beholder
+	// meter when unset.
+	// TODO: make this required once chainlink core passes it explicitly.
```
cll-gg, line 156: This TODO will still be part of the current ticket?

mostlyconsistent: its a chicken and egg problem. Once this lands, I can land a version of this https://github.com/smartcontractkit/chainlink/pull/23726, and then undo the TODO :)

Up to you how many tickets you wanna have for the work.

cll-gg (approved):

jmank88 (approved):

PR #557, opened 2026-09-15
OEV-1803 Remove Meter() TODO post core bump

mostlyconsistent:
Makes `legacyevm.ChainOpts.Meter` required and drops the `beholder.GetMeter()` fallback in `newChain`

The fallback was a temporary bridge added with `evm_chain_config_info` (#548): chainlink core did not pass a meter, so chains silently fell back to the global beholder meter. smartcontractkit/chainlink#23726 has landed and core now wires `Meter: beholder.GetMeter()` explicitly in `application.go`, so the TODO can go and the dependency becomes explicit rather than implicit global state.


pkg/chains/legacyevm/chain.go
```diff
@@ -187,6 +184,9 @@ func (o ChainOpts) Validate() error {
 	if o.DS == nil {
 		err = errors.Join(err, errors.New("nil DS"))
 	}
+	if o.Meter == nil {
+		err = errors.Join(err, errors.New("nil Meter"))
+	}
```
jmank88, line 189: It might be better to wait a bit before making this strictly required, since there is still a chance that we need to patch an older release with a newer chainlink-evm.

cll-gg (approved):

PR #559, opened 2026-09-16
Bump common and remove legacy write target

cedric-cordenier:


pkg/relay/evm.go
```diff
@@ -371,49 +355,7 @@ func NewOCR3CapabilityConfigProvider(ctx context.Context, lggr logger.Logger, ch
 
 // NewPluginProvider, but customized to use a different config provider
 func (r *Relayer) NewOCR3CapabilityProvider(ctx context.Context, rargs commontypes.RelayArgs, pargs commontypes.PluginArgs) (commontypes.OCR3CapabilityProvider, error) {
```
dhaidashenko, line 357: This seems to be used https://github.com/smartcontractkit/chainlink/blame/fa897c3d55f584de6ece6255c061b8ef6904d0d5/core/services/relay/relay.go#L57

cedric-cordenier: I've tweaked this so that it just returns a noop transmitter; this should make it safe with other users

dhaidashenko (approved):

george-dorin (approved):

